PR #7, opened 2016-01-12
Fuse 2 9 bugfix

awaiskhan2007:


Nikratio:
Could you rebase this on current `fuse_2_9_bugfix`? You seem to have started from some unusual commit (since your pull request includes commits from Miklos). Probably Miklos rebased his tree at some point and you cloned before that.

Nikratio:
Looks like this was an accident.

PR #16, opened 2016-02-09
remove double calculation of filename length and dirent size

trapexit:


Nikratio:
Hm. It sounds like you're removing duplicated code, but the commit actually adds more new lines than it removes old ones.. so is this really an improvement? Or did I misunderstand the purpose of the patch?

trapexit:
I can split it up into multiple pull requests if you prefer. I removed the double calling of strlen and fuse_dirent_size (which required the function restructuring) and cleaned up the type <-> mode conversion to use the proper macro.

Nikratio:
No, there's no need to split it up. I was just wondering about the benefit of this if the net-gain is an increase in lines of code. If there is duplicated code, you would expect that factoring it out should result in a net *loss* of LOC.

trapexit:
Yeah. The overage is largely formatting, comments (it wasn't clear why you'd allow a NULL buffer), and the macro.

lib/fuse_lowlevel.c
```diff
@@ -320,38 +320,46 @@ int fuse_reply_iov(fuse_req_t req, const struct iovec *iov, int count)
 	return res;
 }
 
-static size_t fuse_dirent_size(size_t namelen)
-{
-	return FUSE_DIRENT_ALIGN(FUSE_NAME_OFFSET + namelen);
-}
+#ifndef IFTODT
+# define IFTODT(mode) (((mode) & 0170000) >> 12)
+#endif
```
Nikratio: Could you document what this function does? What does `IF` and `DT` stand for?

lib/fuse_lowlevel.c
```diff
@@ -320,38 +320,46 @@ int fuse_reply_iov(fuse_req_t req, const struct iovec *iov, int count)
...
 
-static void fuse_add_dirent(struct fuse_dirent *dirent, const char *name,
-			    const struct stat *stbuf, off_t off)
+static void fuse_add_dirent(struct fuse_dirent *dirent,
+                            const char *name, const size_t namelen,
+                            const size_t entlen, const size_t entlen_padded,
+			    const struct stat *stbuf, const off_t off)
```
Nikratio: I think these are way too many parameters, especially since most of them are only used once.  How about this: make the new signature:

```
static void fuse_add_dirent(struct fuse_dirent *dirent, const char *value, const size_t len,
			    const struct stat *stbuf, off_t off)
```
..and then..

lib/fuse_lowlevel.c
```diff
@@ -320,38 +320,46 @@ int fuse_reply_iov(fuse_req_t req, const struct iovec *iov, int count)
...
+	unsigned padlen = entlen_padded - entlen;
 
 	dirent->ino = stbuf->st_ino;
 	dirent->off = off;
 	dirent->namelen = namelen;
-	dirent->type = (stbuf->st_mode & 0170000) >> 12;
+	dirent->type = IFTODT(stbuf->st_mode);
 	strncpy(dirent->name, name, namelen);
 	if (padlen)
 		memset(dirent->name + namelen, 0, padlen);
```
Nikratio: ..and make these two lines the responsibility of the caller?

Nikratio:
Looking at this again in more detail, I agree with your sentiment that this could be done better. But I dislike the proliferation of new parameters introduced in your patch. Is there anything I missed, or could we avoid that by making the chance as suggested above?

lib/fuse_lowlevel.c
trapexit: IFTODT is a standard macro as defined in dirent.h.

http://www.gnu.org/software/libc/manual/html_node/Directory-Entries.html

Nikratio: Ah, thanks. But then, why do you anticipate that it may not be defined?

lib/fuse_lowlevel.c
trapexit: I don't follow. The values must be calculated to determine whether or not to perform `fuse_add_dirent` in the first place and to leave the pad length clearing to the caller is denormalizes the behavior. Either it'd need to calculate it twice or would need to leave to caller to clumsily know to clear the rest of the dirent. One is what existed before and the other is dangerous.

lib/fuse_lowlevel.c
trapexit: Fear of the unknown. Not sure why IFTODT wasn't used prior.

trapexit:
I could make `fuse_add_direntry` take the buffer length and return the padded size. Make it noop in the function if padded <= bufsize.

lib/fuse_lowlevel.c
Nikratio: Yes, but why do the values need to get passed to `fuse_add_dirent`? But you are right about the clearining be a little clumy because of the need to cast.

Nikratio:
I think this sounds like a promising approach. That way the only new argument would be the buffer length, right?

trapexit:
It'd remove entity length and the padded version from the signature and add the buffer size so the comparison can be done in fuse_add_dirent.

Nikratio:
Sorry for letting this slack for so long. I just can't really convince myself that this is an improvement - but neither am I certain enough to reject the pull. So you could try to implement your most recent proposal, but I don't give a guarantee of merging it.

trapexit:
Sorry I've let this linger. Been busy. But I don't see how it's not an improvement. The code clearly computes the same value twice and strlen is O(n). Filenames may not be more than 255 bytes on most filesystems but why waste the cycles?

I'll try to make the tweak today and resubmit.

trapexit:
I've pushed a slightly altered, hopefully clearer, version. There just isn't a good way to handle this. You either need to normalize the code and put the functions into add_direntry and add_direntry_plus or you have more arguments to handle the fact that the padding and alignment for the two structures are slightly different. It may be more lines of code but it's less logic and no duplication of of computing the size.

Nikratio:
You're saving CPU cycles, but reading the code puts more cognitive load on the programmer.

trapexit:
I've normalized the code and wrote functions which do what it did before but more explicitly. Can you articulate how it's harder to read? The code that's there is hard to read. denormalized, and as a result duplicated behavior.

The interface isn't the best and probably could use a full rewrite but if you're doing a lot of readdir'ing this cuts the data scanned by half.

Nikratio:
My main concern is that the `fuse_add_dirent` function has so many parameters that its prototype (4 lines) is longer than its implementation (4 lines that could be reduced to 3). But I think we're getting somewhere. Could we just inline this function in the two places where it's used? 

Also, if `IFTODT` is not defined, can you really rely on our definition to be correct? I think it'd be better to define it in terms of `S_IFMT`, and error out if that is not defined either.

PR #22, opened 2016-02-22
Two symbols (fuse_main and fuse_new) now point to a recent version of the function

dvorszkid:
In the libfuse.so shared object the "fuse_main" pointed to "fuse_main_compat2@@FUSE_2.2", the "fuse_new" pointed to "fuse_new_compat2@@FUSE_2.2". After this patch they will point to "fuse_main@@FUSE_2.2" and "fuse_new@@FUSE_2.6".

All the other usages of FUSE_SYMVER seemed to be valid (either they point to a correct method by a correct symbol (like "fuse_setup@FUSE_2.5" -> "fuse_setup_compat25"), or they seem to be there for legacy reasons - the ones starting with "__").

I don't think that plain function names (without prefix and postfix) should point to older versions of the API functions, that does not seem logical. If someone needs an explicit (older) version of the function, load that version explicitly.

Altough I am not sure if this patch breaks any programs depending on libfuse, so please review. :)

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -4886,7 +4886,6 @@ FUSE_SYMVER(".symver fuse_exited,__fuse_exited@");
 FUSE_SYMVER(".symver fuse_process_cmd,__fuse_process_cmd@");
 FUSE_SYMVER(".symver fuse_read_cmd,__fuse_read_cmd@");
 FUSE_SYMVER(".symver fuse_set_getcontext_func,__fuse_set_getcontext_func@");
-FUSE_SYMVER(".symver fuse_new_compat2,fuse_new@");
 FUSE_SYMVER(".symver fuse_new_compat22,fuse_new@FUSE_2.2");
 
 #endif /* __FreeBSD__ || __NetBSD__  */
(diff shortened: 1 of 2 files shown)


Nikratio:
I believe libfuse didn't always use symbol versioning. If you compiled against such an old version, the loader will attempt to load e.g. "fuse_new". Prior to your patch, this was resolved to "fuse_new_compat2" (i.e., a compatibility wrapper). With your patch, it will point to the newest version and the program will break. Am I missing something?

I don't understand the second change either. If a program was compiled against libfuse 2.2 and uses "fuse_main", why should this be resolved to "fuse_main" rather than "fuse_main_compat2"?

dvorszkid:
I see your point. So it's normal that I have to specify the symbol version number of ex. fuse_new when loading it dynamically, even though I just wan't the latest version of it. It seems logical actually, but then I should specify all the symbol versions, just to avoid breaking the program when a new version of a fuse API function comes out with different behavior / parameters..

Thanks for your time!

ngg:
If libfuse would have originally contained symbol versions then you wouldn't need to specify it, because there could be always a default version (like libfuse@@FUSE_2.6) and older versions (like libfuse@FUSE_2.2), and the linker would select all the default versions at compile time and save their versions. So if a precompiled binary runs with a newer libfuse then it would load the versions that were the default when compiling. This doesn't work if there is an unversioned symbol, because then the linker cannot save the default version, so it would select the default at runtime.

PR #24, opened 2016-03-09
fuse.h doc fix: The f_frsize field is not ignored by the statfs operation

jblumsch:
Apparently the `f_frsize` field has been passed on transparently by `statfs` since
2b4781100812d42e704c39c51303cd28ad3f9aa6 (Nov 28, 2005).

There is no point in discouraging users from revealing their fs fragment size…


Nikratio:
Thanks for your contribution!

PR #29, opened 2016-03-21
Remove leading _ on header guards to comply with reserved identifier requirements

### LINKED ISSUE #28: Avoid use of reserved identifiers
According to the C standard section 17.6.4.3.2 ("Global Names"):

> Certain sets of names and function signatures are always reserved to the implementation:
* Each name that contains a double underscore __ or begins with an underscore followed by an uppercase letter (2.12) is reserved to the implementation for any use.
* Each name that begins with an underscore is reserved to the implementation for use as a name in the global namespace.

Below is a list of ``define``s which violate this section (by using reserved identifiers):

```
include/fuse_common.h:#define _FUSE_COMMON_H_
include/fuse_opt.h:#define _FUSE_OPT_H_
include/fuse_lowlevel.h:#define _FUSE_LOWLEVEL_H_
include/cuse_lowlevel.h:#define _CUSE_LOWLEVEL_H_
include/fuse.h:#define _FUSE_H_
include/fuse_kernel.h:#define _LINUX_FUSE_H
```

Simply removing the leading ``_`` corrects this issue.

HalosGhost:
should close #28 (note that anything that manually checks header guards might be broken by this—though that should be rare since essentially nothing should be doing that).

### DIFF
--- a/include/cuse_lowlevel.h
+++ b/include/cuse_lowlevel.h
@@ -9,8 +9,8 @@
   Read example/cusexmp.c for usages.
 */
 
-#ifndef _CUSE_LOWLEVEL_H_
-#define _CUSE_LOWLEVEL_H_
+#ifndef CUSE_LOWLEVEL_H_
+#define CUSE_LOWLEVEL_H_
 
 #ifndef FUSE_USE_VERSION
 #define FUSE_USE_VERSION 29
@@ -84,4 +84,4 @@ int cuse_lowlevel_main(int argc, char *argv[], const struct cuse_info *ci,
 }
 #endif
 
-#endif /* _CUSE_LOWLEVEL_H_ */
+#endif /* CUSE_LOWLEVEL_H_ */
--- a/include/fuse_kernel.h
+++ b/include/fuse_kernel.h
@@ -104,8 +104,8 @@
  *  - add FUSE_NO_OPEN_SUPPORT flag
  */
 
-#ifndef LINUX_FUSE_H
-#define LINUX_FUSE_H
+#ifndef _LINUX_FUSE_H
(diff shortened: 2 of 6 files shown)


include/cuse_lowlevel.h
```diff
@@ -9,8 +9,8 @@
   Read example/cusexmp.c for usages.
 */
 
-#ifndef _CUSE_LOWLEVEL_H_
-#define _CUSE_LOWLEVEL_H_
+#ifndef CUSE_LOWLEVEL_H_
```
pramodk-git: Given the leading "_" is being removed, it might make sense to remove the trailing one as well.

I think this is a good catch though; of all the major open source projects, I could see only gcc using the correct notation.

HalosGhost: @pramodk-git, the trailing ``_`` is a stylistic one not mandated by the standard. I am happy to remove them if the project would prefer it to be done that way before the merge, but that is not a decision I feel comfortable dictating since it is a stylistic one.

include/fuse_kernel.h
```diff
@@ -104,8 +104,8 @@
  *  - add FUSE_NO_OPEN_SUPPORT flag
  */
 
-#ifndef _LINUX_FUSE_H
```
Nikratio: I think this file should not be changed unless you also get the changes into the kernel (http://lxr.free-electrons.com/source/include/uapi/linux/fuse.h#L111). Other than that, this looks fine.

HalosGhost: Well, it's deeply depressing to find out that the kernel itself violates this… I will revert that guard tonight.

PR #38, opened 2016-04-17
Add support for reading from directories

emilymaier:
read(2) on a directory is an odd syscall, but the VFS permits it and existing filesystems use it, so this is the userspace side of letting FUSE filesystems implement it. I also removed fuse_kernel.h since it duplicates an outdated version of the kernel FUSE header.

### DIFF
--- a/example/.gitignore
+++ b/example/.gitignore
@@ -9,3 +9,4 @@
 /fselclient
 /cusexmp
 /fuse_lo-plus
+/dirread
--- a/example/Makefile.am
+++ b/example/Makefile.am
@@ -3,7 +3,7 @@
 AM_CPPFLAGS = -I$(top_srcdir)/include -D_REENTRANT
 noinst_HEADERS = fioc.h
 noinst_PROGRAMS = fusexmp fusexmp_fh null hello hello_ll fioc fioclient \
-		  fsel fselclient cusexmp fuse_lo-plus
+		  fsel fselclient cusexmp fuse_lo-plus dirread
 
 LDADD = ../lib/libfuse3.la
 fusexmp_fh_LDADD = ../lib/libfuse3.la @fusexmp_fh_libs@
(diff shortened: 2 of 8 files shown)


Nikratio:
I believe FUSE deliberately ships its own version of `fuse_kernel.h` - but unfortunately I don't remember the exact rationale. I'll ask Milklos, but expect that to take a while...

Nikratio:
Thanks for the patch! A few comments:

- I believe adding a new operation (`FUSE_DIR_READ`) requires a corresponding patch to the FUSE kernel module. Could you point me to the relevant (kernel) commit?
- Why do you add a new `dir_read` function? Wouldn't it make sense to just use `read`? At least for the low-level API I do not see any problems.
- Will `dir_read` be preceded by a call to `open` or `opendir`? Please mention this in the comments (of both `dir_read` and the relevant open functions)
- Could you please add an example that can be used to test this functionality? Something similar to `examples/{fsel,fselclient}.c` would be great.
- Is it possible to split out the removal of `fuse_kernel.h`, or does your patch depend on the newer version? As I said, I'd like to wait for Miklos comments on that.

emilymaier:
I believe some of this is responded to in the kernel patch thread :) As for the rest, this branch won't build without the patch I put on fuse-devel, no. It depends on FUSE_DIR_READ being defined. I could add it here as well, but I noticed that fuse_kernel.h is duplicated from the kernel and also somewhat outdated. The FS receives opendir from FUSE before this call; I'll add in some example code.

Nikratio:
I've heard back from Miklos. `fuse.h` is shipped both in the kernel and in the library so that they don't have to be upgraded in lockstep. If `fuse.h` was availabel only in the kernel, you'd have to update the kernel to update libfuse (or vice-versa). So please update this file in both kernel and userspace.

emilymaier:
I've pushed a new branch that adds an example, keeps fuse_kernel.h, and puts in a few other fixes.

Nikratio:
Please ping me once the kernel part is in. Thanks!

Nikratio:
I'm closing this issue for now. Please just submit a pull request when the kernel patches have been merged. Thanks!

Issue #27, opened 2016-03-18
parent exit() vs. child setsid() race


dannf:
Forwarding from http://bugs.launchpad.net/bugs/1558967
== Comment: #21 - Hendrik Brueckner - 2016-03-16 06:44:09 ==
Package: libfuse2
Version: 2.9.4-1ubuntu2

The cmsfs-fuse program is used to transfer files from a CMSFS dasd (on z/VM) to Linux. The procedure is to mount, copy files, umount. All commands are issued from within an application over an SSH connection.

The problem is that the copy intermittently fails with "Transport endpoint is not connected". The procedure is as follows:

   #mount cmsfs
   sudo /usr/bin/cmsfs-fuse /dev/dasdb /usr/wave/wavedisk
   # copy file
   /bin/cp -f /usr/wave/wavedisk/WAVEDATA.SCRIPT /usr/wave/wavedata
   /bin/cp: cannot stat '/usr/wave/wavedisk/WAVEDATA.SCRIPT': Transport endpoint is not connected
   #umount
   umount /usr/wave/wavedisk

Because the application uses JSCH to issue the commands, I worked on a non-Java reproducer using SSH.

The problem can be easily re-created with ssh as follows:

root@r3559004:~# ssh -t root@localhost "cmsfs-fuse /dev/disk/by-path/ccw-0.0.0190 /CMSFS"
Connection to localhost closed.
root@r3559004:~# ls /CMSFS
ls: cannot access '/CMSFS': Transport endpoint is not connected

Problem analysis will follow but not that is not specific to cmsfs-fuse; the problem might also occur with other fuse file systems that are mounted through an SSH connection.

== Comment: #23 - Hendrik Brueckner - 2016-03-16 07:07:30 ==
After debugging and some code review on the libfuse library, I think that
we identified the root cause. As suggested, the problem is not related
to cmsfs-fuse directly.

The cmsfs-fuse main program calls into the libfuse library() using the
fuse_main() function. The fuse_main() function later calls the
fuse_daemonize() to fork the daemon process to handle the fuse file
system I/O.

The fuse_daemonize() look at follows:

180 int fuse_daemonize(int foreground)
181 {
182 if (!foreground) {
183 int nullfd;
184
185 /*
186 * demonize current process by forking it and killing the
187 * parent. This makes current process as a child of 'init'.
188 */
189 switch(fork()) {
190 case -1:
191 perror("fuse_daemonize: fork");
192 return -1;
193 case 0:
194 break;
[...]

PR #55, opened 2016-06-16
libfuse/fuse_daemonize:  wait until daemon child process is ready

hbrueckner:
Mounting a FUSE file system remotely using SSH in combination with
pseudo-terminal allocation (-t), results in "Transport endpoint is
not connected" errors when trying to access the file system contents.

For example:

  # ssh -t root@localhost  "cmsfs-fuse /dev/disk/by-path/ccw-0.0.0190 /CMSFS"
  Connection to localhost closed.
  # ls /CMSFS
  ls: cannot access '/CMSFS': Transport endpoint is not connected

The cmsfs-fuse main program (which can also be any other FUSE file
system) calls into the fuse_main() libfuse library function.
The fuse_main() function later calls fuse_daemonize() to fork the
daemon process to handle the FUSE file system I/O.

The fuse_daemonize() function calls fork() as usual.  The child
proceeds with setsid() and then redirecting its file descriptors
to /dev/null etc.  The parent process, simply exits.

The child's functions and the parent's exit creates a subtle race.
This is seen with an SSH connection.  The SSH command above calls
cmsfs-fuse on an allocated pseudo-terminal device (-t option).

If the parent exits, SSH receives the command completion and closes
the connection, that means, it closes the master side of the
pseudo-terminal.  This causes a HUP signal being sent to the process
group on the pseudo-terminal.  At this point in time, the child might
not have completed the setsid() call and, hence, becomes terminated.
Note that fuse daemon sets up its signal handlers after fuse_daemonize()
has completed.

Even if the child has the chance to disassociate from its parent process
group to become it's own process group with setsid(), the child still
has the pseudo-terminal opened as stdin, stdout, and stderr.  So the
pseudo-terminal still behave as controlling terminal and might cause a
SIGHUP at closing the the master side.

To solve the problem, the parent has to wait until the child (the fuse
daemon process) has completed its processing, that means, has become
its own process group with setsid() and closed any file descriptors
pointing to the pseudo-terminal.

Closes: #27

Reported-by: Ofer Baruch <oferba@il.ibm.com>
Reviewed-by: Gerald Schaefer <gerald.schaefer@de.ibm.com>
Signed-off-by: Hendrik Brueckner <brueckner@linux.vnet.ibm.com>

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -176,6 +176,13 @@ int fuse_daemonize(int foreground)
 {
 	if (!foreground) {
 		int nullfd;
+		int waiter[2];
+		char completed;
+
+		if (pipe(waiter)) {
+			perror("fuse_daemonize: pipe");
+			return -1;
+		}
 
 		/*
 		 * demonize current process by forking it and killing the
@@ -188,6 +195,7 @@ int fuse_daemonize(int foreground)
 		case 0:
 			break;
 		default:
+			read(waiter[0], &completed, sizeof(completed));
 			_exit(0);
 		}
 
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks for the patch! Don't worry about the test failure, that problem has always been present in the code base (and only recently become visible when adding clang tests).

PR #61, opened 2016-07-13
Should fuse_do_release call do_forget() ?

treestem:
I see the name and id caches continually growing, even with "-o remember=60".

My test is an infinite loop of open, write, close on random file names.

fuse_clean_cache is not reaping old nodes because forget isn't called and we never decrement nlookup. With this patch, nodes go onto the lru list and are eventually reaped. Memory no longer grows continuously.


treestem:
While this problem does exist, the proposed patch is not correct.
This needs more thought.

Nikratio:
Thanks for the update, I'll close this for now.

PR #66, opened 2016-07-21
add error checking to setuid calls

Alex-Richman:
This was raised by a static analysis tool and seemed worth fixing.


Nikratio:
Thanks!

PR #74, opened 2016-09-08
Add support for posix ACLs

sforshee:
This pull request requires kernel support. Patches for Linux can be found at:

http://lkml.kernel.org/r/1472478397-131967-1-git-send-email-seth.forshee@canonical.com

With kernel support, these patches add support for posix ACLs by passing the raw posix ACL xattrs between the kernel and userspace. If supported by the filesystem, libfuse will parse the raw xattrs into an easier to use data structure and pass these to the filesystem's setacl/getacl callbacks for processing. Helpers are provided for common operations like converting between the posix and fuse ACL data structures, calculating the file mode from ACLs, updating ACLs when the file mode changes, and determining what ACLs should be inherited by children of a directory with default ACLs.

### DIFF
--- a/example/fusexmp.c
+++ b/example/fusexmp.c
@@ -33,6 +33,7 @@
 #endif
 
 #include <fuse.h>
+#include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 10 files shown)


sforshee:
Closing this based on this message from Miklos indicating that there's currently not a use case for this in libfuse right now.

https://sourceforge.net/p/fuse/mailman/message/35387972/

PR #89, opened 2016-10-14
Add fchmod FUSE operation.

jabolopes:
fchmod is analogous ftruncate: it extends the chmod operation with the
fuse_file_info argument, which includes the open file handle.

This operation allows clients of the FUSE library to distinguish between
the chmod and fchmod system calls.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -15,6 +15,19 @@ pkgconfig_DATA = fuse3.pc
 
 $(pkgconfig_DATA): config.status
 
+.PHONY: setuid_fusermount
+setuid_fusermount:
+	@echo "Attempting to use sudo to make util/fusermount setuid root"
+	@echo "If this fails, set permissions manually and re-run make test"
+	test $$(ls -n util/fusermount | awk 'NR==1 {print $$3}') -eq 0 || \
+	    sudo chown root util/fusermount
+	test -u util/fusermount || \
+	    sudo chmod u+s util/fusermount
+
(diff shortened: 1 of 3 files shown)


include/fuse.h
```diff
@@ -574,6 +574,22 @@ struct fuse_operations {
...
+	 * Change the permission bits of a file
+	 *
+	 * This method is called instead of the chmod() method if the
+	 * call was invoked from an fchmod() system call.
+	 *
+	 * TODO: Determine the version of the Linux kernel below.
+	 *
+	 * If this method is not implemented or under Linux kernel
+	 * versions earlier than X.X.X, the chmod() method will be
+	 * called instead.
```
Nikratio: I don't see any dependence on the Linux kernel anywhere. You've changed how FUSE responds to the setattr request. Why / how do you think the Linux kernel version would come into effect?

include/fuse.h
```diff
@@ -574,6 +574,22 @@ struct fuse_operations {
 	 */
 	int (*fallocate) (const char *, int, off_t, off_t,
 			  struct fuse_file_info *);
+
+	/**
+	 * Change the permission bits of a file
+	 *
+	 * This method is called instead of the chmod() method if the
+	 * call was invoked from an fchmod() system call.
```
Nikratio: Please use the same terminology as eg. the `fgetattr` and `ftruncate` docstrings.

PR #93, opened 2016-10-17
Fix gcc warnings

forgottenswitch:
Stub compile warnings. Maybe some should be error-checked.

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -151,7 +151,9 @@ int fuse_daemonize(int foreground)
 		case 0:
 			break;
 		default:
-			read(waiter[0], &completed, sizeof(completed));
+			if (read(waiter[0], &completed, sizeof(completed))) {
+				/* Ignore errors */
+			}
 			_exit(0);
 		}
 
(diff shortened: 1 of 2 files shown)


Nikratio:
Hmm. I think gcc warning about `(void) chdir("/");` is a bug in GCC. The explicit cast makes it clear that we are deliberately ignoring the return value. In my opinion the empty `if` clauses introduced by your patch are worse than the GCC warning, so I'd rather add some more explicit `(void)` castings instead and hope that gcc gets fixed.

forgottenswitch:
Then it's a bug in LLVM/clang too.
  #define IGNORE_RESULT(op) if ((op)) { }
?

Nikratio:
Amazing that there's actually a discussion about this. I didn't read the bug to the end (so I don't know which opinion eventually prevailed), but the only thing that's going to get into FUSE is a void cast and (if necessary) a patch to enable `-Wno-unused-result` when compiling with gcc6 (in gcc 4.9 the behavior is still - dare I say it? - sensible).

Nikratio:
Not interested in at least adding the `(void)` castings? I can easily to it, but then you won't be the commit author (and you found the problem).

PR #99, opened 2016-10-18
man: Document suid requirement

forgottenswitch:
/usr/bin/fusermount could not be set-uid-root if package manager blacklists suid executables.
Currently, a mount() fails, printing `"fusermount: mount failed:%s", "Operation not permitted"`,
which looks like there is a mis-configuration of fstab.


Nikratio:
Merged manually, thanks!

PR #120, opened 2016-11-23
util/fusermount.c: Fix user mounting under grsec kernel.

przemoc:
If the user was not in readproc group, realpath() would fail causing
fusermount to fail.  The privilege drop for realpath() is unnecessary
as the real access checks are done in other places.

Based on patch by Timo Teräs for fuse package in Alpine Linux's aports:
http://git.alpinelinux.org/cgit/aports/commit/?id=b5d81e456487d4dbfbdf0d07ae6ca5cf3f59d186

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -1200,18 +1200,15 @@ int main(int argc, char *argv[])
 
 	origmnt = argv[optind];
 
-	drop_privs();
 	mnt = fuse_mnt_resolve_path(progname, origmnt);
-	if (mnt != NULL) {
-		res = chdir("/");
-		if (res == -1) {
-			fprintf(stderr, "%s: failed to chdir to '/'\n", progname);
-			goto err_out;
-		}
-	}
-	restore_privs();
 	if (mnt == NULL)
 		exit(1);
+
+	res = chdir("/");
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks for the patch! Since this affects a setuid binary, I want to review this very carefully and its going to take a little longer. Could you maybe expand on where the "real access checks" are done? Or is this going to be obvious once I look at the code (I haven't done that yet).

przemoc:
Today Timo linked me to the discussion from ML (~1.5y ago), that I wasn't aware of.
https://sourceforge.net/p/fuse/mailman/message/33378384/

Nikratio:
It seems the consensus back then was to either
* Drop realpath() only if `/etc/mtab` is symlink to `/proc/mounts`. However, I think this should be a run-time rather than compile time check.
* Drop privileges, open the directory, and then use the resulting fd for all futher work.
Would either of this solve the problem for grsec?

Nikratio:
@przemoc Do you have an update on this?

Nikratio:
Since there hasn't been any activity for more than 6 months, and from what I've heard the GRsec kernel is also no longer publically available, I'll close this for now.

PR #121, opened 2016-11-23
Continue renaming stuff for fuse3

przemoc:
Some stuff has been renamed already, but in most places old names remain. It's better to be consistent.
Nothing in examples/ directory has been changed yet.


Nikratio:
Thanks! Most of this looks good. I am not a friend of renaming the `mount.fuse` manpage though, I think because it is not specific to libfuse but also documents the kernel mount options. Also, the FUSE2 version is quite a mess, it mixes up kernel options with high level and low level options. Maybe it makes more sense to just drop this one from fuse 2 instead...

przemoc:
You're right, I went too far with this renaming `mount.fuse.8` to `mount.fuse3.8`. I'll hopefully fix it tomorrow.

Nikratio:
Thanks! Any specific reason why you excluded the `examples` directory?

Nikratio:
Ops, and the same applies to the renaming of `/etc/fuse.conf`. No need for that, IMO.

przemoc:
I excluded `examples/`, because I'll have to prepare some code fixes there (it's quite late here today, that's why I'm postponing it, sorry). I always create PRs against upstream's master, so just wanted to avoid potential conflicts later (maybe there wouldn't be any, just being on the safe side).

Same configuration file is potentially problematic from packager's point of view (coexisting fuse 2 & 3), but I agree that having one fuse config file seems cleaner.

BTW Is there any stuff in fuse3's `mount.fuse.8` that does not apply to fuse2? If no, then it would be very practical in distributions to stop shipping `mount.fuse.8` from fuse2 and only take it from fuse3, when it will be officially released.

Nikratio:
I think packagers should either ship libfuse2 and libfuse3 in the same package, or create a separate fuse-common package that contains manpage, udev rules, and configuration file.

At the moment, `mount.fuse.8` is valid for both fuse2 and fuse3. If we add new options, we can just add a remark along the lines of "introduced in fuse xx.yy.".

przemoc:
I have updated the patchset. It should match your expectations better.

PR #123, opened 2016-11-27
lib/{mount.c,mount_util.c}: Depend on IGNORE_MTAB instead of __NetBSD__.

przemoc:
Define `IGNORE_MTAB` in these files if `__NetBSD__` is defined.

It makes the code more consistent (same pattern) and also allows
avoiding call to `mtab_needs_update()` (making it a macro returning 0)
if libfuse was configured with `--disable-mtab`.

util/fusermount.c's preprocessor directives also check `IGNORE_MTAB`, but
`__NetBSD__` is nowhere inspected there, and because I don't have NetBSD
and I don't know NetBSD mounting behavior, I refrained from touching
this file and making `-D__NetBSD__` -> `-DIGNORE_MTAB` a global thing.

### DIFF
--- a/lib/mount.c
+++ b/lib/mount.c
@@ -37,8 +37,11 @@
 #define MS_SYNCHRONOUS	MNT_SYNCHRONOUS
 #define MS_NOATIME	MNT_NOATIME
 
+#ifndef IGNORE_MTAB
+#define IGNORE_MTAB 1
+#endif /* IGNORE_MTAB */
 #define umount2(mnt, flags) unmount(mnt, (flags == 2) ? MNT_FORCE : 0)
-#endif
+#endif /* __NetBSD__ */
 
 #define FUSERMOUNT_PROG		"fusermount3"
 #define FUSE_COMMFD_ENV		"_FUSE_COMMFD"
@@ -481,7 +484,6 @@ static int fuse_mount_sys(const char *mnt, struct mount_opts *mo,
 		goto out_close;
 	}
 
-#ifndef __NetBSD__
 #ifndef IGNORE_MTAB
 	if (geteuid() == 0) {
 		char *newmnt = fuse_mnt_resolve_path("fuse", mnt);
@@ -496,7 +498,6 @@ static int fuse_mount_sys(const char *mnt, struct mount_opts *mo,
 			goto out_umount;
 	}
 #endif /* IGNORE_MTAB */
-#endif /* __NetBSD__ */
 	free(type);
 	free(source);
 
--- a/lib/mount_util.c
+++ b/lib/mount_util.c
@@ -33,10 +33,6 @@
 #define IGNORE_MTAB 1
 #endif /* IGNORE_MTAB */
 #define umount2(mnt, flags) unmount(mnt, (flags == 2) ? MNT_FORCE : 0)
-#else /* !__NetBSD__ */
-#ifdef IGNORE_MTAB
-#undef IGNORE_MTAB
-#endif /* IGNORE_MTAB */
 #endif /* __NetBSD__ */
 
 #ifdef IGNORE_MTAB
--- a/lib/mount_util.c
+++ b/lib/mount_util.c
@@ -29,7 +29,17 @@
 #include <sys/param.h>
 
 #ifdef __NetBSD__
+#ifndef IGNORE_MTAB
+#define IGNORE_MTAB 1
+#endif /* IGNORE_MTAB */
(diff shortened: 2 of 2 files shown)


Nikratio:
I don't fully understand this patch. As far as I can tell, there are two unrelated changes:

First, you define that in `mount.c` and `mount_util.c`, using NetBSD should imply `IGNORE_MTAB`. This happens not to change the behavior because `IGNORE_MTAB` isn't used for anything else in these files, but I don't think this makes it a good idea. If NetBSD really should imply `IGNORE_MTAB`, shouldn't this be handled in `configure.ac` instead, so that it applies in *all* source files?

Secondly, you are turning `mtab_needs_update` into a no-op if `IGNORE_MTAB` is defined. This looks like a reasonable change.

przemoc:
> If NetBSD really should imply `IGNORE_MTAB`, shouldn't this be handled in configure.ac instead, so that it applies in all source files?

To be honest, I don't know whether compiling on NetBSD should imply `IGNORE_MTAB` globally. That's why I explained in last paragraph of my commit message, why global implication has not been applied.

In `mount.c` and `mount_util.c` questions when (if we build on NetBSD) and why (if we ignore mtab) were mixed. That's why I found depending solely on `IGNORE_MTAB` in the code part better, because it makes the intent (and the code) clearer.

In case of NetBSD status quo has been preserved, i.e. the compiled code remains the same. I have to admit, though, that I succumbed to idea of having simpler path that also fixes bug in other cases (well, I think it's ultimately a bug that `--disable-mtab` configure option was not respected).

If I split this commit, making the first one the same as current one, but in `mount_util.c` with added `#else` for `__NetBSD__` (i.e. `!defined(__NetBSD__)` case), where `IGNORE_MTAB` becomes undefined if it was already defined (making it clear that `IGNORE_MTAB` has been ignored so far, because such commit will preserve status quo), and add another commit removing the `#else` case, then will it be ok for you?

Nikratio:
Maybe, I'd need to see the code to be sure. It's a little hard to tell from your description.

przemoc:
PR has been updated and rebased on master.

Nikratio:
Are you sure you pushed the right commits? I don't see e.g. an "added `#else`" anywhere.

przemoc:
Yes, I am sure. There are 2 commits in this branch and you apparently looked into changes after applying both of them (manually or maybe via "Files changes" tab on GitHub), so you don't see the change introduced in the first commit. Please always check each commit (and its message), not only the whole series changes.

https://github.com/libfuse/libfuse/pull/123/commits/64a0c9fca2d57ae69faacbdf4b35bced47b31e28
https://github.com/libfuse/libfuse/pull/123/commits/d9e4091e35362936f743deb660a4a2a6606c239a

Nikratio:
Sorry, in that case we misunderstood each other. The patch that I would like to merge was the one that turned `mtab_needs_update` into a no-op if `IGNORE_MTAB` is defined (while keeping it a no-op for `__NetBSD__`). I do not want to locally make `__NetBSD__` imply `IGNORE_MTAB`. If this is a good idea, then it should go into `configure.ac`.

Nikratio:
A quick Google search seems to confirm that NetBSD generally doesn't have `/etc/mtab`, so I think adding this to `configure.ac` would be fine.

przemoc:
It's a fact that in these 2 files mtab is ignored when compiled for NetBSD, so making it clear in the code is better. And to preserve status quo for non-NetBSD case I introduced the `#else` branch in one of files (which exposes kind of bug, that IGNORE_MTAB wasn't respected so far and `mtab_needs_update` wasn't turning into a no-op if you weren't using NetBSD). Only in second patch I'm fixing `mtab_needs_update` to make it a no-op for non-NetBSD case if `IGNORE_MTAB` is defined.

You're maintainer, so you can better tell whether activating `IGNORE_MTAB` code paths for NetBSD in the file I haven't touched, i.e. util/fusermount.c, is fine, but it shouldn't be part of this patchset (because I cannot test it; what I provided changes nothing in NetBSD case, so it's fine).

If you find that that fusermount.c is safe for defining `IGNORE_MTAB` if it's NetBSD, then I suggest introducing `#define IGNORE_MTAB` when `__NetBSD__` is defined (like I did in 2 files) and make change in configure.ac in yet another commit, because then moving it to configure.ac will become a no change for the code.

Nikratio:
I have committed ac25c instead. Thanks for bringing this up!

PR #132, opened 2016-12-16
Revert "Dropped example/null.c"

csabahenk:
This reverts commit d5cdbb94a0650b0a462682cf0a84463ff1513900.

null works completely fine, just the mountpoint should
be a regular file -- so there is no need to dismiss it.

Also fixing up compiler warnings.

### DIFF
--- a/example/.gitignore
+++ b/example/.gitignore
@@ -1,5 +1,6 @@
 /passthrough
 /passthrough_fh
+/null
 /hello
 /hello_ll
 /ioctl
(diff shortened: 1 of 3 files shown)


Nikratio:
Oh, interesting. I didn't know that a file can be a mountpoint. Thanks!

However, before adding this again, can we please document this, so that other's don't fell into the same trap as me? Ideally, `main()` would check if the mountpoint is indeed a file and abort if it's not....

csabahenk:
Yes, I thought of that, first I just didn't bother as I foremost just wanted to see if you support the restoration.

Now added the check.

Nikratio:
Merged, thanks! I also added some documentation, a test case, and removed the use of uninitialized memory.

PR #154, opened 2017-03-28
Do not use magic constant

guraga:


Nikratio:
Thanks for the patch!

PR #159, opened 2017-04-13
Set default options before parsing

tchajed:
Previously `attr_timeout`, `entry_timeout`, and `negative_timeout` from the mount options were ignored because the defaults were set after parsing.


Nikratio:
Thanks!

PR #166, opened 2017-05-11
Add fuse_file_info flags: no_read_buf, not_write_buf

OrenKishon:
Currently the decision whether to use _read_ (_write_) or _read_buf_ (_write_buf_) is done in compilation time. We at Ctera Networks are developing a cloud-cache file system which upon a cache hit opens a spliacable disk file and upon a cache miss opens an a no-fd network stream. 
We would like to have the option to decide whether to use the _\_buf_ functions at run time (on _open_).
This can be done by adding "no_read_buf" and "no_write_buf" flags to _struct fuse_file_info_

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -77,6 +77,10 @@ struct fuse_file_info {
 	/** Requested poll events.  Available in ->poll.  Only set on kernels
 	    which support it.  If unsupported, this field is set to zero. */
 	uint32_t poll_events;
+
+	/** Can be filled in by open to indicate not to use read/write_buf */
+	unsigned int no_read_buf : 1;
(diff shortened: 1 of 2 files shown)


Nikratio:
Why don't you *always* use the `_buf` functions? Just read the network stream into a memory buffer. Using `_buf` doesn't *require* you to provide an fd.

OrenKishon:
Indeed this is our current implementation but it is a duplication of the
code that comes after the checks "if (op.write_buf)... else" etc, which i
see as redundant.

Nikratio:
Sorry, I still do not understand. Could you share the code?

Nikratio:
Closing this for now since there's been no activity for a while - please feel free to re-open if you want to continue working on this.

PR #172, opened 2017-05-24
Make tests build on bsd

bnaylor:
Note that the new build stuff also doesn't work on freebsd (no libdl, at least.)  I don't know anything about meson so I took a pass on that, still using ./configure

### DIFF
--- a/test/test_setattr.c
+++ b/test/test_setattr.c
@@ -20,7 +20,12 @@
 #include <stddef.h>
 #include <unistd.h>
 #include <pthread.h>
+
+#ifndef __linux__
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks! Would you mind filing an issue about the error that you get when running Meson on FreeBSD? libdl is the Linux dynamic library loader, so there is most likely an equivalent thing on BSD..

bnaylor:
Sure, will do.  dlopen is just build into libc on FreeBSD, no need for -ldl, but meson doesn't seem to know that.  Will file an issue.

PR #177, opened 2017-06-04
Fix comparison of integers of different signs

pablomh:
Some variables of different size and sign were getting compared
without any safe casting.
The build system also throws warnings at this and, being this
library used for filesystems, it's really important to ensure
stability.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2399,12 +2399,12 @@ static const char *opname(enum fuse_opcode opcode)
 static int fuse_ll_copy_from_pipe(struct fuse_bufvec *dst,
 				  struct fuse_bufvec *src)
 {
-	int res = fuse_buf_copy(dst, src, 0);
+	ssize_t res = fuse_buf_copy(dst, src, 0);
 	if (res < 0) {
 		fprintf(stderr, "fuse: copy from pipe: %s\n", strerror(-res));
 		return res;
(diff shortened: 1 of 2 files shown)


lib/fuse_lowlevel.c
```diff
@@ -2940,7 +2940,7 @@ int fuse_req_getgroups(fuse_req_t req, int size, gid_t list[])
 		goto out_free;
 	}
 
-	if (ret == bufsize) {
+	if ((size_t)ret == bufsize) {
```
Nikratio: I think to make this safe, the condition above has to be changed from `ret == -1` to `ret < 0`.

Nikratio:
Thanks for the patch! Looks good except for one comment above.

kholk:
You should be right about that, but the patch is made to only solve the obvious problem of comparison of different types. 

In any case, remember that errors on Linux and UNIX-based operating systems are identified with "any" (stated in errno.h) NEGATIVE number so, while the check can be improved for a specific error check case, the current version (ret<0) is just... Safe.

Nikratio:
The current version does not have `ret < 0` but `ret == -1`, that's the problem (line 2938). We shouldn't be casting to an unsigned type (`size_t`) if the value may be negative.

kholk:
That's absolutely right, but please note that I don't know when I'll find time to fix the commit!

PR #181, opened 2017-06-26
Add idle_threads parameter

joe0184:
Add a high level API parameter to configure the number of idle worker threads for op handling.


trapexit:
While I think this is a necessary change there is still an issue with the design. Spawning threads on the fly with a hard reaping behavior is costly. I've seen massive thread churn in highly concurrent situations leading to high load.

I've been experimenting with fixing the number of threads and removing all the mutex usage in the main loop. Simplifies the code, reduces contention (there is no need for the mutexs there), and seems to perform better under load.

That said it's been all anecdotal and I need to build tests to ensure it makes a reasonable difference.

joe0184:
I did notice that this really only answers half of your enhancement to improve threading. As far as the current implementation goes, we noticed when testing our workloads that increasing the number of idle threads to 32 can give us a 30% performance boost. This may be moot since it sounds like you are close to making your changes.

trapexit:
I've made these changes to 2.9.7 (though simple enough to port to 3.x. I don't see any conflict with your changes with what I've done so if @Nikratio is happy with them I think its fine. I need to confirm what I've done makes a difference before trying to upstream them. I also need to port my fs to the 3.x API at some point.

Nikratio:
This looks good to me. However, you still need to add code to preserve backwards compatibility. If you change the definition of *fuse_loop_mt()*, programs that were compiled against earlier versions of libfuse will crash when they are linked against the more recent one. And attempts to compile against the more recent version will fail. You can take a look at the recent introduction of *fuse_lib_help* for ideas of how to do that. Note, however, that there is still a problem with the symbol redefinition (see comments on the commit).

Without backwards compatibility, this would have to wait until libfuse4.

joe0184:
Would it work to use a new max_idle_threads global variable instead of changing the function definition to pass the value? On one hand, it seems to depart from the project's coding style. On the other, it may reduce complexity by avoiding symbol versioning. Perhaps there are other implications that I'm not thinking of?

Nikratio:
Yeah, it would work, and yeah, that would be a pretty terrible solution.  Let's do it the right away.

As a matter of fact, let's make sure the problem does not re-appear in the future. Instead of adding a new parameter, let's introduce a `struct fuse_loop_config` that starts out with `clone_fd` and `max_idle_threads` fields and pass that instead of `clone_fd`. That still requires symbol versioning, but future extensions can than be added easily. (I can see a `thread_init_fn` function on the horizon that would be called whenever a new thread is created).

joe0184:
I added the new struct and version of fuse_session_loop_mt, but I did not change the libfuse minor version number.

include/fuse_lowlevel.h
```diff
@@ -1826,11 +1844,15 @@ int fuse_session_loop(struct fuse_session *se);
  * fuse_session_loop().
  *
  * @param se the session
- * @param clone_fd whether to use separate device fds for each thread
- *                 (may increase performance)
+ * @param config additional configuration parameters
```
Nikratio: "additional" is a little confusing. In addition to what? Just call it "session loop configuration".

lib/fuse.c
```diff
@@ -4421,6 +4421,7 @@ static const struct fuse_opt fuse_lib_opts[] = {
 	FUSE_LIB_OPT("noforget",              remember, -1),
 	FUSE_LIB_OPT("remember=%u",           remember, 0),
 	FUSE_LIB_OPT("modules=%s",	      modules, 0),
+	FUSE_LIB_OPT("idle_threads=%u",   max_idle_threads, 0),
```
Nikratio: I think the configuration option should also include `max_`.

include/fuse_lowlevel.h
```diff
@@ -1826,11 +1844,15 @@ int fuse_session_loop(struct fuse_session *se);
...
  * @return see fuse_session_loop()
  */
-int fuse_session_loop_mt(struct fuse_session *se, int clone_fd);
+#if FUSE_USE_VERSION <= 31
```
Nikratio: libfuse 3.1.0 has already been released, so this needs to be `<= 32`.

lib/fuse.c
```diff
@@ -4467,7 +4468,8 @@ void fuse_lib_help(struct fuse_args *args)
 "    -o ac_attr_timeout=T   auto cache timeout for attributes (attr_timeout)\n"
 "    -o noforget            never forget cached inodes\n"
 "    -o remember=T          remember cached inodes for T seconds (0s)\n"
-"    -o modules=M1[:M2...]  names of modules to push onto filesystem stack\n");
+"    -o modules=M1[:M2...]  names of modules to push onto filesystem stack\n"
+"    -o idle_threads=N      maximum number of idle threads allowed\n");
```
Nikratio: corresponding change here (`max_`)

lib/fuse.c
```diff
@@ -4702,6 +4704,11 @@ struct fuse *fuse_new_31(struct fuse_args *args,
 		}
 	}
 
+
+	if (f->conf.max_idle_threads == 0) {
+		f->conf.max_idle_threads = 10;
+	}
+
```
Nikratio: What if someone does not want any idle threads? At the very least, the special meaning of zero has to be documented.

lib/fuse_loop_mt.c
```diff
@@ -300,18 +301,21 @@ static void fuse_join_worker(struct fuse_mt *mt, struct fuse_worker *w)
 	free(w);
 }
 
-int fuse_session_loop_mt(struct fuse_session *se, int clone_fd)
+int fuse_session_loop_mt_config(struct fuse_session *se, struct fuse_loop_config *config);
+FUSE_SYMVER(".symver fuse_session_loop_mt_config,fuse_session_loop_mt@@FUSE_3.2");
+int fuse_session_loop_mt_config(struct fuse_session *se, struct fuse_loop_config *config)
```
Nikratio: Please use `_fuse32` and `fuse31` suffixes instead of `_config`. That keeps it consistent with the naming of `fuse_new()`.

include/fuse_lowlevel.h
joe0184: Shouldn't it either be "<= 31" or "< 32"?

lib/fuse.c
```diff
@@ -4421,6 +4421,7 @@ static const struct fuse_opt fuse_lib_opts[] = {
 	FUSE_LIB_OPT("noforget",              remember, -1),
 	FUSE_LIB_OPT("remember=%u",           remember, 0),
 	FUSE_LIB_OPT("modules=%s",	      modules, 0),
+	FUSE_LIB_OPT("max_idle_threads=%u",   max_idle_threads, 0),
```
Nikratio: You have defined `max_idle_threads` as `int`, so that this should be `%d` rather than `%u` (or you could declare `max_idle_threads` as `unsigned`, which may make more sense).

lib/fuse.c
```diff
@@ -4702,6 +4705,12 @@ struct fuse *fuse_new_31(struct fuse_args *args,
 		}
 	}
 
+
+	if (f->conf.max_idle_threads == -1) {
+		/* If the number of idle threads was not set, default to 10. */
+		f->conf.max_idle_threads = 10;
+	}
+
```
Nikratio: Why set it to -1, only to later set it to 10? Can't you just initialize it with the default value of 10 (and avoid the need to support negative values)?

include/fuse_lowlevel.h
```diff
@@ -1826,11 +1844,15 @@ int fuse_session_loop(struct fuse_session *se);
  * fuse_session_loop().
  *
  * @param se the session
- * @param clone_fd whether to use separate device fds for each thread
- *                 (may increase performance)
+ * @param config session loop configuration 
  * @return see fuse_session_loop()
  */
-int fuse_session_loop_mt(struct fuse_session *se, int clone_fd);
+#if FUSE_USE_VERSION < 32
```
Nikratio: You are right, what you had here before was correct as well. Sorry about that.

lib/fuse.c
joe0184: I was thinking about the comment you made the other day, that a user may not want any idle threads. Here I was hoping to keep the default 10 threads when the user does not specify max_idle_threads, but also allow the user to specify "max_idle_threads=0".

Nikratio: Wouldn't you achieve exactly that if you set the default to 10 when initializing the structure (before the command line is parsed), and get rid of this `if` clause completely?

trapexit: Setting the value to 0 should probably not be allowed. At least warned about. I'm not even sure this dynamic behavior should be done at all. Not without smarter logic around culling unneeded threads. Spawning a thread on demand is costly. This isn't a Erlang or some other light weight process. I've seen my system thrash due to in part to libfuse rapidly spawning new threads and then killing them as soon as they finished with their task. In my branch of libfuse I've completely removed that logic and made a fixed pool that is user configurable and defaults to the number of cores. By doing so it removes all those mutexes and simplifies the code quite a bit.

In my tests I also found that N - 1 threads led to the best performance in general. Both in situations where there was one app hammering the FS and where there were several. (N = number of physical cores, not logical).

I'm all for this change to enable setting the number of threads but it should be well documented how the threads come into and go out of existence.

Nikratio:
Looks good to me now! I agree with @trapexit  that more documentation would be better, but I still think that this is an improvement over the status quo. So the only remaining issue is to fix the conflicts with the master branch, so that I can actually merge :-P.

Nikratio:
Working on the merge myself now...

Nikratio:
While attempting to merge, I found some additional problems, sorry:

* Currently, the new `max_idle` field in `struct fuse_config` is never actually used (only set).
* The calls to `fuse_session_loop_mt()` in `fuse.c` still use the `clone_fd` parameter. 
* The `fuse_loop()` function should also be changed to take a `struct loop_config` instead of `int clonefd`
* Please add a ChangeLog entry and update the fuse version

lib/fuse.c
```diff
@@ -4421,6 +4433,7 @@ static const struct fuse_opt fuse_lib_opts[] = {
 	FUSE_LIB_OPT("noforget",              remember, -1),
 	FUSE_LIB_OPT("remember=%u",           remember, 0),
 	FUSE_LIB_OPT("modules=%s",	      modules, 0),
+	FUSE_LIB_OPT("max_idle_threads=%u",   max_idle_threads, 0),
```
Nikratio: As above, `clone_fd` and `max_idle_threads` should be on the same footing. So either add `clone_fd` here, or remove the `max_idle_threads` setting.

lib/fuse.c
```diff
@@ -4870,3 +4885,8 @@ const char *fuse_pkgversion(void)
 {
 	return PACKAGE_VERSION;
 }
+
+struct fuse_config *fuse_get_config(struct fuse *f)
```
Nikratio: This seems to be used only by `helper.c`. Why can't `helper.c` include the right header to access `struct fuse`  directly?

lib/helper.c
```diff
@@ -326,8 +326,13 @@ int fuse_main_real(int argc, char *argv[], const struct fuse_operations *op,
...
 	if (opts.singlethread)
 		res = fuse_loop(fuse);
-	else
-		res = fuse_loop_mt(fuse, opts.clone_fd);
+	else {
+		struct fuse_loop_config loop_config;
+		struct fuse_config *config = fuse_get_config(fuse);
+		loop_config.clone_fd = opts.clone_fd;
+		loop_config.max_idle_threads = config->max_idle_threads;
+		res = fuse_loop_mt(fuse, &loop_config);
```
Nikratio: This is ugly. `fuse` already includes the value for `max_idle_threads`, so we're effectively passing it twice. This is the result of treating `clone_fd` and `max_idle_threads` differently, and yet another reason to put them on the same footing :-).

joe0184:
Sorry that it took so long for me to get around to this. It turns out that `max_idle_threads` was being set to 0 and I guess that was causing the test to slow down to the point of failure. Is this still something we want users to be allowed to do? I could change `fuse_parse_cmdline` to allow it again. Otherwise, where would be the best place to document why you can't do it?

lib/helper.c
```diff
@@ -326,8 +329,18 @@ int fuse_main_real(int argc, char *argv[], const struct fuse_operations *op,
 
 	if (opts.singlethread)
 		res = fuse_loop(fuse);
-	else
-		res = fuse_loop_mt(fuse, opts.clone_fd);
+	else {
+		struct fuse_loop_config loop_config;
+		loop_config.clone_fd = opts.clone_fd;
+		if (opts.max_idle_threads == 0) {
```
Nikratio: Just assign the default *prior* to calling `fuse_parse_cmdline` (line 279), then you don't need the `if` condition. It actually looks like there is a bug there already, in that `opts` is not explicitly zerod. It should be:

```
 	struct fuse_cmdline_opts opts;
 	int res;
 
        memset(&opts, 0, sizeof(opts));
        opts.max_idle_threads = 10;        
 	if (fuse_parse_cmdline(&args, &opts) != 0)
return 1;
```

Nikratio:
I think you should still allow the user to do it (if only becomes it actually makes the code cleaner, see my other comment). The right place to warn about the performance implications is In the description of `struct fuse_loop_config` (in `fuse_common.h`).

Thanks for working on this!

lib/helper.c
joe0184: `fuse_parse_cmdline` does perform a memset of the struct. I can move that line out of that function, however anyone who relies on it to clear opts (such as the examples) will probably need to change as well.

lib/helper.c
```diff
@@ -132,7 +133,9 @@ void fuse_cmdline_help(void)
 	       "    -f                     foreground operation\n"
 	       "    -s                     disable multi-threaded operation\n"
 	       "    -o clone_fd            use separate fuse device fd for each thread\n"
-	       "                           (may improve performance)\n");
+	       "                           (may improve performance)\n"
+		   "    -o max_idle_threads    the maximum number of idle worker threads\n"
+		   "                           allowed\n");
```
Nikratio: `... "(default: 10)"`

lib/helper.c
Nikratio: Ah, sorry. In that case simply move the default setting into `fuse_parse_cmdline`, after the `memset`.

Nikratio:
Sorry that this has been taking so long. But with all the merges and separate commits, it's been very difficult to review. I just squashed and rebased everything on master. This should make things easier.

PR #187, opened 2017-07-28
Ask pkgconfig where we should install udev rules

heirecka:


Nikratio:
Why do you think this is better? On Debian, *libdir* is `usr/lib`, but udev rules are supposed to go in `/lib/`.

heirecka:
All those small differences between distros ;-)

Guess I could ask udev about udevdir via pkg-config to make everybody happy.

Nikratio:
Is there a way to use pkg-config? That would be great.

Nikratio:
Thanks!

PR #189, opened 2017-08-03
Freebsd

bapt:
Upstream FreeBSD changes and fix build with meson

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -92,7 +92,10 @@ thread_dep = dependency('threads')
 #
 # Read build files from sub-directories
 #
-subdirs = [ 'lib', 'include', 'util', 'example', 'doc', 'test' ]
+subdirs = [ 'lib', 'include', 'example', 'doc', 'test' ]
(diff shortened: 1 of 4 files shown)


Nikratio:
Wow, thanks a lot! Much appreciated.

Nikratio:
...where "this patch" is the "Support fsname= option" commit.

meson.build
```diff
@@ -92,7 +92,10 @@ thread_dep = dependency('threads')
 #
 # Read build files from sub-directories
 #
-subdirs = [ 'lib', 'include', 'util', 'example', 'doc', 'test' ]
+subdirs = [ 'lib', 'include', 'example', 'doc', 'test' ]
+if not host_machine.system().startswith('freebsd')
```
Nikratio: Are you sure this works? According to http://mesonbuild.com/Reference-manual.html#build_machine-object the test should be `host_machine.system() == 'bsd'`.

PR #190, opened 2017-08-03
Fix support for FUSE_POSIX_ACL

marcinsulikowski:
I tried to use the POSIX ACL feature which has been included in Linux kernels ≥ 4.9 but it did not work for me. After looking into the libfuse and Linux kernel sources I concluded that the kernel expects the `FUSE_POSIX_ACL` to be included when replying the initial `FUSE_INIT` request and the library never sets it. The change included in the attached commit fixes the problem for me.

However, maybe I did something wrong? I can see there are a few other flags which are never set by libfuse in `outarg.flags` in `do_init` (e.g., `FUSE_HANDLE_KILLPRIV`). Is the lack of `FUSE_POSIX_ACL` in `outarg.flags` an oversight in the library or maybe there is something else that file system implementations have to do to enable ACLs? How about the other flags?

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,10 @@
+Unreleased Changes
+==================
+
+* Fixed support for `FUSE_CAP_POSIX_ACL`: setting this capability
+  flag had no effect in the previous versions of libfuse 3.x;
+  now ACLs should actually work.
+
 libfuse 3.1.1 (2017-08-06)
 ==========================
 
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -1981,6 +1981,8 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 		outarg.flags |= FUSE_ASYNC_DIO;
 	if (se->conn.want & FUSE_CAP_WRITEBACK_CACHE)
 		outarg.flags |= FUSE_WRITEBACK_CACHE;
+	if (se->conn.want & FUSE_CAP_POSIX_ACL)
+		outarg.flags |= FUSE_POSIX_ACL;
 	outarg.max_readahead = se->conn.max_readahead;
 	outarg.max_write = se->conn.max_write;
 	if (se->conn.proto_minor >= 13) {


Nikratio:
Thanks for the patch! No, the lack of `FUSE_POSIX_ACL` in libfuse is not a design choice. libfuse has not yet been updated to support ACLs. But it looks as if this is about to change :-).

I'll try to review this soon.

Nikratio:
Alright, that patch is a little shorter than expected :-).

We do not need to add support for this unconditionally. You need to add a capability flag that the filesystem can test for (and set) in its init method (similarly to eg. `FUSE_CAP_WRITEBACK_CACHE`). The documentation for the capability flag should also clearly state what is expected of filesystems that enable this feature. It has been a while since I looked at the corresponding kernel patch, but I believe among other things it requires the file system to support getxattr and setxattr. Where possible, libfuse should also check for these requirements programmatically (and e.g. abort if the filesystems claims to support ACL's but doesn't define xattr handlers).

marcinsulikowski:
> libfuse has not yet been updated to support ACLs

> We do not need to add support for this unconditionally. You need to add a capability flag that the filesystem can test for (and set) in its init method (similarly to eg. FUSE_CAP_WRITEBACK_CACHE). The documentation for the capability flag should also clearly state what is expected of filesystems that enable this feature.

Haven't you already add this flag in e16fdc06d7473f00499b6b03fb7bd06259a22135 and 4fdf83f4dd2b0c3a0688d23de23ad953abf473ac? Those commits seem to add the required `FUSE_CAP_POSIX_ACL` flag and document it. The only issue is that even if the filesystem implementation sets the capability flag in its `init` method, the corresponding feature flag is not passed to the kernel and this is what my commit tries to fix.

Nikratio:
I'm getting old, yes, it seems I already added most of what's required. In that case, please just add a ChangeLog entry that documents that only now ACLs will actually work.

marcinsulikowski:
I updated the changelog file in my commit and also rebased the commit to the latest `master` branch to avoid conflicts with the changes in the changelog file related to the `fuse-3.1.1` release.

Nikratio:
Thanks!

PR #192, opened 2017-08-04
Allow inode cache invalidation in high-level API

saf:
> We introduce a new routine to the high-level FUSE API,
> fuse_notify_inval_fs_inode, which is responsible for invalidating the
> cached data for an inode.
> 
> The routine uses a new mapping of the "FS inodes" (inode numbers
> returned in the 'st_ino' field from 'getattr') into the nodes in the
> high-level FUSE API. Naturally, because multiple paths may refer to
> the same inode on the file system due to hard links, the mapping is
> one-to-many, and is implemented by means of a hash table. Internally,
> the high-level API routine calls the low-level routine
> fuse_notify_inval_inode.
> 
> This is useful for network-based file systems which use the high-level
> FUSE API. Up to now, there was no way for those file systems to notify
> the kernel about external changes.

I know this PR is not in perfect shape now (it could be using dynamically-sized hash tables, and it's based on fuse_2_9_bugfix), but before fixing and rebasing it, I'd like to ask for feedback from the maintainers on whether this would be an acceptable solution for providing external change notifications from file systems which use the high-level FUSE API.

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -794,6 +794,19 @@ void fuse_stop_cleanup_thread(struct fuse *fuse);
  */
 int fuse_clean_cache(struct fuse *fuse);
 
+/**
+ * Notify the kernel to invalidate cache for the specified FS inode.
+ *
+ * This can be used to notify the kernel about external changes
+ * to files and directories, for example in network file systems.
+ *
+ * This uses a one-to-many mapping from the inode numbers reported
+ * by the file system (in the `st_ino` field in `getattr` output)
+ * and FUSE nodes. This calls `fuse_lowlevel_notify_inval_inode`
+ * internally.
+ */
+int fuse_notify_inval_fs_inode(struct fuse *fuse, ino_t fs_inode);
+
 /*
  * Stacking API
  */
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -51,6 +51,7 @@
 #define OFFSET_MAX 0x7fffffffffffffffLL
 
 #define NODE_TABLE_MIN_SIZE 8192
+#define FS_INODE_HASH_SIZE (64 * 1024)
 
 struct fuse_config {
 	unsigned int uid;
@@ -152,6 +153,7 @@ struct fuse {
 	int pagesize;
 	struct list_head partial_slabs;
 	struct list_head full_slabs;
+	struct list_head fs_inode_hash[FS_INODE_HASH_SIZE];
 	pthread_t prune_thread;
 };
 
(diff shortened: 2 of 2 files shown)


Nikratio:
Thanks for the pull request!

My first instictive reaction is:
* Having cache invalidation in the high-level API would be great
* It should probably be based on path names rather than inodes (or at least offer pathname based invalidation as an option), because the use of path names is about 50% of the high-level API. If you only use pathnames, you could also use the existing pathname-inode hash table and would not need to introduce another one.

I'll look into it in more detail when I have more time. But maybe you can already ponder the above :-).

include/fuse.h
```diff
@@ -794,6 +794,19 @@ void fuse_stop_cleanup_thread(struct fuse *fuse);
...
+ *
+ * This can be used to notify the kernel about external changes
+ * to files and directories, for example in network file systems.
+ *
+ * This uses a one-to-many mapping from the inode numbers reported
+ * by the file system (in the `st_ino` field in `getattr` output)
+ * and FUSE nodes. This calls `fuse_lowlevel_notify_inval_inode`
+ * internally.
+ */
+int fuse_notify_inval_fs_inode(struct fuse *fuse, ino_t fs_inode);
```
Nikratio: The name seems odd. Why the "fs" in the middle? I think `fuse_session_notify_inval_inode` would be more consistent with the existing naming scheme.

saf:
Thank you for your feedback. The changes used the "FS inode" rather than the paths because that's what our network-based file system's server talks. It's probably OK, just a bit less convenient, for it to supply paths rather than inode numbers.

So, I can try my hand at reworking the solution. As you pointed out, all the necessary pieces are already there. Actually, I found code which used to be the `fuse_invalidate` in an old commit by Miklos (`891b874eedc`); how would you feel about just reviving that (in terms of implementation and naming)?

trapexit:
It'd be convenient to have the high level API offer both the node id & generation id as well as the path. There have been a few situations where keying off that value would have been nicer (unfortunately it's too early in the morning for me to remember the situation.)

Nikratio:
I don't think this code will be useful, it hasn't aged well. The implementation should really call the `fuse_lowlevel_notify` function, and I think the name should be similar.

saf:
Yes, @Nikratio, I meant reviving that code _and_ adapting it to the modern codebase, of course. It turned out to be pretty neat, but let me do some testing before I update the PR. 

@trapexit, can you please elaborate on the scenarios where `nodeid`/`generation` would be useful for invalidation in high-level API? Aren't file systems which use the high-level API generally unaware of the node IDs, let alone node generations? Thanks.

trapexit:
I wasn't referring to invalidation. I'm speaking more generally. I would like if the API gave all information it held to the user. {node,generation} and internal inode (when not using use_ino) come to mind. Knowing {node,generation} is unique, for example, can make it more convenient as a key in certain data structures should you want to use the high level API and let it do the lower level node management but still want to efficiently reference those nodes.

saf:
I'm going to close this PR in favor of another one which I'm preparing which will be based on the `master` branch and answer @Nikratio's remarks.

PR #195, opened 2017-08-08
Allow inode cache invalidation in high-level API

saf:
We re-introduce the functionality of invalidating the caches for an
inode specified by path by adding a new routine
fuse_invalidate_path. This is useful for network-based file systems
which use the high-level API, enabling them to notify the kernel about
external changes.

This is a revival of Miklos Szeredi's original code for the
fuse_invalidate routine.

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -1004,6 +1004,19 @@ int fuse_getgroups(int size, gid_t list[]);
 int fuse_interrupted(void);
 
 /**
+ * Invalidates cache for the given path.
+ *
+ * This calls fuse_lowlevel_notify_inval_inode internally.
+ *
+ * @return 0 on successful invalidation, negative error value otherwise.
+ *         This routine may return -ENOENT to indicate that there was
+ *         no entry to be invalidated, e.g., because the path has not
+ *         been seen before or has been forgotten; this should not be
+ *         considered to be an error.
+ */
+int fuse_invalidate_path(struct fuse *f, const char *path);
+
+/**
  * The real main function
  *
  * Do not call this directly, use fuse_main()
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,6 +1,8 @@
 Unreleased Changes
 ==================
 
+* Added new `fuse_invalidate_path()` routine for cache invalidation
+  from the high-level FUSE API, along with an example and tests.
 * Fixed support for `FUSE_CAP_POSIX_ACL`: setting this capability
   flag had no effect in the previous versions of libfuse 3.x;
   now ACLs should actually work.
--- a/example/.gitignore
+++ b/example/.gitignore
@@ -11,5 +11,6 @@
 /cuse_client
 /passthrough_ll
 /notify_inval_inode
+/notify_inval_inode_fh
 /notify_store_retrieve
 /notify_inval_entry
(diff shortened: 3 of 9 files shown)


include/fuse.h
```diff
@@ -1004,6 +1004,17 @@ int fuse_getgroups(int size, gid_t list[]);
 int fuse_interrupted(void);
 
 /**
+ * Invalidates cache for the given path.
+ *
+ * This calls fuse_lowlevel_notify_inval_inode internally.
+ * This only does something for nodes which are present in
+ * the hashes; it is a no-op for paths to forgotten nodes.
+ *
+ * @return 0 on success, negative error value otherwise.
```
Nikratio: Why not return -ENOENT if the node isn't present? Returning zero instead deliberately removes information that might be useful.

saf: I followed the original code by Miklos here, and I think the high-level file system does not need to be concerned with nodes being forgotten by FUSE, and so the invalidating routine should not return an error if the node to be is not present in the hashes. If I understand it correctly, if a node is not present in the caches (and FUSE hashes), the operating system does not care about the caches being invalidated. Please note that `lookup_path_in_cache` never does an actual lookup with the user-space file system, so returning `ENOENT` from here could actually be misleading.

An example situation to show this might be where in a network file system with a centralized server, the server notifies the client (which maps the server storage into a FUSE-based file system) about a change in an inode which the client has not "seen" yet, and hence it's not present in the caches & hashes. My feeling is that there should be no error for the client in that case if it passes the notification as a cache invalidation request to `libfuse`.

trapexit: Why would you keep possibly useful information from the caller? Especially when there is basically zero cost to do so?

It's not an error. It's an indication that there was no cached value to invalidate. The docs should clearly indicate that. If there are true error conditions that should be pointed out as well.

IMO the high level API already hides too much.

saf: I'm still not seeing why the info whether there was a cache entry to be invalidated would be useful. From the point of view of our file system's client, it would make calling `fuse_invalidate_path` more cumbersome because on an error it would have to check and ignore `ENOENT` errors (or whatever we choose the error to be; I still think `ENOENT` would be misleading, perhaps a more obscure errno, such as `ENOKEY`, would be in order). I see your point about hiding info, @trapexit, but IMHO we should favor making the API easier to use in this particular case.

Or should we go for yet another solution and return `1` if there was an entry to invalidate, `0` if there was not, and `-1` on errors? This would make the usage a bit more convenient in that checking for errors would only require checking `fuse_invalidate_path(...) < 0` or `!= -1`.

@Nikratio, what are your thoughts on this?

Nikratio: I do not have an example of what this information would be useful for either. And I concede your point about the call becoming more complicated.

However, I am concerned about the time when someone comes up with a use-case that would benefit from having this information. At that point, the API cannot be changed without breaking backwards compatibility. Therefore, I would like to see the ENOENT returned as-is.
People who would like to ignore this error can then trivially write a wrapper function. If we did not return ENOENT, there would be no recourse for people who want it.

From that point of view, reducing the return value to just `(1,-1,0)` would be even worse.

saf:
I have posted another commit which makes `fuse_invalidate_path` return `-ENOENT` if the entry does not exist as per your advice. Please let me know if you have any other remarks. Should I squash the commits (they are meant to be squashed) or is that done when accepting the PR?

Nikratio:
Thanks! I can squash the commits on merge. However, I don't actually see any test case. Are you referring to the new file in the `examples` directory? If so, you need to make sure that the unit tests actually run that example. Take a look at `test/test_examples.py` to see how this is done.

saf:
Oh, sorry, the modified `test_ctests.py` got lost in my `build` folder. I now updated that and also rebased & squashed the commits because there were conflicts with some recent changes to `master`.

Nikratio:
Merged, thanks for your contribution!

Issue #198, opened 2017-08-12
Support link-time optimization (-lto)


userwithuid:
To reproduce, simply use the meson option `-Db_lto=true`.

3.1.0 works fine.
3.1.1 or master fail (due to I think https://github.com/libfuse/libfuse/commit/503e32d01e4db00e90d7acfd81ab05386559069f and all the follow ups).
```
/tmp/ccDe4RMB.ltrans0.ltrans.o: In function `fuse_main_real':
<artificial>:(.text+0x1694a): undefined reference to `fuse_new'
```

Nikratio:
Thanks for the report! I'm afraid unless someone can provide a fix that doesn't affect buildability with the compilers mentioned in the commit discussion, this will stay unfixed. I rather have libfuse compile on various systems without lto than on just a few systems with lto.

Expert help welcome!

userwithuid:
@Nikratio So this should probably reopened since https://github.com/libfuse/libfuse/commit/f12d9686d4d673e59b8f561c49996820763220b1 broke lto again...

Using the versioned functions everywhere as a workaround like we did earlier seems kinda ugly, there's gotta be a better way. I don't know anything about this symver stuff though, so I guess it's best to just wait for someone that does. :-)

Nikratio:
I've committed a partial fix. However, now the problem is that with lto, the `libfuse.so` library is missing symbols. Without lto, we have:

```
$ nm lib/libfuse3.so | grep fuse_new
0000000000011070 T fuse_new_30
0000000000010a00 t fuse_new_31
0000000000011070 T fuse_new@FUSE_3.0
0000000000010a00 T fuse_new@@FUSE_3.1
```

and with LTO:

```
$ nm lib/libfuse3.so | grep fuse_new
0000000000019a70 T fuse_new_30
0000000000019270 t fuse_new_31
```

Nikratio:
Let's see if the internet can help:

https://stackoverflow.com/questions/46304742/compiling-with-lto-hides-renamed-symbols

trofi:
Filed https://gcc.gnu.org/PR91186: gcc occasionally misses `asm .symver` entries when partitioning LTO work.

kloczek:
Looks like 3.10.0 with latest gcc builds with LTO.
IMO this ticket can be closed now.

Nikratio:
Ok, thanks for checking & reporting!

PR #200, opened 2017-08-14
fix lto build

userwithuid:
attempt at a workaround for https://github.com/libfuse/libfuse/issues/198

same as what you did [here](https://github.com/libfuse/libfuse/commit/5a5edc3353f468600933112422f080004189a665) for `fuse.c`, but you ignored the same call in `helper.c`.

builds on your gcc 4.8 travis. also, on my arch setup with gcc 7.1, this builds and runs fuse3+sshfs with -flto.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -9,6 +9,8 @@ Unreleased Changes
 
 * Fixed installation directory for udev rules.  
 
+* Fixed compilation with LTO.
+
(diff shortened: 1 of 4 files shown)


Nikratio:
Makes sense, thanks for looking into this!

Could you please add a ChangeLog entry though?

userwithuid:
changelog updated

PR #210, opened 2017-09-27
Adding pointer dereferencing after calling dlsym()

mununum:
dlsym() resolves the location of the loaded symbol,
therefore dlsym() returns the type (fuse_module_factory_t *), not (fuse_module_factory_t).
Added pointer dereferencing to correctly refer the factory function.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -263,7 +263,7 @@ static int fuse_load_so_module(const char *module)
 	}
 
 	sprintf(tmp, "fuse_module_%s_factory", module);
-	factory = dlsym(so->handle, tmp);
+	factory = *((fuse_module_factory_t *) dlsym(so->handle, tmp));
 	if (factory == NULL) {
 		fprintf(stderr, "fuse: symbol <%s> not found in module: %s\n",
 			tmp, dlerror());


Nikratio:
Thanks for the patch! Am I right that this just silences some (which?) compiler warning? Or are there any practical problems fixed by this? In that case it would be great to have a test case.

mununum:
This issue is not the one of some compiler warning issues,
and there can be practical problems with this issue.
In order to test this issue, I first bypassed out builtin module registration in fuse.c
to execute the dlopen() routine when loading subdir module.

from fuse.c:4712 in master branch
```
    pthread_mutex_lock(&fuse_context_lock);
#if 0
    static int builtin_modules_registered = 0;
    /* Have the builtin modules already been registered? */
    if (builtin_modules_registered == 0) {
        /* If not, register them. */
        fuse_register_module("subdir", fuse_module_subdir_factory, NULL);
#ifdef HAVE_ICONV
        fuse_register_module("iconv", fuse_module_iconv_factory, NULL);
#endif
        builtin_modules_registered= 1;
    }
#endif
    pthread_mutex_unlock(&fuse_context_lock);
```

Then I have built libfusemod_subdir.so by compiling modules/subdir.c with -fPIC -shared option.

I tested basic FUSE implementation, "hello" by entering following command:
`$ ./hello -o modules="subdir,subdir=/subdir" /mnt/fuse`
Without this issue resolved, this execution crashes with segfault.

Nikratio:
Amazing. So this has been broken since 2013. I am almost tempted to remove the entire loadable module support, since apparently no one is using it. But I guess that wouldn't be fair to you, since I assume you are planning to use it now?

PR #215, opened 2017-10-19
make udevrulesdir configurable

Mic92:
on nixos we install fuse in its own hierarchy independent from systemd.

cc @primeos


Nikratio:
Thanks!

PR #216, opened 2017-10-20
Add example configuration file

admorgan:
Add a configuration file with all options disabled that includes
all valid options and their description. This file is called out
in the packaging notes to be included in the -common package


Nikratio:
Thanks!

Issue #211, opened 2017-09-27
Strip `nofail` option in mount.fuse helper


jsoref:
Please excuse my ignorance.

I hit https://github.com/systemd/systemd/issues/5032 and they said it isn't their fault.

Here's the original bug report (by @vicinus):

### Expected behaviour you didn't see

> I want to use the nofail option in /etc/fstab to ensure that the boot up process is finished and no emergency mode is entered on trying to mount some fuse.bindfs mounts during boot up.

> It should be possible to use the nofail option in /etc/fstab with all types of filesystems.

### Unexpected behaviour you saw

> The Problem is, that the nofail option is passed by the mount command line tool to the mount tool specific to the filesystem type and fuse.bindfs based filesystems do not support the nofail option and therefore fail, which is rater ironically...

> One possible solution would be to add x-systemd.nofail and x-systemd.fail with the behaviour to override an eventually existing nofail and fail option. As the new options are systemd specific they will not be passed to filesystem type specific mount tool and can therefore be safely used filesystem type specific mount tool does not support nofail.
In case of bug report: Steps to reproduce the problem

> Install bindfs and add a line like:

> `/usr /mnt fuse.bindfs defaults,nofail 0 0`

> Try to mount it and the error message is:

> `fuse: unknown option 'nofail'`

> Without the option nofail the mount works fine.

### Below are details from my system

```
Ubuntu 16.04.3 LTS
ii  libsystemd0:amd64                  229-4ubuntu19          amd64                  systemd utility library
ii  systemd                            229-4ubuntu19          amd64                  system and service manager
ii  systemd-sysv                       229-4ubuntu19          amd64                  system and service manager - SysV links
ii  fuse                               2.9.4-1ubuntu3.1       amd64                  Filesystem in Userspace
ii  libfuse2:amd64                     2.9.4-1ubuntu3.1       amd64                  Filesystem in Userspace (library)
ii  sshfs                              2.5-1ubuntu1           amd64                  filesystem client based on SSH File Transfer Protocol
```

```fstab
sshfs#user@host:/usr/something/ /usr/something fuse defaults,noauto,allow_other,idmap=user,reconnect,ServerAliveInterval=15,ServerAliveCountMax=3,IdentityFile=~someone/.ssh/id_rsa,x-systemd.requires=/usr/prereq,nofail,x-systemd.automount
```

### System console output
```console
[   12.614072] systemd[1]: usr-something.automount: Failed to check directory /usr/something: No such file or directory
[...]

PR #221, opened 2017-11-23
Naive attempt at handling nofail

jsoref:
issue #211 

I have no idea if this will work. But is this the right file/direction?


Nikratio:
That really depends on what you are trying to achieve. Maybe you could elaborate on your goals?

Nikratio:
Sorry, just saw the issue reference. No, this is the wrong way to do it. As the issue says, this needs to be filtered out in the mount helper.

jsoref:
@Nikratio: where is that?

Nikratio:
https://github.com/libfuse/libfuse/blob/master/util/mount.fuse.c#L155

jsoref:
@nikratio: how's this?

Nikratio:
Doesn't look right to me. Why do you think the presence of `nofail` should affect the exit code?

jsoref:
Sorry. The documentation for this stuff is more or less vaporous. `nofail` definitely affects *someone*'s handling of an error status. But, I think you're right, it's just the caller's understanding of the error status, not the callee.

I'll try to post the obvious/simpler fix (closer to my original, but for this file) tonight.

jsoref:
@Nikratio : how about this? :)

jsoref:
@Nikratio : done

ShapeShifter499:
Is this fix good for all of fuse or do I have to contact the dev behind megerfs as well?

@trapexit Pinging you here since I see this trying to use mergerfs in fstab with the 'nofail' option

PR #232, opened 2018-02-04
Adding possibility enforce a prefix

thopiekar:
In most scenarios, the script is installing to the default locations, where eg. device nodes or udev rules are normally placed. However, since I want to install libfuse to a prefix, which is built with user permissions, I don't have the permissions to place files there.
Therefore I added a new variable, which forces to place the files into a custom prefix.
Of course, there is also DESTDIR, but in my case I don't want to use it, because (if prefix == destdir) I would get `/my/custom/prefix/path/my/custom/prefix/path` as the location for all the other files.

Any other solutions and thoughts are welcome!


Nikratio:
Can't you use DESTDIR and leave PREFIX unset to get the effect you want?

Nikratio:
(Superseded by #233)

PR #233, opened 2018-02-04
meson: Introducing skip-systemfiles

thopiekar:
Adding this option is an alternative to a PR (#232) I created before. By introducing it we can tell ninja to skip installing system files.


Nikratio:
Thanks for the patch! Could you change the help text to state exactly what won't be installed? I wouldn't know what 'systemfiles' refers to.

Nikratio:
Obsolete with PR 296

PR #240, opened 2018-03-25
Fix build error on DragonFlyBSD (sync with other *BSD)

kusumi:
> Meson encountered an error in file util/meson.build, line 24, column 2:
> Native dependency 'udev' not found

DragonFlyBSD has no "bsd" in uname, so add 'dragonfly' to conditionals.
While here do the same for test/util.py.

-- e.g. uname(1) in DragonFlyBSD
[root@ ~]# uname
DragonFly
[root@ ~]# python -c "import sys; print(sys.platform)"
dragonfly5


Nikratio:
thanks!

Issue #245, opened 2018-04-23
Deadlock in get_path2()


bnaylor:
We have some testcases that prove this comment correct 😄 

`        /* FIXME: locking two paths needs deadlock checking */`

(https://github.com/libfuse/libfuse/blob/master/lib/fuse.c#L1258)

Considering tackling this since I can reproduce the problem.  Did you have any thoughts as to how a fix should be implemented?

Nikratio:
Nope, you can free reign :-). I'd be great to get this solved.

jpandre:
You need to lock paths leading to nodeid1 and nodeid2 which are both known, so maybe you only have to always lock the one with the smallest nodeid first.

billziss-gh:
I have experienced the same problem on the Mac. See osxfuse/osxfuse#495.

The following patch appears to fix the problem on the Mac.

```diff
diff --git a/lib/fuse.c b/lib/fuse.c
index 9d295c8..569d1a2 100644
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -1223,7 +1223,7 @@ static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
 {
 	int err;
 
-	/* FIXME: locking two paths needs deadlock checking */
+	/* locking two paths needs deadlock checking -- see get_path2_ordered */
 	err = try_get_path(f, nodeid1, name1, path1, wnode1, true);
 	if (!err) {
 		err = try_get_path(f, nodeid2, name2, path2, wnode2, true);
@@ -1237,7 +1237,7 @@ static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
 	return err;
 }
 
-static int get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
+static int get_path2_ordered(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
 		     fuse_ino_t nodeid2, const char *name2,
 		     char **path1, char **path2,
 		     struct node **wnode1, struct node **wnode2)
@@ -1270,6 +1270,20 @@ static int get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
 	return err;
 }
 
+static int get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
+		     fuse_ino_t nodeid2, const char *name2,
+		     char **path1, char **path2,
+		     struct node **wnode1, struct node **wnode2)
+{
[...]

PR #249, opened 2018-05-08
Order locks in get_path2 to avoid deadlocks

billziss-gh:
This PR should fix issue #245. The same patch fixes a similar problem in OSXFUSE; see osxfuse/osxfuse#495

The PR passes all tests performed by `sudo python3 -m pytest test/`. I have not performed any additional testing at this time.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -1255,7 +1255,7 @@ static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
 {
 	int err;
 
-	/* FIXME: locking two paths needs deadlock checking */
+	/* locking two paths needs deadlock checking -- see get_path2_ordered */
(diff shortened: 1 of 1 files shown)


bnaylor:
Do not consider this a barrier to acceptance by any means, but I will also test this patch tomorrow with our internal tests that also uncovered this issue.  Will update with results.

billziss-gh:
@bnaylor absolutely. Looking forward to hear about the results.

bnaylor:
I still experience hangs with this change, but as we may be discovering in the issue page, my problem may be something specific to the bsd fuse kernel module.

billziss-gh:
I am withdrawing this patch as I have also discovered cases that it does not handle. My apologies for the noise. More information on the issue page.

PR #250, opened 2018-05-13
rename: perform user mode dir loop check when not done in kernel

billziss-gh:
This PR resolves the issue discussed in #245; the PR is not needed in Linux but is needed on BSD which libfuse purports to [support](https://github.com/libfuse/libfuse#supported-platforms).

Linux performs the directory loop check (`rename("a", "a/b/c")` or `rename("a/b/c", "a")`, etc.) at the VFS layer [[link](https://github.com/torvalds/linux/blob/v4.16/fs/namei.c#L4579-L4587)]. Unfortunately other libfuse supported systems do not perform this check, which results in a hang in `get_path2` as explained [here](https://github.com/libfuse/libfuse/issues/245#issuecomment-387845131).

The PR resolves the issue by adding a directory loop check in user mode for those system that do not perform it in kernel mode. The PR also adds a directory loop test in `test_syscalls.c`.

The PR has been tested on the following systems:

- On Linux where the PR is **NOT** needed using `python3 -m pytest test/` and the small test program included at the end of this post.

- On OSXFUSE where the same patch applies against [osxfuse/fuse](https://github.com/osxfuse/fuse) and resolves the same issue (osxfuse/osxfuse#495).

- **NOTE**: this PR has not been tested against FreeBSD or other BSD's. @bnaylor I am requesting your help here.

For completeness here is the small test program I used for my cross-platform testing:

```C
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <sys/stat.h>
#include <unistd.h>

void prename(const char *oldpath, const char *newpath, int expect_err)
{
    int err;

    err = -1 == rename(oldpath, newpath) ? errno : 0;
    if (err != expect_err)
        printf("rename(\"%s\", \"%s\") = %d (expect %d)\n", oldpath, newpath, err, expect_err);
    else
        printf("rename(\"%s\", \"%s\") = %d\n", oldpath, newpath, err);
}

int main()
{
    mkdir("a", 0700);
    prename("a", "a", 0);
    prename("a", "a/b", EINVAL);

    mkdir("a/b", 0700);
    mkdir("a/b/c", 0700);
    prename("a", "a/b/c", EINVAL);
    prename("a", "a/b/c/a", EINVAL);
    prename("a/b/c", "a", ENOTEMPTY);

    close(open("a/foo", O_CREAT));
    prename("a/foo", "a/bar", 0);
    prename("a/bar", "a/foo", 0);
    prename("a/foo", "a/b/bar", 0);
    prename("a/b/bar", "a/foo", 0);
    prename("a/foo", "a/b/c/bar", 0);
    prename("a/b/c/bar", "a/foo", 0);

    close(open("a/bar", O_CREAT));
    prename("a/foo", "a/bar", 0);
    unlink("a/bar");

    prename("a/b", "a/d", 0);
    prename("a/d", "a/b", 0);

    mkdir("a/d", 0700);
    prename("a/b", "a/d", 0);
    prename("a/d", "a/b", 0);

    mkdir("a/d", 0700);
    mkdir("a/d/e", 0700);
    prename("a/b", "a/d", ENOTEMPTY);
    rmdir("a/d/e");
    rmdir("a/d");

    rmdir("a/b/c");
    rmdir("a/b");
    rmdir("a");

    return 0;
}
```

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -1248,6 +1248,49 @@ static int get_path_wrlock(struct fuse *f, fuse_ino_t nodeid, const char *name,
 	return get_path_common(f, nodeid, name, path, wnode);
 }
 
+#if defined(__FreeBSD__)
+#define CHECK_DIR_LOOP true
+#else
+#define CHECK_DIR_LOOP false
+#endif
+
+static int check_dir_loop(struct fuse *f,
+			  fuse_ino_t nodeid1, const char *name1,
+			  fuse_ino_t nodeid2, const char *name2)
+{
+	struct node *node, *node1, *node2;
+	fuse_ino_t id1, id2;
+
+	node1 = lookup_node(f, nodeid1, name1);
+	id1 = node1 ? node1->nodeid : nodeid1;
+
+	node2 = lookup_node(f, nodeid2, name2);
+	id2 = node2 ? node2->nodeid : nodeid2;
+
+	for (node = get_node(f, id2); node->nodeid != FUSE_ROOT_ID;
+	     node = node->parent) {
+		if (node->name == NULL || node->parent == NULL)
+			break;
+
+		if (node->nodeid != id2 && node->nodeid == id1)
+			return -EINVAL;
+	}
+
+	if (node2)
+	{
+		for (node = get_node(f, id1); node->nodeid != FUSE_ROOT_ID;
+		     node = node->parent) {
+			if (node->name == NULL || node->parent == NULL)
+				break;
+
+			if (node->nodeid != id1 && node->nodeid == id2)
+				return -ENOTEMPTY;
+		}
+	}
+
+	return 0;
+}
+
 static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
 			 fuse_ino_t nodeid2, const char *name2,
 			 char **path1, char **path2,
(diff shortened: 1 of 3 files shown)


bnaylor:
I ran pjdfstest against Ontap with this patch and it passed.  I also confirmed that the new 'rename_dir_loop' test passes on stock FreeBSD.

billziss-gh:
@bnaylor thank you!

lib/fuse.c
```diff
@@ -1272,11 +1315,20 @@ static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
...
 		     char **path1, char **path2,
-		     struct node **wnode1, struct node **wnode2)
+		     struct node **wnode1, struct node **wnode2,
+		     bool dir_loop)
 {
 	int err;
 
 	pthread_mutex_lock(&f->lock);
+
+	if (dir_loop)
```
Nikratio: This looks like more churn than necessary. Is there a reason not to omit the new parameter and replace the `if(dir_loop)` with `ifdef CHECK_DIR_LOOP`?

billziss-gh: I agree. I struggled with the question of how to properly handle the conditional and decided to submit to get your feedback.

Are you ok with keeping the `dir_loop` parameter passed to `get_path2`, but then being ignored on Linux? Like so:

```C
#if defined(CHECK_DIR_LOOP)
static int check_dir_loop(struct fuse *f,
			  fuse_ino_t nodeid1, const char *name1,
			  fuse_ino_t nodeid2, const char *name2)
{
			  ...
}
#endif

...

#if defined(CHECK_DIR_LOOP)
	if (dir_loop)
	{
		err = check_dir_loop(f, nodeid1, name1, nodeid2, name2);
		if (err)
			goto out_unlock;
	}
#else
	(void)dir_loop;
#endif
```

billziss-gh:
Please see review response above.

billziss-gh:
As per your request I have updated the patch to not include any of the new code under Linux. I have also updated the Changelog.

This new patch has been tested:

- Under Linux (where the patch is inactive).
- Under FreeBSD (where the patch is active).
- Under OSX where the same patch fixes a similar `rename` deadlock.

I am requesting another review.

billziss-gh:
Fantastic. Thank you @Nikratio.

billziss-gh:
@Nikratio just wondering if we should also include this bugfix in the 2.9 series? Let me know and I can send PR.

Nikratio:
If you have the time, absolutely. Thanks!

PR #252, opened 2018-05-22
fix documentation for opendir in fuse_operations

edquist:
the filehandle from opendir is passed to releasedir - there is no
closedir function in fuse_operations


Nikratio:
Thanks!

PR #259, opened 2018-06-27
Add support for copy_file_range()

nixpanic:
This series updates the `fuse_kernel.h` file to the most recent version in Linus' tree.

`copy_file_range()` support in the FUSE kernel module has not been merged yet. The [first version of the patch that adds it can be found on LKML](https://lkml.org/lkml/2018/6/27/86) and other mailinglist archives.

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -750,6 +750,23 @@ struct fuse_operations {
 	 */
 	int (*fallocate) (const char *, int, off_t, off_t,
 			  struct fuse_file_info *);
+
+	/**
+	 * Copy a range of data from one file to another
+	 *
+	 * Performs an optimized copy between two file descriptors without the
+	 * additional cost of transferring data through the FUSE kernel module
+	 * to user space (glibc) and then back into the FUSE filesystem again.
+	 *
+	 * In case this method is not implemented, glibc falls back to reading
+	 * data from the source and writing to the destination. Effectively
+	 * doing an inefficient copy of the data.
+	 */
+	ssize_t (*copy_file_range) (const char *path_in,
+				    struct fuse_file_info *fi_in,
+				    off_t offset_in, const char *path_out,
+				    struct fuse_file_info *fi_out,
+				    off_t offset_out, size_t size, int flags);
 };
 
 /** Extra context that may be needed by some filesystems
@@ -1165,6 +1182,11 @@ int fuse_fs_poll(struct fuse_fs *fs, const char *path,
 		 unsigned *reventsp);
 int fuse_fs_fallocate(struct fuse_fs *fs, const char *path, int mode,
 		 off_t offset, off_t length, struct fuse_file_info *fi);
+ssize_t fuse_fs_copy_file_range(struct fuse_fs *fs, const char *path_in,
+				struct fuse_file_info *fi_in, off_t off_in,
+				const char *path_out,
+				struct fuse_file_info *fi_out, off_t off_out,
+				size_t len, int flags);
 void fuse_fs_init(struct fuse_fs *fs, struct fuse_conn_info *conn,
 		struct fuse_config *cfg);
 void fuse_fs_destroy(struct fuse_fs *fs);
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -1160,6 +1160,42 @@ struct fuse_lowlevel_ops {
 	 */
 	void (*readdirplus) (fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
 			 struct fuse_file_info *fi);
+
+	/**
+	 * Copy a range of data from one file to another
+	 *
+	 * Performs an optimized copy between two file descriptors without the
+	 * additional cost of transferring data through the FUSE kernel module
+	 * to user space (glibc) and then back into the FUSE filesystem again.
+	 *
+	 * In case this method is not implemented, glibc falls back to reading
+	 * data from the source and writing to the destination. Effectively
+	 * doing an inefficient copy of the data.
+	 *
+	 * If this request is answered with an error code of ENOSYS, this is
+	 * treated as a permanent failure with error code EOPNOTSUPP, i.e. all
+	 * future copy_file_range() requests will fail with EOPNOTSUPP without
+	 * being send to the filesystem process.
+	 *
+	 * Valid replies:
+	 *   fuse_reply_write
+	 *   fuse_reply_err
+	 *
+	 * @param req request handle
+	 * @param ino_in the inode number or the source file
+	 * @param off_in starting point from were the data should be read
+	 * @param fi_in file information of the source file
+	 * @param ino_out the inode number or the destination file
+	 * @param off_out starting point where the data should be written
+	 * @param fi_out file information of the destination file
+	 * @param len maximum size of the data to copy
+	 * @param flags passed along with the copy_file_range() syscall
+	 */
+	void (*copy_file_range) (fuse_req_t req, fuse_ino_t ino_in,
+				 off_t off_in, struct fuse_file_info *fi_in,
+				 fuse_ino_t ino_out, off_t off_out,
+				 struct fuse_file_info *fi_out, size_t len,
+				 int flags);
 };
 
 /**
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('libfuse3', 'c', version: '3.3.0',
+project('libfuse3', 'c', version: '3.4.0',
         meson_version: '>= 0.42',
         default_options: [ 'buildtype=debugoptimized' ])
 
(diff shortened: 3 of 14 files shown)


Nikratio:
Thanks! Could you please ping me again when the kernel part is merged? (I'll try to keep an eye on it, but I will probably forget).

nixpanic:
Yes, of course! I am not sure if there will be any requests to change
the kernel headers, so until then this should not get merged in any
case.

include/fuse_lowlevel.h
```diff
@@ -1160,6 +1160,34 @@ struct fuse_lowlevel_ops {
...
+	 * @param fd_in file description used for reading the data from
+	 * @param off_in starting point of fd_in from were the data is read
+	 * @param fd_out file description used for writing the data to
+	 * @param off_out starting point of fd_out where the data is written
+	 * @param len maximum size of the data to copy
+	 * @param flags passed along with the copy_file_range() syscall
+	 */
+	void (*copy_file_range) (fuse_req_t req, int fd_in, off_t off_in,
+				 int fd_out, off_t off_out, size_t len,
+				 int flags);
```
marcinsulikowski: This definition is incorrect – FUSE operations which operate on file descriptors take `fuse_file_info* fi` arguments, not `int fd`.

Additionally, all other FUSE low-level operations which operate on file descriptors are additionally given inode numbers (see, e.g, `fallocate` or `read`). The inode number is needed by the high-level interface to provide file path to the high-level callback. As such, in my opinion, this low-level callback should be actually declared as:
```c
void (*copy_file_range) (fuse_req_t req,
      fuse_ino_t ino_in, off_t off_in, fuse_file_info *fi_in,
      fuse_ino_t ino_out, off_t off_out, fuse_file_info *fi_out,
      size_t len, int flags);
```
There is a caveat though – the kernel module change proposed in https://lkml.org/lkml/2018/6/27/114 provides only `ino_in` argument because that is passed in the request's header (`args.in.h.nodeid = ff_in->nodeid;` in `fs/fuse/fuse.c` in the kernel) whereas `ino_out` is not given by the kernel to the userspace implementation. Isn't this an oversight in the kernel module? The lack of the inode number would make it impossible to create high-level interface which follows the libfuse's convention where the high-level callbacks are given paths to files which are to be modified.

include/fuse.h
```diff
@@ -750,6 +750,15 @@ struct fuse_operations {
...
 			  struct fuse_file_info *);
+
+	/**
+	 * Copy a range of data from one file to another
+	 *
+	 * Performs an optimized copy between two file descriptors without the
+	 * additional cost of transferring data through the FUSE kernel module
+	 * to user space (glibc) and then back into the FUSE filesystem again.
+	 */
+	ssize_t (*copy_file_range) (int, off_t, int, off_t, size_t, int);
```
marcinsulikowski: This definition is incorrect – FUSE operations which operate on file descriptors take `fuse_file_info* fi` arguments, not `int fd`. Additionally, this declaration doesn't follow conventions used in the FUSE's high-level interface where callbacks which operate on file descriptors are also given paths to the files. Thus, this should rather be declared as:
```c
ssize_t (*copy_file_range) (const char* path_in, fuse_file_info* fi_in, off_t offset_in,
                            const char* path_out, fuse_file_info* fi_out, off_t offset_out,
                            size_t size, int flags);
```

include/fuse_lowlevel.h
nixpanic: I'll see if I can add `ino_out` or change the file-descriptor to use `struct fuse_file_into*`.

The `ino_out` is not required for supporting the functionality with Gluster (does not use libfuse). If there is an advantage to having ino_out, it can be added to the kernel module. Do you want to send that suggestion as a reply to the kernel email so it is tracked there? I'll include your suggestion in v3 later this week.

include/fuse.h
nixpanic: Ok, if that is the convention, I'll change it.

nixpanic:
Thanks @marcinsulikowski , [an updated FUSE kernel change](https://lkml.org/lkml/2018/6/29/334) has been posted. This PR now uses that API and includes the suggestions about the libfuse improvements.

nixpanic:
Latest update, based on the review comments from Miklos. The matching [fuse module patch](https://lkml.org/lkml/2018/8/21/392) has been posted as well.

Nikratio:
Thanks! Please ping me when the kernel part has been merged.

nixpanic:
@Nikratio, the kernel change has been merged in Linus' tree as [commit 88bc7d50](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/fs/fuse?id=88bc7d5097a11d9bdcf08ecf85c81ba998353437). Please review this PR and see if it can be included in fuse-3.4?

include/fuse.h
```diff
@@ -750,6 +750,19 @@ struct fuse_operations {
 	 */
 	int (*fallocate) (const char *, int, off_t, off_t,
 			  struct fuse_file_info *);
+
+	/**
+	 * Copy a range of data from one file to another
+	 *
+	 * Performs an optimized copy between two file descriptors without the
+	 * additional cost of transferring data through the FUSE kernel module
+	 * to user space (glibc) and then back into the FUSE filesystem again.
```
Nikratio: Please also describe what happes if this method isn't implemented (I assume the kernel will assume a bunch of separate read() and write() requests)?

include/fuse_lowlevel.h
```diff
@@ -1160,6 +1160,38 @@ struct fuse_lowlevel_ops {
 	 */
 	void (*readdirplus) (fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
 			 struct fuse_file_info *fi);
+
+	/**
+	 * Copy a range of data from one file to another
+	 *
+	 * Performs an optimized copy between two file descriptors without the
+	 * additional cost of transferring data through the FUSE kernel module
+	 * to user space (glibc) and then back into the FUSE filesystem again.
```
Nikratio: same here

test/test_examples.py
```diff
@@ -585,6 +585,10 @@ def tst_utimens(mnt_dir, ns_tol=0):
         assert abs(fstat.st_atime_ns - atime_ns) <= ns_tol
         assert abs(fstat.st_mtime_ns - mtime_ns) <= ns_tol
 
+def tst_copy_file_range(mnt_dir):
+    # no os.copy_file_range yet, https://github.com/python/cpython/pull/7255
+    pass
+
```
Nikratio: Let's just omit this definition then.

include/fuse.h
nixpanic: Currently it is actually glibc doing a fallback. There seems to be work in progress in the kernel to do this, but that has not been merged yet (or even no final design?).

nixpanic:
Thanks for the comments. I think I've addressed everything you mentioned.

meson.build
```diff
@@ -1,4 +1,4 @@
-project('libfuse3', 'c', version: '3.3.0',
+project('libfuse3', 'c', version: '3.4.0',
```
Nikratio: Could you please also bump `FUSE_USE_VERSION` throughout the codebase, and merge the API bump commit with the commit adding the feature ("libfuse: add copy_file_range() support")?

nixpanic: Rebased and squashed the version change.

Did you mean the `FUSE_USE_VERSION` in the `meson.build` files, only? I did not notice others that need updating.

Nikratio: The others don't stricly *need* updating, but I would like to keep them at the most recent version so that it doesn't appear that the examples are relying on behavior that has since changed:

```
$ ack FUSE_USE_VERSION
test/test_write_cache.c
10:#define FUSE_USE_VERSION 30

test/test_setattr.c
10:#define FUSE_USE_VERSION 30

include/fuse_common.h
759:#if !defined(FUSE_USE_VERSION) || FUSE_USE_VERSION < 30

include/fuse_lowlevel.h
16: * IMPORTANT: you should define FUSE_USE_VERSION before including this
21:#ifndef FUSE_USE_VERSION
22:#error FUSE_USE_VERSION not defined
1884:#if FUSE_USE_VERSION < 32

include/fuse.h
16: * IMPORTANT: you should define FUSE_USE_VERSION before including this header.
885:#if FUSE_USE_VERSION == 30
981:#if FUSE_USE_VERSION < 32

include/cuse_lowlevel.h
15:#ifndef FUSE_USE_VERSION
16:#define FUSE_USE_VERSION 29

example/hello.c
22:#define FUSE_USE_VERSION 31

[...]
```

nixpanic: And now with updated `examples/` :-)

Nikratio:
Hmm. It seems the tests already made use of deprecated functionality (at least if I interpret the CI results correctly). Sorry about that! It would of course be great if you could get them to work with the newest API, but if you just want to redefine FUSE_USE_API to the earlier version that's fine too.

nixpanic:
I've reverted the `FUSE_USE_VERSION` changes in `examples/` now. As a reminder, #324 has been created. Not sure if/when I will have time to look at that though.

Nikratio:
Thanks!

PR #260, opened 2018-06-27
Small improvements

gcsideal:
Fix man page name, build static library as well. Mostly for Debian packaging but would be good for everyone.


Nikratio:
Looks like both_libraries has only been introduced in Meson 0.46 (i.e. the newest release). Is there a reason to use this instead of two separate targets (library + static_library)? I'd prefer not to depend on such a recent Meson version, but if this is required please bump the minimum version in `meson.build` and the version used in the CI (`tests/travis-install.sh`).

gcsideal:
The use of both_libraries was to keep library definitions together. But you are right, for supporting older Meson versions the distinct shared and static libraries build is encouraged. I think I'll have time for it in this afternoon.

Nikratio:
Actually, I don't think we need to change the build file at all. In recent meson versions, you can just use `meson configure -D default_library=both` (cf. https://github.com/mesonbuild/meson/pull/2711). I've merged the other changes.

PR #264, opened 2018-07-09
Fix SIGSEGV when fuse_interrupted() is called outside the eventloop

Alexander-TX:
The issue was previously raised on [fuse-devel](http://fuse.996288.n3.nabble.com/how-to-detect-if-within-Fuse-thread-calling-fuse-interrupted-from-non-Fuse-thread-td10832.html), but the proposed fix never got merged.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -4417,7 +4417,9 @@ int fuse_getgroups(int size, gid_t list[])
 
 int fuse_interrupted(void)
 {
-	return fuse_req_interrupted(fuse_get_context_internal()->req);
+	struct fuse_context_i *c = fuse_get_context_internal();
+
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks!

Alexander-TX:
@Nikratio are there plans to release libfuse 2.9.8 with this fix? I have a patch, adding interruption support to a third-party filesystem, but this issue prevents me from upstreaming it.

Nikratio:
Yeah, sort-of. My efforts have been focused on fuse 3 and I haven't done a libfuse 2.x release for several years so I have actually forgotten what the procedure was. I have automated and documented most of it for libfuse 3 because it was somewhat cumbersome, so I am somewhat dreading to get back to it. But yeah, I'll make an effort.

PR #268, opened 2018-07-17
fusermount hardening

thejh:
Some of these patches are bugfixes for issues that I believe to be non-exploitable, the others are hardening changes that should make `fusermount` more robust against potential future problems.

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -819,10 +819,16 @@ static int do_mount(const char *mnt, char **typep, mode_t rootmode,
 						flags |= flag;
 					else
 						flags  &= ~flag;
-				} else {
+				} else if (opt_eq(s, len, "default_permissions") ||
+					   opt_eq(s, len, "allow_other") ||
+					   begins_with(s, "max_read=") ||
+					   begins_with(s, "blksize=")) {
 					memcpy(d, s, len);
 					d += len;
 					*d++ = ',';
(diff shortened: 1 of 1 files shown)


util/fusermount.c
```diff
@@ -790,10 +820,16 @@ static int do_mount(const char *mnt, char **typep, mode_t rootmode,
 						flags |= flag;
 					else
 						flags  &= ~flag;
-				} else {
```
Nikratio: Not sure about this. Is passing unknown options to the kernel really a significant attack surface? It has the drawback that it requires updating libfuse when the kernel starts supporting a new option.

util/fusermount.c
```diff
@@ -711,6 +722,23 @@ static int get_string_opt(const char *s, unsigned len, const char *opt,
 	return 1;
 }
 
+/* The kernel silently truncates the "data" argument to PAGE_SIZE-1 characters.
+ * This can be dangerous if it e.g. truncates the option "group_id=1000" to
+ * "group_id=1".
+ * This wrapper detects this case and bails out with an error.
+ */
```
Nikratio: Thanks for noticing that!

Nikratio:
Makes sense to me. Could you provide a pull request? Please don't forget to add a brief entry to ChangeLog.rst as well.

PR #269, opened 2018-07-20
Fix readdir() bug when a non-zero offset is specified in filler

rostislav:
Hello,

The first commit adds a test of seekdir(), and the second fixes the bug that is exposed by the test.
I am not sure this is the best fix, but I tried to change as little as possible.

Another thing is that 2.9.x branch is also affected by the same bug. Would a similar fix be accepted there as well?

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -2,6 +2,9 @@ libfuse 3.2.5
 ==========================
 
 * Added a test of `seekdir` to test_syscalls.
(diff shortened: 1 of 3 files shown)


Nikratio:
Thanks for tracking this down! I think your fix is fine. I have just one request: could you please also add a comment to ChangeLog.rst?

rostislav:
Thanks for review! I've added comments to `ChangeLog.rst` and edited commit messages a bit.

PR #270, opened 2018-07-21
Fix memory leaks

rostislav:
This series of commits fixes all memory leaks I found when running tests with LeakSanitizer enabled.


Nikratio:
Thanks! Much appreciated.

PR #271, opened 2018-07-24
Fix readdir() bug when a non-zero offset is specified in filler

rostislav:
Hi Nikolaus,

This is a backport of my earlier commit in master fixing the same bug.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+Unreleased Changes
+==================
+* Added a test of `seekdir` to test_syscalls.
+* Fixed `readdir` bug when non-zero offsets are given to filler and the
+  filesystem client, after reading a whole directory, re-reads it from a
+  non-zero offset e. g. by calling `seekdir` followed by `readdir`.
+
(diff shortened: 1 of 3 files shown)


ChangeLog
```diff
@@ -2,6 +2,10 @@ FUSE 2.9.8
 ==========
 
 * Fixed rename deadlock on FreeBSD.
+* Added a test of `seekdir` to test_syscalls.
```
Nikratio: 2.9.8 has already been released, so this needs a new entry ("Unreleased Changes")

rostislav:
When updating the changelog as per your request, I noticed that the changes actually included in 2.9.8 are under "Unreleased Changes" in the changelog (e. g. fusermount hardening). I guess it's not intended?

Nikratio:
Indeed, I forgot to commit this file. Should be fixed now. thanks for the pointer!

rostislav:
Updated my commit with your latest update of `fuse_2_9_bugfix`, now my changes in ChangeLog are under "Unreleased Changes"

PR #278, opened 2018-08-04
Add FAT to mountpoint file system whitelist

gcsideal:
FAT needs to whitelisted as well.


Nikratio:
Thanks! Ben was a little faster than you :-). https://github.com/libfuse/libfuse/pull/277

PR #282, opened 2018-08-15
passthrough_ll update

szmi:
This fixes a couple of bugs in the passthrough_ll example filesystem and adds support for lots of missing operations.  Options are added to configure a base path and to control caching.

### DIFF
--- a/example/passthrough_ll.c
+++ b/example/passthrough_ll.c
@@ -329,7 +329,7 @@ static void lo_lookup(fuse_req_t req, fuse_ino_t parent, const char *name)
 	if (lo_debug(req))
 		fprintf(stderr, "lo_lookup(parent=%" PRIu64 ", name=%s)\n",
 			parent, name);
-	
+
 	err = lo_do_lookup(req, parent, name, &e);
 	if (err)
 		fuse_reply_err(req, err);
@@ -730,7 +730,7 @@ static void lo_create(fuse_req_t req, fuse_ino_t parent, const char *name,
 	if (lo_debug(req))
 		fprintf(stderr, "lo_create(parent=%" PRIu64 ", name=%s)\n",
 			parent, name);
-			
+
 	fd = openat(lo_fd(req, parent), name,
 		    (fi->flags | O_CREAT) & ~O_NOFOLLOW, mode);
 	if (fd == -1)
@@ -758,8 +758,7 @@ static void lo_fsyncdir(fuse_req_t req, fuse_ino_t ino, int datasync,
 	fuse_reply_err(req, res == -1 ? errno : 0);
 }
 
-static void lo_open(fuse_req_t req, fuse_ino_t ino,
-		    struct fuse_file_info *fi)
+static void lo_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
 {
 	int fd;
 	char buf[64];
@@ -853,7 +852,7 @@ static void lo_write_buf(fuse_req_t req, fuse_ino_t ino,
 	if (lo_debug(req))
 		fprintf(stderr, "lo_write(ino=%" PRIu64 ", size=%zd, off=%lu)\n",
 			ino, out_buf.buf[0].size, (unsigned long) off);
-	
+
 	res = fuse_buf_copy(&out_buf, in_buf, 0);
 	if(res < 0)
 		fuse_reply_err(req, -res);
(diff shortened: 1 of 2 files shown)


example/passthrough_ll.c
```diff
@@ -87,6 +88,9 @@ static const struct fuse_opt lo_opts[] = {
 	  offsetof(struct lo_data, writeback), 1 },
 	{ "no_writeback",
 	  offsetof(struct lo_data, writeback), 0 },
+	{ "source=%s",
```
Nikratio: I'd much rather see this as a parameter (like the mountpoint). Any reason you made it a mount option?

Nikratio:
Thanks! This is great, I really like having a more feature-rich low-level example. My only concerns are the  two minor issues mentioned above.

Nikratio:
*ping* Would you be able to add the documentation update, so this can get merged? Should be just a few lines if I understand the issue correctly.

Nikratio:
*ping* Are you no longer interested in working on this?

The only thing that prevents a merge is the missing testcase...

szmi:
I'm definitely interested in this getting merged.  It's part of a bigger project that we are working on internally and will hopefully go public in some weeks. Getting it upstream is obviously very important.

OTOH, I've a lot on my plate and python is not the language that I'm highly skilled in, so I'd definitely accept some help there.

Nikratio:
Got you. Will add a test case later today, if I can find out how to extend the push request (there was some way to do that...)

Nikratio:
Ok, I've extended the testcases. There is one new failure when running with active writeback. It comes from the `test_syscalls` test which is in C :-).

```
50 [open_acc(O_RDONLY) mode: 0400 message: 'Success'] OK
 51 [open_acc(O_WRONLY) mode: 0200 message: 'Success'] do_test_open_acc() - open: Permission denied
 52 [open_acc(O_RDONLY | O_TRUNC) mode: 0400 message: 'Permission denied'] OK
```

Unfortunately I do not have permission to extend your pull request with the test changes. Could you give me permissions? Instructions at 
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

Nikratio:
For now, I have pushed the updated branch to https://github.com/libfuse/libfuse/commits/miklos-pull-request

szmi:
Thank's for helping.

Ticked the box to give permissions.

The failed testcase is expected.  This mode only works if the server has elevated privileges (CAP_DAC_OVERRIDE) because in writeback mode the kernel may also issue read requests (even if file was opened for write-only) and hence the passthrough fs has to open the file for reading as well as for writing, which fails if the file is only openable for write and the server lacks privileges.

I think the right solution to that would be to disable writeback cache for the inode where read permission is denied (an unlikely situation).  This is not currently possible, as writeback mode is per-superblock not per-inode, but that's fixable in the fuse kernel module and in API's.

Nikratio:
Right. I don't suppose you have a patch ready for implementing per-inode cache settings? :-)

What do you think is the best way to handle the current situation? I think we can either drop this particular test completely, or add an additional command line parameter to `test_syscalls` that causes it to skip the test only if writeback caching is enabled.

szmi:
Adding "-51" to the test command line will disable this test.  If you only add it for the "-owriteback" case, all the better.

Nikratio:
done, let's see

Nikratio:
Are those numbers dynamic? Test 51 is now skipped, but now test 52 is the failing one?!

szmi:
Umm, a seekdir test was added by commit 5f125c5e6be2 "Fix readdir() bug when a non-zero offset is specified in filler (#269)".  The problem is that it was added in the middle, thus causing all the test numbers after it to shift by one.

Nikratio:
I don't follow. The test passes locally (with 51 excluded), but did not pass on Travis. I haven't rebased anything, it seems that the numbers are simply different between the test runs.

szmi:
The travis run has a seekdir test at position12 while the local run doesn't have it.  So I guess travis is doing a merge with HEAD before running the tests.

Nikratio:
Explicitly rebased and updated the test number now. Let's see...

Nikratio:
Looks good now! Could you still document the  that . and .. should not get an increased lookup count? As far as I can tell, this isn't documented anywhere yet the example refers to this as a requirement.

Nikratio:
*ping*

Any chance you could write a few words so we can merge this? :-)

szmi:
It seems to be documented in fuse_lowlevel.h:

	 * In contrast to readdir() (which does not affect the lookup counts),
	 * the lookup count of every entry returned by readdirplus(), except "."
	 * and "..", is incremented by one.

The sentence is structured a bit awkwardly, maybe it should be
 
	 * In contrast to readdir() the lookup count of every entry returned by readdirplus() is incremented by one, with the exception of "." and ".."

Nikratio:
Right you are. Weird, I was sure I carefully looked at exactly that section of the file...

Thanks for you patience! Merging now.

Issue #284, opened 2018-08-23
util/install_helper.sh problems when running not under root user (using DESTDIR)


AnimusPEXUS:
patch
```diff
diff -ru fuse-3.2.5/util/install_helper.sh 01.SOURCE/util/install_helper.sh
--- fuse-3.2.5/util/install_helper.sh	2018-07-24 09:45:33.000000000 +0300
+++ 01.SOURCE/util/install_helper.sh	2018-08-23 12:17:45.474509001 +0300
@@ -11,12 +11,17 @@
 udevrulesdir="$3"
 prefix="${MESON_INSTALL_DESTDIR_PREFIX}"
 
-chown root:root "${prefix}/${bindir}/fusermount3"
-chmod u+s "${prefix}/${bindir}/fusermount3"
+if test -z "${DESTDIR}"; then
+    chown root:root "${prefix}/${bindir}/fusermount3"
+    chmod u+s "${prefix}/${bindir}/fusermount3"
+fi
+
 
-if test ! -e "${DESTDIR}/dev/fuse"; then
-    mkdir -p "${DESTDIR}/dev"
-    mknod "${DESTDIR}/dev/fuse" -m 0666 c 10 229
+if test -z "${DESTDIR}"; then
+    if test ! -e "${DESTDIR}/dev/fuse"; then
+        mkdir -p "${DESTDIR}/dev"
+        mknod "${DESTDIR}/dev/fuse" -m 0666 c 10 229
+    fi
 fi
 
 install -D -m 644 "${MESON_SOURCE_ROOT}/util/udev.rules" \
```

Nikratio:
Closing this in favor of the pull request.

PR #285, opened 2018-08-23
fix ninja install when running not under root

AnimusPEXUS:
https://github.com/libfuse/libfuse/issues/284

### DIFF
--- a/util/install_helper.sh
+++ b/util/install_helper.sh
@@ -11,12 +11,17 @@ bindir="$2"
 udevrulesdir="$3"
 prefix="${MESON_INSTALL_DESTDIR_PREFIX}"
 
-chown root:root "${prefix}/${bindir}/fusermount3"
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks for the patch! But could you explain which problem you are solving? How is installation as non-root related to DESTDIR? And why would you not install the init script when DESTDIR is set?

AnimusPEXUS:
I'm building packages not under root.
trying to 'ninja install DESTDIR=(something)' installs some parts to '(something)', but fails on the part, where external script tries to create /dev/fuse (mknod) and chown/chmod on "${prefix}/${bindir}/fusermount3"

I've used 'test -z "${DESTDIR}"' check, to keep up in the mood of other checks inside util/install_helper.sh, keeping in mind, what DESTDIR variable - most probably used not under the root.

PR #286, opened 2018-08-23
Add mknod and setattr to passthrough_ll

mtheall:


mtheall:
Sorry, I didn't notice there was another passthrough_ll pr that already adds this.

PR #290, opened 2018-08-28
return different non-zero error codes

guss77:
For different errors, we should have different error codes to help a developer figure out what exactly we don't like about their system.

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -303,30 +303,30 @@ int fuse_main_real(int argc, char *argv[], const struct fuse_operations *op,
 	if (!opts.show_help &&
 	    !opts.mountpoint) {
 		fprintf(stderr, "error: no mountpoint specified\n");
-		res = 1;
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks for the patch! I think you forgot the most important part though: updating the documentation so that developers can actually make use of the new error codes 😉.

Would you be able to do that as well?

guss77:
Current documentation (Assuming I found the correct location) just say "return ... nonzero on failure". Which is still accurate ;-).

I can try to analyze the reason for all the different codes (including `1`) and add some text - is this what you mean?

Nikratio:
I mean you should describe what error code is returned in which situation - otherwise there's no point in returning different error codes, because they don't mean anything, right?

guss77:
@Nikratio - you mean in `fuse.h` line 821 ?

Nikratio:
yes

Issue #77, opened 2016-09-20
allow filesystem operations before mount(2) completes in libfuse-3.0


szmi:
Historically libfuse supports the following sequence of initialization:

1) open /dev/fuse
2) mount filesystem at mountpoint, attaching the device fd to the mount at the same time
3) start processing requests

While this is the simplest from the libfuse API point of view, it causes problems in practice.

One such problem is that security modules want to read extended attributes on the root before completing the mount.

A smaller problem is INIT reply comes when the filesystem is already active (though other requests are blocked by the fuse kernel module) so it's problematic making changes to super block setup (e.g. sb->s_flags) that normally happen before the initial mount completes.

So it would be better if the initialization sequence looked like this:

1) open /dev/fuse
2) start processing requests in the background
3) mount filesystem

Fortunately the high level fuse_main() API would be unaffected.  Unfortunately everything else related to initialization would be.  Will try to cook up a patch...

Nikratio:
Could you hold on with that patch for a day? I am working on something related that will make the rebase rather difficult.

Nikratio:
Actually, I'll push it on a separate branch. It's not working yet, but it should give you an idea of what I have in mind.

Nikratio:
Done, could you take a look at https://github.com/libfuse/libfuse/tree/init-rework? Your comments would be much appreciated.

Nikratio:
@szmi , could you take a look at the updated API in the master branch and check if it's compatible with the change that you're proposing? I'd like to get out a FUSE 3 release soon, so if allowing file system operations before mounting requires changes to the API I'd rather incorporate that before the release.

Nikratio:
@szmi Do you know when/if you want to look at this?

Nikratio:
I looked into this a little. Moving the `open(/dev/fuse)` into `fuse_session_new` is easy enough. However, there are two problems:

- Under certain conditions, `fuse_kern_mount` falls back on creating a pipe, forking, and executing `fusermount` to obtain the fd. But to determine if that's necessary, we have to actually attempt the mount operation. So unless there is some way to start with one fd, and then continue with a different one later one, we need a way to determine if `fusermount` will be required without actually attempting to mount.

-  It seems that the kernel does not send the INIT request until the file system is mounted - even if the fd is opened earlier (at least with kernel 4.7). So unless we want the file system to start handling replies before INIT (which I think is a  bad idea), making this work would require some changes to the kernel.

Nikratio:
[...]

PR #291, opened 2018-08-29
Allow passing `/dev/fuse` file descriptor from parent process 

saittam:
This adds a new "pre-mounted" mode of operation in which the FUSE file
system helper is launched after the /dev/fuse file descriptor has been
opened opened and the file system been mounted by a suitably
privileged process.

Pre-mounted mode is requested by passing an empty string as
mountpoint, which reflects the fact that the FUSE helper doesn't
actually perform or trigger any mounting itself.

The main benefit of pre-mounted mode is that no privileged operations
need to be performed by the file system implementation itself directly
or indirectly, so the FUSE process can run with zero privilege.
Moreover, mechanisms like securebits and no_new_privs can be used to
prevent subprocesses from re-acquiring privilege, which further helps
reduce risk for the case the FUSE helper gets exploited by a malicious
file system.

Below is an example that illustrates this. Note that I'm using shell
for presentation purposes, the assumption is that there's a privileged
daemon that handles mounting and spawning the FUSE helper in a
suitable sandbox.

  # Make binaries and libs executable by anyone
  $ chmod o+rx build/lib/libfuse3.so* example/hello
  $ export LD_LIBRARY_PATH=$PWD/build/lib

  # example/hello can mount successfully with privilege
  $ sudo sh -c "LD_LIBRARY_PATH=build/lib ./example/hello /mnt/tmp"
  $ sudo cat /mnt/tmp/hello
  Hello World!
  $ sudo umount /mnt/tmp

  # example/hello fails to mount without privilege
  $ sudo capsh --drop=all --secbits=0x3f -- -c 'LD_LIBRARY_PATH=build/lib ./example/hello -f /mnt/tmp'
  fusermount3: mount failed: Operation not permitted

  # Pre-mounting allows example/hello to work without privilege
  $ sudo sh -c '
    exec 0<>/dev/fuse
    mount -i -o nodev,nosuid,noexec,fd=0,rootmode=40000,user_id=0,group_id=0 -t fuse hello /mnt/tmp
    capsh --drop=all --secbits=0x3f -- -c "LD_LIBRARY_PATH=build/lib example/hello -f \"\""
  ' &
  [1] 55491
  $ sudo cat /mnt/tmp/hello
  Hello World!
  $ sudo umount /mnt/tmp
  [1]+  Done                    sudo sh -c '
    exec 0<>/dev/fuse
    mount -i -o nodev,nosuid,noexec,fd=0,rootmode=40000,user_id=0,group_id=0 -t fuse hello /mnt/tmp
    capsh --drop=all --secbits=0x3f -- -c "LD_LIBRARY_PATH=build/lib example/hello -f \"\""
  '

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -8,6 +8,10 @@ Unreleased Changes
 * The description of the FUSE_CAP_READDIRPLUS_AUTO flag has been
   improved.
 
+* Allow open `/dev/fuse` file descriptors to be passed via mountpoints of the
+  special format `/dev/fd/%u`. This allows mounting to be handled by the parent
+  so the FUSE filesystem process can run fully unprivileged.
+
 libfuse 3.2.6 (2018-08-31)
 ==========================
 
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -12,6 +12,11 @@ Unreleased Changes
   special format `/dev/fd/%u`. This allows mounting to be handled by the parent
   so the FUSE filesystem process can run fully unprivileged.
 
+* Add a `drop_privileges` option to mount.fuse3 which causes it to open
+  `/dev/fuse` and mount the file system itself, then run the FUSE file
+  filesystem fully unprivileged and unable to re-acquire privilege via setuid,
+  fscaps, etc.
+
 libfuse 3.2.6 (2018-08-31)
 ==========================
 
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -16,6 +16,7 @@
 #include "fuse_kernel.h"
 #include "fuse_opt.h"
 #include "fuse_misc.h"
+#include "mount_util.h"
 
 #include <stdio.h>
 #include <stdlib.h>
@@ -2891,6 +2892,24 @@ int fuse_session_mount(struct fuse_session *se, const char *mountpoint)
 			close(fd);
 	} while (fd >= 0 && fd <= 2);
 
+	/*
+	 * To allow FUSE daemons to run without privileges, the caller may open
+	 * /dev/fuse before launching the file system and pass on the file
+	 * descriptor by specifying /dev/fd/N as the mount point. Note that the
+	 * parent process takes care of performing the mount in this case.
+	 */
+	fd = fuse_mnt_parse_fuse_fd(mountpoint);
+	if (fd != -1) {
+		if (fcntl(fd, F_GETFD) == -1) {
+			fprintf(stderr,
+				"fuse: Invalid file descriptor /dev/fd/%u\n",
+				fd);
+			return -1;
+		}
+		se->fd = fd;
+		return 0;
+	}
+
 	/* Open channel */
 	fd = fuse_kern_mount(mountpoint, se->mo);
 	if (fd == -1)
@@ -2916,9 +2935,11 @@ int fuse_session_fd(struct fuse_session *se)
 
 void fuse_session_unmount(struct fuse_session *se)
 {
-	fuse_kern_unmount(se->mountpoint, se->fd);
-	free(se->mountpoint);
-	se->mountpoint = NULL;
+	if (se->mountpoint != NULL) {
+		fuse_kern_unmount(se->mountpoint, se->fd);
+		free(se->mountpoint);
+		se->mountpoint = NULL;
+	}
 }
 
 #ifdef linux
--- a/test/util.py
+++ b/test/util.py
@@ -7,6 +7,7 @@ import time
 from os.path import join as pjoin
 import sys
 import re
+import itertools
 
 basename = pjoin(os.path.dirname(__file__), '..')
 
@@ -138,6 +139,12 @@ def fuse_test_marker():
 
     return pytest.mark.uses_fuse()
 
+def powerset(iterable):
+  s = list(iterable)
+  return itertools.chain.from_iterable(
+      itertools.combinations(s, r) for r in range(len(s)+1))
+
+
 # Use valgrind if requested
 if os.environ.get('TEST_WITH_VALGRIND', 'no').lower().strip() \
    not in ('no', 'false', '0'):
@@ -147,6 +154,8 @@ else:
 
 # Try to use local fusermount3
 os.environ['PATH'] = '%s:%s' % (pjoin(basename, 'util'), os.environ['PATH'])
+# Put example binaries on PATH
+os.environ['PATH'] = '%s:%s' % (pjoin(basename, 'example'), os.environ['PATH'])
 
 try:
     (fuse_proto, fuse_caps) = test_printcap()
--- a/util/meson.build
+++ b/util/meson.build
@@ -6,10 +6,12 @@ executable('fusermount3', ['fusermount.c', '../lib/mount_util.c'],
            install_dir: get_option('bindir'),
            c_args: '-DFUSE_CONF="@0@"'.format(fuseconf_path))
 
-executable('mount.fuse3', ['mount.fuse.c'], 
+executable('mount.fuse3', ['mount.fuse.c'],
            include_directories: include_dirs,
+           link_with: [ libfuse ],
            install: true,
-           install_dir: get_option('sbindir'))
+           install_dir: get_option('sbindir'),
+           c_args: '-DFUSE_USE_VERSION=33')
 
 
 udevrulesdir = get_option('udevrulesdir')
--- a/util/mount.fuse.c
+++ b/util/mount.fuse.c
@@ -14,6 +14,18 @@
 #include <unistd.h>
 #include <errno.h>
 #include <stdint.h>
+#include <fcntl.h>
+#include <pwd.h>
+#include <sys/wait.h>
+
+#ifdef linux
+#include <sys/prctl.h>
+#include <sys/syscall.h>
+#include <linux/capability.h>
+#include <linux/securebits.h>
+#endif
+
+#include "fuse.h"
 
 static char *progname;
 
@@ -80,6 +92,124 @@ static char *add_option(const char *opt, char *options)
 	return options;
 }
 
+static int prepare_fuse_fd(const char *mountpoint, const char* subtype,
+			   const char *options)
+{
+	int fuse_fd = -1;
+	int flags = -1;
+	int subtype_len = strlen(subtype) + 9;
+	char* options_copy = xrealloc(NULL, subtype_len);
+
+	snprintf(options_copy, subtype_len, "subtype=%s", subtype);
+	options_copy = add_option(options, options_copy);
+	fuse_fd = fuse_open_channel(mountpoint, options_copy);
+	if (fuse_fd == -1) {
+		exit(1);
+	}
+
+	flags = fcntl(fuse_fd, F_GETFD);
+	if (flags == -1 || fcntl(fuse_fd, F_SETFD, flags & ~FD_CLOEXEC) == 1) {
+		fprintf(stderr, "%s: Failed to clear CLOEXEC: %s\n",
+			progname, strerror(errno));
+		exit(1);
+	}
+
+	return fuse_fd;
+}
+
+#ifdef linux
+static uint64_t get_capabilities(void)
+{
+	/*
+	 * This invokes the capset syscall directly to avoid the libcap
+	 * dependency, which isn't really justified just for this.
+	 */
+	struct __user_cap_header_struct header = {
+		.version = _LINUX_CAPABILITY_VERSION_3,
+		.pid = 0,
+	};
+	struct __user_cap_data_struct data[2];
+	memset(data, 0, sizeof(data));
+	if (syscall(SYS_capget, &header, data) == -1) {
+		fprintf(stderr, "%s: Failed to get capabilities: %s\n",
+			progname, strerror(errno));
+		exit(1);
+	}
+
+	return data[0].effective | ((uint64_t) data[1].effective << 32);
+}
+
+static void set_capabilities(uint64_t caps)
+{
+	/*
+	 * This invokes the capset syscall directly to avoid the libcap
+	 * dependency, which isn't really justified just for this.
+	 */
+	struct __user_cap_header_struct header = {
+		.version = _LINUX_CAPABILITY_VERSION_3,
+		.pid = 0,
+	};
+	struct __user_cap_data_struct data[2];
+	memset(data, 0, sizeof(data));
+	data[0].effective = data[0].permitted = caps;
+	data[1].effective = data[1].permitted = caps >> 32;
+	if (syscall(SYS_capset, &header, data) == -1) {
+		fprintf(stderr, "%s: Failed to set capabilities: %s\n",
+			progname, strerror(errno));
+		exit(1);
+	}
+}
+
+static void drop_and_lock_capabilities(void)
+{
+	/* Set and lock securebits. */
+	if (prctl(PR_SET_SECUREBITS,
+		  SECBIT_KEEP_CAPS_LOCKED |
+		  SECBIT_NO_SETUID_FIXUP |
+		  SECBIT_NO_SETUID_FIXUP_LOCKED |
+		  SECBIT_NOROOT |
+		  SECBIT_NOROOT_LOCKED) == -1) {
+		fprintf(stderr, "%s: Failed to set securebits %s\n",
+			progname, strerror(errno));
+		exit(1);
+	}
+
+	/* Clear the capability bounding set. */
+	int cap;
+	for (cap = 0; ; cap++) {
+		int cap_status = prctl(PR_CAPBSET_READ, cap);
+		if (cap_status == 0) {
+			continue;
+		}
+		if (cap_status == -1 && errno == EINVAL) {
+			break;
+		}
+
+		if (cap_status != 1) {
+			fprintf(stderr,
+				"%s: Failed to get capability %u: %s\n",
+				progname, cap, strerror(errno));
+			exit(1);
+		}
+		if (prctl(PR_CAPBSET_DROP, cap) == -1) {
+			fprintf(stderr,
+				"%s: Failed to drop capability %u: %s\n",
+				progname, cap, strerror(errno));
+		}
+	}
+
+	/* Drop capabilities. */
+	set_capabilities(0);
+
+	/* Prevent re-acquisition of privileges. */
+	if (prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) == -1) {
+		fprintf(stderr, "%s: Failed to set no_new_privs: %s\n",
+			progname, strerror(errno));
+		exit(1);
+	}
+}
+#endif
+
 int main(int argc, char *argv[])
 {
 	char *type = NULL;
@@ -88,10 +218,12 @@ int main(int argc, char *argv[])
 	char *basename;
 	char *options = NULL;
 	char *command = NULL;
-	char *setuid = NULL;
+	char *setuid_name = NULL;
 	int i;
 	int dev = 1;
 	int suid = 1;
+	int pass_fuse_fd = 0;
+	int drop_privileges = 0;
 
 	progname = argv[0];
 	basename = strrchr(argv[0], '/');
@@ -167,7 +299,12 @@ int main(int argc, char *argv[])
 							      "_netdev",
 							      NULL};
 				if (strncmp(opt, "setuid=", 7) == 0) {
-					setuid = xstrdup(opt + 7);
+					setuid_name = xstrdup(opt + 7);
+					ignore = 1;
+				} else if (strcmp(opt,
+						  "drop_privileges") == 0) {
+					pass_fuse_fd = 1;
+					drop_privileges = 1;
 					ignore = 1;
 				}
 				for (j = 0; ignore_opts[j]; j++)
@@ -184,6 +321,16 @@ int main(int argc, char *argv[])
 				}
 				opt = strtok(NULL, ",");
 			}
+		}
+	}
+
+	if (drop_privileges) {
+		uint64_t required_caps = CAP_TO_MASK(CAP_SETPCAP) |
+				CAP_TO_MASK(CAP_SYS_ADMIN);
+		if ((get_capabilities() & required_caps) != required_caps) {
+			fprintf(stderr, "%s: drop_privileges was requested, which launches the FUSE file system fully unprivileged. In order to do so %s must be run with privileges, please invoke with CAP_SYS_ADMIN and CAP_SETPCAP (e.g. as root).\n",
+			progname, progname);
+			exit(1);
 		}
 	}
 
@@ -209,6 +356,51 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	if (setuid_name && setuid_name[0]) {
+#ifdef linux
+		if (drop_privileges) {
+			/*
+			 * Make securebits more permissive before calling
+			 * setuid(). Specifically, if SECBIT_KEEP_CAPS and
+			 * SECBIT_NO_SETUID_FIXUP weren't set, setuid() would
+			 * have the side effect of dropping all capabilities,
+			 * and we need to retain CAP_SETPCAP in order to drop
+			 * all privileges before exec().
+			 */
+			if (prctl(PR_SET_SECUREBITS,
+				  SECBIT_KEEP_CAPS |
+				  SECBIT_NO_SETUID_FIXUP) == -1) {
+				fprintf(stderr,
+					"%s: Failed to set securebits %s\n",
+					progname, strerror(errno));
+				exit(1);
+			}
+		}
+#endif
+
+		struct passwd *pwd = getpwnam(setuid_name);
+		if (setgid(pwd->pw_gid) == -1 || setuid(pwd->pw_uid) == -1) {
+			fprintf(stderr, "%s: Failed to setuid to %s: %s\n",
+				progname, setuid_name, strerror(errno));
+			exit(1);
+		}
+	} else if (!getenv("HOME")) {
+		/* Hack to make filesystems work in the boot environment */
+		setenv("HOME", "/root", 0);
+	}
+
+	if (pass_fuse_fd)  {
+		int fuse_fd = prepare_fuse_fd(mountpoint, type, options);
+		char *dev_fd_mountpoint = xrealloc(NULL, 20);
+		snprintf(dev_fd_mountpoint, 20, "/dev/fd/%u", fuse_fd);
+		mountpoint = dev_fd_mountpoint;
+	}
+
+#ifdef linux
+	if (drop_privileges) {
+		drop_and_lock_capabilities();
+	}
+#endif
 	add_arg(&command, type);
 	if (source)
 		add_arg(&command, source);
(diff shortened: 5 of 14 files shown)


lib/fuse_lowlevel.c
```diff
@@ -2890,16 +2890,38 @@ int fuse_session_mount(struct fuse_session *se, const char *mountpoint)
...
+		}
+		se->fd = STDIN_FILENO;
 
-	/* Save mountpoint */
-	se->mountpoint = strdup(mountpoint);
-	if (se->mountpoint == NULL)
-		goto error_out;
+		/* TODO: Check whether any mount options have been passed on the
+		 * command line. If so, print a diagnostic that these aren't
+		 * allowed in pre-mounted mode and bail out. */
```
saittam: Checking whether any mount options have been passed will require some changes to the option parser. I figured to first get feedback on the idea in general before diving into that.

lib/fuse_lowlevel.c
```diff
@@ -2890,16 +2890,38 @@ int fuse_session_mount(struct fuse_session *se, const char *mountpoint)
...
+		 * Pre-mounted mode: To allow FUSE helpers to run without
+		 * privileges, the caller may open /dev/fuse and mount the file
+		 * system before launching the helper. The helper doesn't need
+		 * to handle mounting in this case, which is reflected by the
+		 * mountpoint parameter being the empty string. The /dev/fuse FD
+		 * is expected to be open on file descriptor 0 in that case.
+		 */
+		struct stat st;
+		if (fstat(STDIN_FILENO, &st) == -1 || !S_ISCHR(st.st_mode) ||
+		    isatty(STDIN_FILENO)) {
```
thejh: I think you could also check for `major(st.st_dev)==MISC_MAJOR && minor(st.st_dev)==FUSE_MINOR`

Nikratio:
I'm a little confused by the goal here. Normally the "suitable privileged process" is fusermount, so the filesystem does not need any special permissions. Why doesn't that work for you?

saittam:
Regarding why fusermount doesn't work for us: fusermount requires setuid (technically, you could make it work by setting CAP_SYS_ADMIN in fscaps, but that doesn't change much). setuid (and fscaps just as well) is known to be risky since it allows unprivileged processes to gain privilege. Using setuid in a bullet-proof way is practically impossible. The conceptual problem is that there's too much attack surface in startup code of the setuid binary that the unprivileged process can freely access. I usually cite https://www.qualys.com/2017/06/19/stack-clash/stack-clash.txt as an example to demonstrate how hopeless the situation is ;-) Given that fusermount is dynamically linked, chances are pretty high that a variant of the attack described in the qualys writeup linked above would have worked against it (I haven't tried though).

Given the above background, Chrome OS has adopted the general policy to ban setuid and fscaps binaries in general. This means that we'd like to avoid fusermount if we can. Our typical approach is to use no_new_privs to prevent a process or its children to ever regain privileges that it has dropped (capabilities + securebits as used in my demo commands are a weaker form of this idea). This not only defends against the issues above, but also makes it much simpler to argue about security at a high level - privilege can no longer be created from thin air, so you can be pretty confident that certain things just can't happen in certain parts of the system.

Of course, this sometimes implies architecture changes to reorder or move privileged operations to different processes. A standard pattern is to have a privileged parent process that handles all privileged operations and isolate the code that's exposed to attacks into a heavily sandboxed child process. This is exactly what this patch allows, i.e. the privileged parent process takes care of mounting, then spawns the helper and setting up things such that the helper or its children can't gain privilege. "suitably privileged process" was meant to refer to the parent that spawns the FUSE binary, not a child of the FUSE binary (as would be the case with setuid). Sorry for being unclear about this.

I hope the above helps clarify things, let me know if there are further questions!

Nikratio:
Thanks for the explanations, this sounds reasonable. My main concern is potential incompatibility (or the lack of coordination) with the plans described in issue #77.

szmi:
Interesting work.  Yes, there's interaction with issue #77, so if you want to be "future compatible" with this, then don't design it with "pre-mounted" in mind.  The generic sequence should be:

 1) grab device descriptor
 2) start unprivileged fuse daemon in the background
 3) do the mounting

Problem with that sequence is that with current kernels you'll get EPERM when reading the device before it's mounted.   Not nice.  So lets start with fixing the kernel not to do that.

saittam:
So I've read through #77 - let me recap the gist of it to make sure I understand correctly. The current state of things is this:

1. fd = open(/dev/fuse)
2. mount(fd)
3. start loop to read(fd) and process events

The problem is that mount() may trigger code in the kernel that would like to access the file system, which is not currently possible until after mount() completes and the FUSE daemon starts processing events. So the proposal is to change things to this instead:

1. fd = open(/dev/fuse)
2. start loop to read(fd) and process events
3. mount(fd)

Correct? 

Side note: My immediate concern with this change are concurrency implications for userspace. IIUC you need to be able to process events while the thread that called mount() is still hanging in the syscall? Or would mount() return early before the mount has actually completed fully (that feels odd too, but I can't claim I know mount() syscall semantics well enough to be able to judge).

Regarding implications to my patch: I don't think there is anything that is fundamentally incompatible. The only thing that changes is that the privileged parent process needs to perform the mount and helper-spawning steps in reverse order (as already noted by Miklos above). That's simple enough, and when switching to the new world the privileged parent can send the PREINIT notification as proposed in #77, then spawn the FUSE daemon, then mount() and things will be fine. Arguably, spawing the FUSE daemon in a separate process might even simplify things for #77 since it naturally allows for concurrency (assuming my concern above is actually valid).

Given the backwards compatibility concerns aren't any different with/without my patch, and that whatever backwards compatibility solution you decide on for #77 will almost certainly work just as well with a spawned FUSE helper, I'm hoping we don't need to block this until the solution to #77 to materializes.

Regarding naming, I chose the "pre-mounted" term only after writing the code for lack of a better description of the idea :-D I'm happy to change naming of course. Maybe call the new FUSE daemon mode "mount-less operation"? I'm having a hard time to come up with something that is both descriptive and short enough, open for suggestions.

szmi:
Re concurrency: normally (no SElinux setup involved) it woudn't be needed.  So a serialized setup would look like:

 1) open /dev/fuse
 2) read init request
 3) write init reply
 4) mount
 5) start loop processing requests

Then there's the requirements from SElinux.  Not sure how to handle that, we could just require the processing loop to start in the background.

Or we could keep the serialized setup by preloading the security xattrs before starting the mount.

In any case the kernel should definitely handle the non-serialized case as well as the serialized one, so we could have mount(2) being called concurrently with I/O on the device descriptor and things should works fine.  As well as handling the legacy sequence, of course.

lib/fuse_lowlevel.c
```diff
@@ -2890,16 +2890,38 @@ int fuse_session_mount(struct fuse_session *se, const char *mountpoint)
...
+			fprintf(stderr ,"fuse: File descriptor 0 must be open on /dev/fuse for pre-mounted mode.\n");
+			return -1;
+		}
+		se->fd = STDIN_FILENO;
```
saittam: Note to self: Using STDIN might not be the best idea, since it'll cause problems with FUSE binaries that expect to read input from the user (sshfs for example).

saittam:
Thanks Miklos for clarifying, sounds like the concurrent mount() + FUSE daemon initializing and processing I/O on the FUSE fd in a subprocess should work fine then and this PR is generally compatible with #77. Let me know in case there are any remaining concerns I've missed.

How do we want to proceed here? I have added a couple comments in the code that you might want to consider, and I'm happy to spin a new patch version with naming changes and whatever other improvements you'd like to see.

szmi:
Okay, so aesthetically the empty mountpoint argument is pretty bad.  Could we have an explicit "--no-mount" argument instead?

Also agree about not using the standard I/O descriptors.  Perhaps explicit argument "--comm-fd=NN" or whatever.

saittam:
Thanks for the feedback! Single --mount-fd option sounds good, and I'll avoid dedicated feature naming as suggested. Will uploaded a new version of the patch that incorporates the feedback.

Regarding duplication of the mounting logic in the parent process, I have actually already thought about this from a somewhat different angle: I realized that adding support for the new mechanism in mount.fuse3 would be a good idea since it is a privileged parent process. If we also add in some basic support for dropping privileges, we get something that allows people to run fully deprivileged FUSE daemons for fstab mounts.

I have hacked together a proof of concept of this yesterday night, now uploaded here: https://github.com/saittam/libfuse/commit/c2a2a7471020f41ec26284af2924bfa2a2014156 The code is super hacky at this point, but it already reuses parse_mount_opts and fuse_kern_mount for setting up the mount in the parent process, so providing a proper library function for this shouldn't be too hard. Are you OK with mount.fuse linking to libfuse.so though? Currently it doesn't and there's a comment at the top of the util/meson.build file that states the goal for stuff in /util is to be kept "as standalone as possible".

saittam:
Interesting discovery: lib/mount_bsd.c already supports what I'm trying to achieve via the FUSE_DEV_FD to pass the file descriptor and FUSE_NO_MOUNT or empty mountpoint parameter to bypass mounting. Maybe we should lift this functionality from lib/mount_bsd.c to lib/fuse_lowlevel.c instead of implementing the same thing with a different interface? Then again, command-line switches are probably simpler to use in some situations than environment variables. WDYT?

saittam:
Uploaded new patch version and adjusted pull request title to reflect new commit message.

Nikratio:
I think extending `mount.fuse3` to do this is a great idea. There is a bit of an impedance mismatch though. `mount.fuse3` is supposed to work with any fuse filesystem. But libfuse does not prescribe a specific way for filesystems to handle their command line parsing (it used to do this in the past, but I really want to get away from that). So there is no guarantee that if `mount.fuse3` passes a `--mount-fd` parameter to the filesystem binary, it will be understood. That means at they very least mount.fuse3 will need to get a --use-mountfd parameter that controls the use of this feature. Using an environment variable, on the other hand, would work with any FUSE filesystem. So maybe it's worth to reconsider.. any thoughts?

saittam:
Regarding mount.fuse3 integration vs. FUSE file system command line handling: Good point. A middle ground might be to pass mount_fd as a mount option (or just reuse the existing fd option that we pass to the kernel?); it should be safe to assume that all FUSE file system implementations will have a way for the user to specify arbitrary mount options and pass them on to libfuse? My gut feeling is that a command line parameter or a mount option are slightly preferable over environment variables since they tend to be easier to discover, but I really don't feel strongly. Let me know what your preference is.

Ack on the additional background regarding util/meson.build comment and BSD situation, useful context to have.

Nikratio:
If we go with reusing the "fd" mount option, is there a way to check if the given fd matches the given mountpoint?

I dislike having a mount option make the mountpoint parameter obsolete (feels like a laying violation). If we could instead check if the fd is consistent with the specified path and mount options, that would make for a nicer experience.

saittam:
Regarding checking whether the /dev/fuse FD is associate with a given mount:
1. stat() on the mount gives you the device number, which is reflected in /sys/fs/fuse/connections/N.
2. I'm not aware of a way to get the device number for a /dev/fuse FD though. The kernel has that information, but I don't see anything that exposes it. How does one know which mount the entries in /sys/fs/fuse/connections refer to?
3. In the presence of bind mounts, there can be multiple points where (a subtree of) the file system is mounted. So the /dev/fuse file descriptor <-> mountpoints relationship is actually a complicated one in general.
4. It's possible that the FUSE filesystem daemon doesn't even see the mountpoint(s) it is mounted at. This can happen if you're running the daemon in a mount namespace and/or use pivot_root to constrain what it can see in the file system. In fact, I'd like to do so for Chrome OS - less stuff a process can see in the FS generally means lower risk.
Bottom line: I'm doubtful whether the consistency check you proposed is meaningful.

As a side note, it is interesting that obsoleting the mountpoint parameter feels like a layering violation to you (in which sense?). IMHO, the file system implementation knowing about higher-layer concerns such as where it's mounted is a layering violation in itself. I'm very aware that my perspective is probably biased by my background though!

Here's another idea that came to mind while thinking about this: Instead of the --mount-fd option, we could also special-case mountpoints of the form "/dev/fd/%u" to refer to an already opened /dev/fuse file descriptor. Would that make more sense to you?

davies:
If I understand it correctly, this patch opens a door for zero-downtime upgrading (spawn the new version and pass the fd to it, then stop reading new requests and flush all buffers,  then tell child process to process new requests), that's very useful, thanks!

Nikratio:
The layering violation that bugs me is unrelated to FUSE, it only relates to command line parsing. I feel that the number of required arguments (i.e., stuff without leading dashes) should not depend on what parameters (i.e., stuff with leading dashes) are specified. The other way around is fine, and quite usual for subcommands. This bugs me somewhat about the `--mount-fd` option, but even more about the `fd` mount option: in that case it is the occurence of a string in the *value* of a parameter that changes the syntax of the whole command line.

My preference is therefore to keep the syntax unchanged (i.e., always require a mount point argument). But that only makes sense if we can make sure that the fd is consistent with the mountpoint, which it seems we sadly cannot do.

The `/dev/fd/%u` suggestion nicely solves the syntax problem, but instead introduces a semantic one - it's no longer clear what the argument actualy specifies.

I can't offer any better alternatives though, and I don't think this should block the patch, so feel free to choose either the `/dev/fd` approach or the `-o fd=%d` approach, depending on what makes cleaner code.

saittam:
Thanks for elaborating on your concerns regarding the mountpoint parameter, I get your point now. In the end, I guess there's no denying that if we're not actually mounting anything, a mountpoint parameter is technically out of place ;-) We could make a separate command line syntax that does not take any mountpoint, but given libfuse doesn't parse the command line directly I think it's better to stay compatible with the current syntax.

All things considered, the `/dev/fd/%u` option seems like the best compromise to me. As you note, it doesn't actually specify a mount point, but it refers to the object that is mounted (i.e. the `/dev/fuse` file descriptor), so it can be thought of as a generalization of mountpoint to "kernel file system handle". A traditional mountpoint argument would merely allow for that handle to be created on the spot. I think it should also allow to support Miklos' proposed filesystem file descriptors (as described here: https://lwn.net/Articles/718638/) if/when that change appears in the kernel.

Regarding zero-downtime upgrading: That's an interesting application indeed, and I'm not aware of conceptual problems in my patch that would prevent this. I have no idea how much state a typical FUSE file system implementation holds on the userspace side though, but assuming flushing and reloading as suggested in @davies reply is practical, that might not be a big deal. Another potential complication might be the initialization handshake, which will not reoccur if you re-use an existing FUSE file descriptor (not sure whether that is something FUSE file systems can be made to deal with?). Another missing piece you need is to get the FUSE file system to pass its existing file descriptor. libfuse library support could solve that, or alternatively you could just connect via ptrace() to the existing process and make it exec() the new version :-p

saittam:
New version is up that addresses all feedback thus far to the best of my knowledge, so this should be ready for finer-grained code review now.

util/mount.fuse.c
```diff
@@ -80,6 +93,116 @@ static char *add_option(const char *opt, char *options)
...
+			&args, &dummy_ops, sizeof(dummy_ops), NULL);
+	free(options_copy);
+	if (session == NULL || fuse_session_mount(session, mountpoint) == -1) {
+		exit(1);
+	}
+
+	/*
+	 * Duplicate the FUSE file descriptor to obtain a copy that's still
+	 * valid after fuse_session_destroy closes its copy. Note that this
+	 * conveniently also gives us a file descriptor with the CLOEXEC flag
```
jlucangelio: I'd say "without the CLOEXEC flag set", but at least "with the CLOEXEC flag cleared". I think my brain reads "with the CLOEXEC flag" and assumes the flag is *set*.

util/mount.fuse.c
```diff
@@ -80,6 +93,116 @@ static char *add_option(const char *opt, char *options)
...
+	data[0].effective = data[0].permitted = caps;
+	data[1].effective = data[1].permitted = caps >> 32;
+	if (syscall(SYS_capset, &header, data) == -1) {
+		fprintf(stderr, "%s: Failed to drop capabilities: %s\n",
+			progname, strerror(errno));
+		exit(1);
+	}
+}
+
+static void drop_privileges(void)
```
jlucangelio: This will only work if launched from something where the bits are not already locked -- not sure if that's a problem.

saittam: That, and we need CAP_SETPCAP. This is all behind an option though, so people who do their own sandboxing aren't forced to use it. There might be a point to offer an option that just enables pass_fuse_fd, but doesn't attempt to drop privileges... I guess this can be added once there's demand for it?

util/mount.fuse.c
saittam: Fixed locally and will be in next upload.

Nikratio:
*ping* Are you no longer interested in working on this?

The only thing that prevents a merge is the missing testcase...

saittam:
I still intend to finish this, but was busy with other stuff, sorry. I have
a test case working, need to polish and upload the code, hopefully early
next week.

util/meson.build
```diff
@@ -16,10 +16,12 @@ executable('fusermount3', ['fusermount.c', mount_util_c],
...
            c_args: '-DFUSE_CONF="@0@"'.format(fuseconf_path))
 
-executable('mount.fuse3', ['mount.fuse.c'], 
+executable('mount.fuse3', ['mount.fuse.c'],
            include_directories: include_dirs,
+           link_with: [ libfuse ],
            install: true,
-           install_dir: get_option('sbindir'))
+           install_dir: get_option('sbindir'),
+           c_args: '-DFUSE_USE_VERSION=32')
```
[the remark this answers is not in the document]
saittam: Done.

saittam:
Looks like the test is failing because it can't set SECUREBITS. Might be a CI environment issue? Is there a way to reproduce what the CI is doing locally so I can nail down what's failing?

Nikratio:
No, I don't know what Travis is doing exactly. We will probably have to disable these tests on Travis - assuming that they work fine locally.

ChangeLog.rst
```diff
@@ -8,6 +8,14 @@ Unreleased Changes
 * The description of the FUSE_CAP_READDIRPLUS_AUTO flag has been
   improved.
 
+* Allow open `/dev/fuse` file descriptors to be passed via mountpoints of the
+  special format `/dev/fd/%u`. This allows mounting to be handled by the parent
+  so the FUSE filesystem process can run fully unprivileged.
+
+* Add an `unprivileged` option to mount.fuse3 which causes it open `/dev/fuse`
```
Nikratio: "causes it *to* open"

Nikratio:
...however, it fails for me locally as well:

```
test/test_examples.py::test_hello[hello-options0-invoke_mount_fuse_unprivileged] FAILED                                                                                                [ 35%]

========================================================================================== FAILURES ==========================================================================================
_________________________________________________________________ test_hello[hello-options0-invoke_mount_fuse_unprivileged] __________________________________________________________________
Traceback (most recent call last):
  File "/home/nikratio/in-progress/libfuse/build/test/test_examples.py", line 64, in test_hello
    wait_for_mount(mount_process, mnt_dir)
  File "/home/nikratio/in-progress/libfuse/build/test/util.py", line 42, in wait_for_mount
    pytest.fail('file system process terminated prematurely')
  File "/home/nikratio/.local/lib/python3.5/site-packages/_pytest/outcomes.py", line 104, in fail
    raise Failed(msg=msg, pytrace=pytrace)
Failed: file system process terminated prematurely
------------------------------------------------------------------------------------ Captured stderr call ------------------------------------------------------------------------------------
/home/nikratio/in-progress/libfuse/build/test/../util/mount.fuse3: Failed to set securebits Operation not permitted
================================================================================== short test summary info ===================================================================================
FAIL test/test_examples.py::test_hello[hello-options0-invoke_mount_fuse_unprivileged]
====================================================================================== warnings summary ======================================================================================
```

This is on Debian stable, not a container, no special hardening settings.

Nikratio:
Might it require root permissions?

test/util.py
```diff
@@ -147,6 +154,11 @@ def fuse_test_marker():
 
 # Try to use local fusermount3
 os.environ['PATH'] = '%s:%s' % (pjoin(basename, 'util'), os.environ['PATH'])
+# Put example binaries on PATH
+os.environ['PATH'] = '%s:%s' % (pjoin(basename, 'example'), os.environ['PATH'])
```
Nikratio: Hmm. Does this mean mounting through mount.fuse3 will only ever work for filesystems in $PATH?

test/util.py
```diff
@@ -147,6 +154,11 @@ def fuse_test_marker():
...
+
+# Enable shared libraries to be resolved from build directory
+os.environ['LD_LIBRARY_PATH'] = pjoin(basename, 'lib')
```
Nikratio: This should not be necessary, I think. Why is it needed now? The examples are linked against libfuse3 and do not need an specific LIBRARY_PATH

test/util.py
saittam: Note that I haven't changed this, IIUC mount.fuse has always worked like this. It eventually launches /bin/sh -c ${command}, so ${command} must be on PATH. My understanding is that when you do mount -t fuse3 ${source} /mnt/mountpoint (also if invoked via fstab), mount looks for the mount.fuse3 binary in /sbin and a couple other locations, then mount mount.fuse3 looks for ${source} on paths and executes it.

test/util.py
saittam: You are right - I must have mistakenly added this when trying to get mount.fuse3 to successfully execute the hello test binary. Removing.

ChangeLog.rst
saittam: Fixing.

saittam:
> Might it require root permissions?

It requires capabilities, yes. The failure is either because it doesn't have CAP_SETPCAP, or because SECUREBITS are already locked. Note that the assumption is that mount.fuse3 is invoked with privileges (otherwise it couldn't mount either!).

What is the command line that used to invoke the test? It passes for me when invoked like this:
  sudo python3 -m pytest build/test/

Can you also try capsh --print in your shell (and under sudo in case you're using that to launch the test) to see how securebits and the capabilities look like? I'll see whether I can reproduce the travis environment to nail this down.

Nikratio:
> > Might it require root permissions?
> 
> It requires capabilities, yes. The failure is either because it doesn't have CAP_SETPCAP, or because SECUREBITS are already locked. Note that the assumption is that mount.fuse3 is invoked with privileges (otherwise it couldn't mount either!).

Not sure I agree. Currently an ordinary user can call mount.fuse3 and things will work (assuming that fusermount3 is setuid). For the un-initiated it will be quite confusing that passing the `unprivileged` option means that mount.fuse3 requires *more* privileges. Can we detect this case and print a more helpful error message?

Also, could you add this option to `doc/mount.fuse3.8` as well? (Apologies, this really is maintained in plain nroff...).

> What is the command line that used to invoke the test? It passes for me when invoked like this:
> sudo python3 -m pytest build/test/
> 
> Can you also try capsh --print in your shell (and under sudo in case you're using that to launch the test) to see how securebits and the capabilities look like? I'll see whether I can reproduce the travis environment to nail this down.

No need, as root it runs just fine. This is also why the Travis tests fail, most of them run as regular user. Could you skip the unprivileged test when there's insufficient privileges? (to skip, just call `pytest.skip(reason)`).

saittam:
> > > Might it require root permissions?
> > 
> > 
> > It requires capabilities, yes. The failure is either because it doesn't have CAP_SETPCAP, or because SECUREBITS are already locked. Note that the assumption is that mount.fuse3 is invoked with privileges (otherwise it couldn't mount either!).
> 
> Not sure I agree. Currently an ordinary user can call mount.fuse3 and things will work (assuming that fusermount3 is setuid). For the un-initiated it will be quite confusing that passing the `unprivileged` option means that mount.fuse3 requires _more_ privileges. Can we detect this case and print a more helpful error message?

Done. Also, I've renamed the "unprivileged" option to "drop_privileges", which makes it clearer that mount.fuse expects to have privileges and then drop them.

> 
> Also, could you add this option to `doc/mount.fuse3.8` as well? (Apologies, this really is maintained in plain nroff...).

Done, created a new section detailing options handled by mount.fuse. While at it, I also documented the setuid option.

> 
> > What is the command line that used to invoke the test? It passes for me when invoked like this:
> > sudo python3 -m pytest build/test/
> > Can you also try capsh --print in your shell (and under sudo in case you're using that to launch the test) to see how securebits and the capabilities look like? I'll see whether I can reproduce the travis environment to nail this down.
> 
> No need, as root it runs just fine. This is also why the Travis tests fail, most of them run as regular user. Could you skip the unprivileged test when there's insufficient privileges? (to skip, just call `pytest.skip(reason)`).

Done.

Nikratio:
Great, thanks for being so response!

The test cases now fail for a different reason. I'm struggling to interpret the error though (`/bin/sh: 1: hello: Permission denied`). Does that tell you anything? It might well be that the Travis environment doesn't like some of the privilege dropping stuff, but this error doesn't sound like that...

saittam:
> Great, thanks for being so response!
> 
> The test cases now fail for a different reason. I'm struggling to interpret the error though (`/bin/sh: 1: hello: Permission denied`). Does that tell you anything? It might well be that the Travis environment doesn't like some of the privilege dropping stuff, but this error doesn't sound like that...

Stumped me as well, but a bit of local testing suggests that the problem might be that root can't exec the hello,hello_ll binaries for the case where it runs without CAP_DAC_OVERRIDE and other permission bits don't allow execution. I've just uploaded a new version with a speculative fix for that (setting umask in the travis script).

saittam:
> > Great, thanks for being so response!
> > The test cases now fail for a different reason. I'm struggling to interpret the error though (`/bin/sh: 1: hello: Permission denied`). Does that tell you anything? It might well be that the Travis environment doesn't like some of the privilege dropping stuff, but this error doesn't sound like that...
> 
> Stumped me as well, but a bit of local testing suggests that the problem might be that root can't exec the hello,hello_ll binaries for the case where it runs without CAP_DAC_OVERRIDE and other permission bits don't allow execution. I've just uploaded a new version with a speculative fix for that (setting umask in the travis script).

Turns out this didn't resolve it. Let me push a version that prints some diagnostics to hopefully find out more...

saittam:
> > > Great, thanks for being so response!
> > > The test cases now fail for a different reason. I'm struggling to interpret the error though (`/bin/sh: 1: hello: Permission denied`). Does that tell you anything? It might well be that the Travis environment doesn't like some of the privilege dropping stuff, but this error doesn't sound like that...
> > 
> > 
> > Stumped me as well, but a bit of local testing suggests that the problem might be that root can't exec the hello,hello_ll binaries for the case where it runs without CAP_DAC_OVERRIDE and other permission bits don't allow execution. I've just uploaded a new version with a speculative fix for that (setting umask in the travis script).
> 
> Turns out this didn't resolve it. Let me push a version that prints some diagnostics to hopefully find out more...

I'm still not any wiser unfortunately. I'll let this sink in and take a fresh look tomorrow.

Nikratio:
> > > > Great, thanks for being so response!
> > > > The test cases now fail for a different reason. I'm struggling to interpret the error though (`/bin/sh: 1: hello: Permission denied`). Does that tell you anything? It might well be that the Travis environment doesn't like some of the privilege dropping stuff, but this error doesn't sound like that...
> > > 
> > > 
> > > Stumped me as well, but a bit of local testing suggests that the problem might be that root can't exec the hello,hello_ll binaries for the case where it runs without CAP_DAC_OVERRIDE and other permission bits don't allow execution. I've just uploaded a new version with a speculative fix for that (setting umask in the travis script).
> > 
> > 
> > Turns out this didn't resolve it. Let me push a version that prints some diagnostics to hopefully find out more...
> 
> I'm still not any wiser unfortunately. I'll let this sink in and take a fresh look tomorrow.

As long as it works locally for you and me, I'm fine with merging it even if the CI complains. Let's just make sure that the tests are skipped on the CI then instead of failing.

Nikratio:
Awesome!

saittam:
Excellent! Thanks for your patience on this.

PR #296, opened 2018-09-06
Add build options for utils and examples

t-chaik:
Allow skipping utils build & installation (-Dutils=false) and examples
build (-Dexamples=false). By default behaviour is unchanged (both are
true: utils and examples get build).


Nikratio:
Thanks!

PR #300, opened 2018-10-15
I faced the problem with meson build version.

Chethangc:
meson build version was old and if we upgrade the OS then no build version error 
so keep your OS updated and upgraded for error less installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ explicitly using a separate set of API functions.
 Supported Platforms
 -------------------
 
-* Linux (fully)
+* Linux (fully) 
 * BSD (mostly/best-effort)
 * For OS-X, please use [OSXFUSE](https://osxfuse.github.io/)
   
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks for the patch! The instructions are a little unclear though - what does it mean that the "operating system is upgraded"? This is a little too general to be useful, especially since libfuse should not require upgrading the entire operating system. Do you remember what exactly you had to upgrade?

Chethangc:
Sorry for the unclear instruction,I had a meson build error ie 0.45 or
higher version is required in Ubuntu 16.10 and 17.10

I was not able to update the meson.
Because for 16.04 the version is 0.42.
I upgraded my operating system to - 18.04 it executed without error.
Meson version changed and its error free installation.

I mean to say that uprade your existing os to 18.04 or higher when ever you
face a problem in installation of fuse.
      Don't try to install in older version of Ubuntu it may give you error.

Chethangc:
As per my knowledge it requires entire os upgrade is required.
Meson is platform dependent.
Installed meson version
16.04- 0.42 build error
17.14- 0.44 error while pytest same 0.45 or higher is required.
18.04 error free installation.

In a class of 60 almost 40 students who have not upgraded os got this
errors.

Nikratio:
Thanks for clarifying! I have just updated the meson.build file accordingly.

PR #302, opened 2018-10-16
When in auto_unmount mode, check filesystem type before unmounting.

kevin-vigor:
This addresses a problem when a fuse filesystem using auto_unmount is mounted
by autofs. Autofs uses a stacked mount approach, mounting the mountpoint as
type 'autofs'; then when the mountpoint is accessed the fuse filesystem is
mounted over the autofs mount.

Prior to this change, when such a filesystem was unmounted by 'umount',
the mountpoint would be unmounted twice, once by autofs, once by fusermount.
This resulted in the autofs mount being umounted, and autofs was broken
thereafter.

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -237,6 +237,7 @@ static int check_is_mount_child(void *p)
 	const char **a = p;
 	const char *last = a[0];
 	const char *mnt = a[1];
+	const char *type = a[2];
 	int res;
 	const char *procmounts = "/proc/mounts";
 	int found;
@@ -284,7 +285,8 @@ static int check_is_mount_child(void *p)
 			continue;
 		}
 		if (entp->mnt_dir[0] == '/' &&
-		    strcmp(entp->mnt_dir + 1, last) == 0) {
+		    strcmp(entp->mnt_dir + 1, last) == 0 &&
+		    (!type || strcmp(entp->mnt_type, type) == 0)) {
 			found = 1;
 			break;
 		}
@@ -317,11 +319,11 @@ static pid_t clone_newns(void *a)
 #endif
 }
 
-static int check_is_mount(const char *last, const char *mnt)
+static int check_is_mount(const char *last, const char *mnt, const char *type)
 {
 	pid_t pid, p;
 	int status;
-	const char *a[2] = { last, mnt };
+	const char *a[3] = { last, mnt, type };
 
 	pid = clone_newns((void *) a);
 	if (pid == (pid_t) -1) {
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks for the patch! Just to make sure, when you wrote *unmounted by umount*, you mean *unmounted by fusermount -u*, right? Otherwise I do not understand how making changes to fusermount will affect the situation.

kevin-vigor:
Hi Nikolaus! No, I actually do mean 'unmounted by umount' (but the same problem occurs when you unmount using fusermount -u).

In auto_unmount mode, libfuse spawns an instance of fusermount to watch the filesystem process when mounting  (see fuse_mount_fusermount()). When the filesystem process dies the watching fusermount process then unmounts the mountpoint (see the auto_unmount case in util/fusermount.c:main()).

This works properly when the filesystem process dies due to a crash or something which leaves the mount in place (the watching fusermount cleans up the mountpoint, preventing the dreaded 'transport endpoint not connected' error). But when the filesystem process is cleanly unmounted, then the watching fusermount unmounts the mountpoint a second time, which is usually not a problem (it just fails silently), but breaks when running under autofs.

I have tested this patch by unmounting the filesystem using umount, fusermount -u, and by simply killing the filesystem process. It works properly in all cases (only the last one worked before).

FWIW a version of this patch has been deployed inside Facebook for several years so it has had a *lot* of test cycles.

Nikratio:
Ah, got you. I somehow managed to overlook the auto_unmount part. 

I am not sure I like this approach though. It fixes the issue for autofs, but what about other situations were the filesystem is mounted on top of an existing mountpoint (especially another FUSE one). Wouldn't that still result in a double unmount? Would it maybe a better approach to unmount *only* if accessing the mountpoint gives a "Transport endpoint not connected" error?

kevin-vigor:
Yeah, this is a partial solution for sure, double unmounts can still occur. It does solve a large class of problems, though, and I was trying to be conservative and change existing behaviour (unmount all the things!) as little as possible.

But I will happily implement the 'unmount only on transport endpoint error' version, if you are willing to accept it.

kevin-vigor:
Updated patch to check that mount is in broken (ENOTCONN) state before unmounting, as well as the filesystem type check previously posted. 

As a bonus, this now only affects the auto_unmount case; the prior version of this diff altered behaviour of 'fusermount -u' which seems undesirable.

Tested unmounting with umount, fusermount -u, fusermount3 -u, and by kill -9'ing the filesystem process. In all cases the fuse filesystem was unmounted and the underlying autofs filesystem remained mounted.

Nikratio:
Thanks, I like this much better. But is there a reason to retain the additional type check? Carrying this around everywhere makes the good more complicated.

Also, please add a line to ChangeLog.rst

kevin-vigor:
Unfortunately I believe it is still necessary to check the filesystem type. The problem is that if autofs is in use and a normal unmount occurs, autofs is the active mount when fusermount runs the cleanup code. If we then try to access the mountpoint to see it is is returning ENOTCONN, autofs will happily mount the filesystem again. The result is that you can never actually unmount because the unmount process causes a new mount.

I attempted to explain this in the comment block preceding should_auto_unmount().

Changelog entry has been added.

PR #305, opened 2018-10-18
Meson test when libfuse is a subproject

ahmed-masud:
A small change to suppress error from `test/wrong_command.c` when libfuse is a subproject, avoids cluttering of test reports in the parent project.


ahmed-masud:
I think i screwed this up, will create another one

PR #310, opened 2018-10-21
Avoid using MESON_SOURCE_ROOT in install script

t-chaik:
When libfuse is used as a meson subproject, `MESON_SOURCE_ROOT` is set to
the parent package source directory in install script environment, thus
breaking source file references. There doesn't seem to be a preferred
way of accessing the current project's source directory from such a script
at the moment [1]. This patch works around this by passing the current
project source directory as an argument of the script.

[1] https://github.com/mesonbuild/meson/issues/4399

### DIFF
--- a/util/install_helper.sh
+++ b/util/install_helper.sh
@@ -6,9 +6,10 @@
 
 set -e
 
-sysconfdir="$1"
-bindir="$2"
-udevrulesdir="$3"
+utildir="$1"
+sysconfdir="$2"
+bindir="$3"
+udevrulesdir="$4"
 
 # Both sysconfdir and bindir are absolute paths (since they are joined
 # with --prefix in meson.build), but need to be interpreted relative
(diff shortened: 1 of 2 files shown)


Nikratio:
Hmm. What is the context here? It seems odd to me that you'd want to run the install_helper script at all if you're using libfuse a s a sub-project. I'd think the only reason to do that if you want to link against the library...

t-chaik:
I'm not a FUSE expert, so please correct me if I'm wrong, but isn't a set-uid root installed `fusermount3` required for unprivileged user mount and unmount operations?

Nikratio:
Yes. But that would normally be provided by a standalone libfuse installation, not when it's used as a subproject. Do you have a specific application that uses libfuse as a subproject and wants to install a `fusermount`? If so, why?

t-chaik:
[BuildBox](https://gitlab.com/BuildStream/buildbox) is a tool that allows running commands in an unprivileged sandbox using [Bubblewrap](https://github.com/projectatomic/bubblewrap) ontop of a FUSE provided file system. It builds using meson, has libfuse dependency and mount/unmount FUSE filesystems at runtime. Building libfuse as a subproject would be of great help on distros where fuse3 is not available, but is less useful if `fusermount3` is not installed along.

Nikratio:
I would be very unhappy if installing Buildbox would end-up installing a general purpose setuid binary in /sbin. Also, the installation instructions (https://gitlab.com/BuildStream/buildbox/blob/master/INSTALL.rst) actually say to install the dependencies separately. So... I'm still not convinced that this is a good idea (and apparently the BuildBox dev's aren't either).

Nikratio:
Closing this for the reasons given above. Feel free to re-open if there's anything I haven't considered.

PR #314, opened 2018-10-24
* Fixed ninja test / meson test when libfuse is a sub-project

ahmed-masud:
Modified test/meson.build and test/wrong_command.c to report a
skip to remove clutter from parent project running meson test harness


Nikratio:
Already merged - guess this was an accidental reopen?

PR #321, opened 2018-11-11
Fix build and mounting on FreeBSD

novel:
Please refer to commit messages of individual commits for more details.

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -1,10 +1,10 @@
 libfuse_sources = ['fuse.c', 'fuse_i.h', 'fuse_loop.c', 'fuse_loop_mt.c',
                    'fuse_lowlevel.c', 'fuse_misc.h', 'fuse_opt.c',
                    'fuse_signals.c', 'buffer.c', 'cuse_lowlevel.c',
-	           'helper.c', 'modules/subdir.c' ]
+                   'helper.c', 'modules/subdir.c', 'mount_util.c' ]
 
 if host_machine.system().startswith('linux')
-   libfuse_sources += [ 'mount.c', 'mount_util.c' ]
+   libfuse_sources += [ 'mount.c' ]
 else
    libfuse_sources += [ 'mount_bsd.c' ]
 endif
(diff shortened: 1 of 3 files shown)


novel:
I thought it's Linux specific because it's only built on Linux:

```
if host_machine.system().startswith('linux')
   libfuse_sources += [ 'mount.c', 'mount_util.c' ]
else
   libfuse_sources += [ 'mount_bsd.c' ]
endif
```

I thought that it could be a mistake (also I've seen some ifndefs __NetBSD__ there), but in fact it doesn't build if I add it to sources list on FreeBSD. I'll take a look, probably I'll be able to fix it without introducing a new file.

novel:
Updated.

stesser:
I'm not opening a new issue since I think the following applies to this one ... (let me know if you prefer a seperate issue for tracking purposes).

I had to add the following diff to make libfuse build on FreeBSD, since the mount_util.c is not compiled on FreeBSD and it contains the body of fuse_mnt_parse_fuse_fd(). With this patch, the code that depends on fuse_mnt_parse_fuse_fd() returning an fd != -1 is optimized out.

--- lib/mount_util.h.orig       2018-11-06 18:57:47 UTC
+++ lib/mount_util.h
@@ -15,4 +15,9 @@ int fuse_mnt_umount(const char *progname, const char *
                    const char *rel_mnt, int lazy);
 char *fuse_mnt_resolve_path(const char *progname, const char *orig);
 int fuse_mnt_check_fuseblk(void);
+
+#if defined(__FreeBSD__)
+#define fuse_mnt_parse_fuse_fd(x) (-1)
+#else
 int fuse_mnt_parse_fuse_fd(const char *mountpoint);
+#endif

novel:
> I'm not opening a new issue since I think the following applies to this one ... (let me know if you prefer a seperate issue for tracking purposes).

This should be fixed as well by this PR I believe, please take a look at this commit: https://github.com/libfuse/libfuse/pull/321/commits/a9177e03f8016bdbdd471eadb2784ee99601fff2

stesser:
Yes, I have verified that commit a9177e0 builds on FreeBSD without any patches.

novel:
I haven't test the fd's  thing, yet to figure out how to test it properly.
I did test regular sshfs mounts though and that works for me now with these patches (that's what I was trying to fix in the first place, noticed build issues later when started looking into it).

Nikratio:
Thanks!

PR #330, opened 2018-12-01
lib/fuse_lowlevel.c: fix memory leak

Mashimiao:
Signed-off-by: Ma Shimiao <shimiao.ma@gmail.com>


Nikratio:
Thanks for the patch! Could you briefly explain why you think there is a leak? Note that your patch results in crashing because of an invalid free() call, while without the patch the tests are passing.

Mashimiao:
Oh, sorry...   that may be my fault.   I find we just do malloc int do_poll, but not free ph in that function.
I rescan the test code, and find it will be freed in other places.  will close this pr.

PR #332, opened 2018-12-06
Fixed paddings exceeds the boundary

browei:
Fixed paddings exceeds the boundary which cost one more unsigned int

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -65,7 +65,7 @@ struct fuse_file_info {
 	unsigned int flock_release : 1;
 
 	/** Padding.  Do not use*/
-	unsigned int padding : 27;
+	unsigned int padding : 26;
 
 	/** File handle.  May be filled in by filesystem in open().
 	    Available in all other file operations */


Nikratio:
This would break  compatibility with binaries compiled against libfuse headers before this change, wouldn't it?

jpandre:
This is unfortunate. The more as the next field is an uint64, which on most 64-bit cpus is aligned to 8 bytes, so a couple of ints are lost. It is even possible that uint64's are aligned differently in 32-bit mode, and as a consequence 32-bit user-space file systems would not run on a 64-bit kernel (check needed).

Nikratio:
Yeah, I agree. Do you see any solution though?

jpandre:
Actually, you are lucky. Formerly there were an int and a bit field, which took two ints. The int was then redefined as a bit field, creating the overflow, and also taking two ints. This is why the error was not noticed earlier (though the used bits fields were relocated).

I suggest you fix the size of the unused field, and add another dummy int in order to show a clean layout (to remain compatible with old binaries, this extra int should be inserted before the fields, but you may prefer to be compatible with recent binaries and insert the extra int after the fields).

Nikratio:
Thanks for looking into this! Would you mind updating the pull request?

As far as I can see all changes in this area of the file are 8+ years old - is that what you mean with "recent binaries"?

jpandre:
Sorry for the delay. In the meantime, I discovered that the header I use (Centos7, updated last month) is not the same as the one on the repository. In particular the old field "fh_old", which had been kept for years to maintain binary compatibility, has been removed, supposedly with no damage.

This means the binary compatibility is not needed any more, and the proposed patch should do.

Nikratio:
Sorry, I still do not understand. `struct fuse_file_info` was last changed 8+ years ago. Any changes made here that change the layout will thus break compatibility with any binary that was compiled between 8 years ago and today. Why do you think that "binary compatibility is not needed any more"?

PR #337, opened 2018-12-20
fix memory leak in print_module_help method

devkral:
The members of the fuse_args struct are never freed, so there is a small memory leak.


Nikratio:
Thanks!

PR #339, opened 2019-01-07
fuse_lowlevel_ops documentation updates

### LINKED ISSUE #333: Documentation updates
I'd like to submit some updates to libfuse's doxygen documentation, but don't have time right now. I'll briefly describe them here in case someone gets to it first.

`fuse_lowlevel_ops::lookup` should mention that it increases the inode's reference count and a corresponding `::forget` will arrive for that inode.

`fuse_lowlevel_ops::open` should mention that returning ENOSYS from open() will prevent release() calls from arriving as well.

`fuse_lowlevel_ops::release` says "For every open call there will be exactly one release call." which is not true under unmount.

`fuse_lowlevel_ops::releasedir` says "For every opendir call there will be exactly one releasedir call." which is not true under unmount.

`fuse_lowlevel_ops::readdir` says "The function does not have to report the '.' and '..' entries, but is allowed to do so." which is true, but this behavior is observable to userspace.

`fuse_lowlevel_ops::opendir` says "Filesystem may also implement stateless directory I/O and not store anything in fi->fh, though that makes it impossible to implement standard conforming directory stream operations in case the contents of the directory can change between opendir and releasedir." but that's not true - the 63 usable bits of off_t passed between readdir calls can be treated as seekdir/readdir cookies and index to the appropriate location in the directory stream, even under concurrent modification. For example, see https://github.com/facebookexperimental/eden/blob/ad044f2da05f2fad54515f950747e336c6add50f/eden/fs/inodes/TreeInode.cpp#L1700

chadaustin:
Add a bit of clarity to the fuse_lowlevel_ops documentation, as it's one of the best high-level descriptions of the FUSE kernel protocol.

Fixes #333.

### DIFF
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -496,8 +496,8 @@ struct fuse_lowlevel_ops {
 	 * If this request is answered with an error code of ENOSYS
 	 * and FUSE_CAP_NO_OPEN_SUPPORT is set in
 	 * `fuse_conn_info.capable`, this is treated as success and
-	 * future calls to open will also succeed without being send
-	 * to the filesystem process.
+	 * future calls to open and release will also succeed without being
+	 * sent to the filesystem process.
 	 *
 	 * Valid replies:
 	 *   fuse_reply_open
@@ -613,6 +613,10 @@ struct fuse_lowlevel_ops {
 	 *
 	 * For every open call there will be exactly one release call.
 	 *
+	 * However, in the case of a forced unmount, the filesystem should
+	 * safely handle the possibility of not receiving remaining release
+	 * calls.
+	 *
 	 * The filesystem may reply with an error, but error values are
 	 * not returned to close() or munmap() which triggered the
 	 * release.
@@ -660,11 +664,12 @@ struct fuse_lowlevel_ops {
 	 * etc) in fi->fh, and use this in other all other directory
 	 * stream operations (readdir, releasedir, fsyncdir).
 	 *
-	 * Filesystem may also implement stateless directory I/O and not
-	 * store anything in fi->fh, though that makes it impossible to
-	 * implement standard conforming directory stream operations in
-	 * case the contents of the directory can change between opendir
-	 * and releasedir.
+	 * Implementing standard conforming directory stream operations when
+	 * the directory is modified between opendir and releasedir calls
+	 * requires either storing state in a handle indexed by fi->fh (and
+	 * using readdir's off_t parameter to index into it) or using
+	 * readdir's off_t parameter as a cookie that will seek to the correct
+	 * entry in the directory even under concurrent modification.
 	 *
 	 * Valid replies:
 	 *   fuse_reply_open
@@ -691,7 +696,9 @@ struct fuse_lowlevel_ops {
 	 * its lookup count.
 	 *
 	 * The function does not have to report the '.' and '..'
-	 * entries, but is allowed to do so.
+	 * entries, but is allowed to do so. Note that, if readdir does
+	 * not return '.' or '..', they will not be implicitly returned,
+	 * and this behavior is observable by the caller.
 	 *
 	 * Valid replies:
 	 *   fuse_reply_buf
@@ -712,6 +719,10 @@ struct fuse_lowlevel_ops {
 	 *
 	 * For every opendir call there will be exactly one releasedir
 	 * call.
+	 *
+	 * However, in the case of a forced unmount, the filesystem should
+	 * safely handle the possibility of not receiving remaining releasedir
+	 * calls.
 	 *
 	 * fi->fh will contain the value set by the opendir method, or
 	 * will be undefined if the opendir method didn't set any value.
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -611,11 +611,8 @@ struct fuse_lowlevel_ops {
 	 * file: all file descriptors are closed and all memory mappings
 	 * are unmapped.
 	 *
-	 * For every open call there will be exactly one release call.
-	 *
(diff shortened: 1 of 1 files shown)


include/fuse_lowlevel.h
```diff
@@ -611,7 +611,9 @@ struct fuse_lowlevel_ops {
 	 * file: all file descriptors are closed and all memory mappings
 	 * are unmapped.
 	 *
-	 * For every open call there will be exactly one release call.
+	 * For every open call there will be exactly one release call,
+	 * excepting unmount, when all open file handles are implicitly
+	 * released.
```
Nikratio: As discussed in the bug, let's research this more carefully. I believe even on unmount you should receive release() calls. If not, this is a bug that should be fixed in the code rather than documented.

include/fuse_lowlevel.h
```diff
@@ -660,11 +662,11 @@ struct fuse_lowlevel_ops {
...
-	 * Filesystem may also implement stateless directory I/O and not
-	 * store anything in fi->fh, though that makes it impossible to
-	 * implement standard conforming directory stream operations in
-	 * case the contents of the directory can change between opendir
-	 * and releasedir.
+	 * Implementing standard conforming directory stream operations when
+	 * the directory is modified between opendir and releasedir calls
+	 * requires either storing state in a handle indexed by fi->fh or using
+	 * readdir's off_t parameter as a cookie that will seek to the correct
+	 * entry in the directory even under concurrent modification.
```
Nikratio: This sounds like you could implement standard conforming stream operations by using just `fi->fh`, but without using `off_t`. Is that what you mean? If so, I how would you do it? I don't see a way...

include/fuse_lowlevel.h
```diff
@@ -711,7 +715,8 @@ struct fuse_lowlevel_ops {
 	 * Release an open directory
 	 *
 	 * For every opendir call there will be exactly one releasedir
-	 * call.
+	 * call, excepting unmount, when all open directory handles are
```
Nikratio: Same as above for `release()`

include/fuse_lowlevel.h
chadaustin: That is what I meant! I'll clarify. :)

The simplest solution to POSIX-compliant readdir is to store the entire directory list in an array indexed by `fi->fh` and then using off_t as an integer offset into that array, but I just want to point out that it's possible to implement readdir correctly, even under concurrent modification, without storing state in `fi->fh` by ensuring that the off_t cookie always seeks to the correct place in the stream even under modification.

chadaustin:
More code review clarifications.

include/fuse_lowlevel.h
```diff
@@ -613,6 +613,10 @@ struct fuse_lowlevel_ops {
 	 *
 	 * For every open call there will be exactly one release call.
```
Nikratio: Let's just add a "(unless the filesystem is force-unmounted)" here.

include/fuse_lowlevel.h
```diff
@@ -613,6 +613,10 @@ struct fuse_lowlevel_ops {
...
 	 *
+	 * However, in the case of a forced unmount, the filesystem should
+	 * safely handle the possibility of not receiving remaining release
+	 * calls.
```
Nikratio: ..and omit the separate paragraph, because I don't think force unmount needs to be safely handled. Force implies unsafe, and handling this correctly may impose a big penalty in regular usage.

include/fuse_lowlevel.h
Nikratio: I'm sorry, I appreciate your efforts, but I don't see why your version is better than the status quo (for example, you cannot generally "index" into a state). Could you remind me what the drawbacks of the current text are?

include/fuse_lowlevel.h
```diff
@@ -713,6 +720,10 @@ struct fuse_lowlevel_ops {
 	 * For every opendir call there will be exactly one releasedir
 	 * call.
 	 *
+	 * However, in the case of a forced unmount, the filesystem should
```
Nikratio: same as above

include/fuse_lowlevel.h
chadaustin: Thanks for the feedback! I'm also not happy with the wording I chose. I'd welcome any suggestions.

The current text states that using `fi->fh` is required to implement POSIX-conforming readdir. This is not true - if off_t is treated as a magic 63-bit cookie that will always seek into the correct part of the directory stream, `fi->fh` is not needed. In fact, that's how filesystems like btrfs and xfs work. (And it's a good thing `fi->fh` is not required - should the FUSE_NO_OPENDIR_SUPPORT patch land, filesystems could opt out of the opportunity to even set it.)

Without being too verbose, I'd like to describe that it's possible to implement POSIX-conforming readdir by using each dirent's offset parameter as a cookie that indexes into a separate b-tree.

The easiest way to get POSIX conformance is to use `fi->fh` as a pointer to an array of N elements and have off_t merely be the index into that array. But it's certainly not required. It took me a while to figure this out myself, so I just wanted to save someone else the work of determining whether `fi->fh` is required for correctness.

Nikratio: I see. How about removing this paragraph completely? The preceding paragraph already says "may", so it should be clear that this is not a must.

Instead, we could flesh out the description of readdir() and add something like: "If entries are added or removed during a readdir cycle, they may or may not be returned. However, they must not cause other entries to be skipped or returned more than once."

..and leave it to the user to figure out how the various ways to achieve that.

Nikratio: (probably need a better phrase than "readdir cycle", something along the lines of "readdir getting an off_t value that it had returned previously but file were added/removed afterwards")

PR #344, opened 2019-01-22
fuse_lowlevel: validate path components

stefanha:
Several FUSE requests contain single path components.  A correct FUSE
client sends well-formed path components but there is currently no input
validation in fuse_lowlevel.c in case something went wrong.

File system implementations might accidentally treat invalid inputs as
absolute/relative paths instead of rejecting them.  Save implementations
from duplicating input validation by handling it in fuse_lowlevel.c.

Signed-off-by: Stefan Hajnoczi <stefanha@redhat.com>

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -1031,9 +1031,23 @@ int fuse_reply_poll(fuse_req_t req, unsigned revents)
 	return send_reply_ok(req, &arg, sizeof(arg));
 }
 
+/* Is `path` a single path component that is not "." or ".."? */
+static int is_safe_path_component(const char *path)
+{
+	if (strchr(path, '/'))
+		return 0;
+
+	return !is_dot_or_dotdot(path);
+}
+
 static void do_lookup(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 {
 	char *name = (char *) inarg;
+
+	if (!is_safe_path_component(name)) {
+		fuse_reply_err(req, EINVAL);
+		return;
+	}
 
 	if (req->se->op.lookup)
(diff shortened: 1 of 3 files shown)


Nikratio:
With "FUSE client", do you mean the kernel? Because that is what your code seems to be checking, but I am not convinced that this is worth it. Or do you mean a userspace client accessing the filesystem? In that case the kernel will do the checks already...

lib/fuse_lowlevel.c
```diff
@@ -1031,10 +1031,24 @@ int fuse_reply_poll(fuse_req_t req, unsigned revents)
...
+
+	return !is_dot_or_dotdot(path);
+}
+
 static void do_lookup(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 {
 	char *name = (char *) inarg;
 
+	if (!is_safe_path_component(name)) {
+		fuse_reply_err(req, EINVAL);
```
szmi: Actually lookup may receive "." or ".." from kernel when NFS export is enabled.  NFS exporting of fuse is not very good, and doesn't support generic file handles.   So longer term this interface probably needs to be replaced, but for now returning an error would be a regression.  This validation needs to be done in filesystems.

This only applies to LOOKUP, other ops should never get "." or ".." in file name.

stefanha:
Perhaps it's simplest to move this input validation into the virtio-fs-specific file system implementation instead of generic libfuse.  My hope was that cheap sanity checks would be useful because they protect all file system implementations and anyone porting FUSE to other platforms or new usecases (like virtio-fs!) will benefit.

stefanha:
Having thought about it more I've closed this pull request and we'll do the sanity checking in the file system instead.

PR #346, opened 2019-01-24
context node userdata

kivikakk:
👋 Greetings! I wanted to wave an idea by, and I figure the best way to do so is with code.

This PR shows a feature addition which I'm using in some currently-private code; namely, the ability to add a piece of userdata to a `struct node` when using the high-level API. There are some awkward edges in the newly introduced API which I'm not happy with, but it's possible I might have missed something obvious.

The motivation to store userdata against `struct node` is for fast, cache-like access. In my use-case, I'm writing a stacking filesystem which stores information in extended attributes on the lower fs. However, making a call to `lgetxattr(2)` every time I need to check the metadata would quickly become a bottleneck. Using this API, I cache the attributes against the `struct node` which the request is executed against. The cache stays attached to the correct node thanks to the high-level API's management of nodes.

One awkward edge is the need for the `parent` argument. In my use-case, we most often need to check/set the xattrs of the directory containing the target node, rather than the node itself, in order to make decisions about how to handle the request. To facilitate this, we store the `fuse_ino_t` and any applicable path name component in the `struct fuse_context_i` when preparing the request. This requires developers to understand how the low-level API works and why, say, `readlink` doesn't have the parent directory accessible in this API, but `mknod` does. This could be fixed with some overhead.

The other awkward edge is when there's more than one "context node" applicable for a given request. This is seen clearly in `rename` and `copy_file_range`. This doesn't affect my work as we only need to access the node userdata in a select number of operations, but does affect the generality of the API.

Do you have any thoughts about this kind of API or how to make it cleaner? Do you think something similar to this might be accepted into FUSE, or is it totally outside your design vision? I decided to try this path first to see if there's something general that might be useful to more than just me, but if we can't make it work, I might just need to copy the high-level API `fuse.c` into my own project with modifications.

Thanks for your time!

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -1016,9 +1016,11 @@ struct fuse_context *fuse_get_context(void);
  * Get the userdata stored against the node the current filesystem operation is
  * targetting.
  *
+ * @param parent get userdata for the parent of the node instead of the node
+ *               itself
  * @return the userdata
  */
-void *fuse_get_context_node_userdata(void);
+void *fuse_get_context_node_userdata(int parent);
 
 /**
  * Set the userdata for the node the current filesystem operation is
@@ -1029,12 +1031,14 @@ void *fuse_get_context_node_userdata(void);
  * If an error occurs, a nonzero value is returned. In this case, you should
  * finalize (or otherwise clean up) the userdata immediately.
  *
+ * @param parent set userdata for the parent of the node instead of the node
+ *               itself
  * @param data the userdata to store against the node
  * @param finalize a function to be called when the node is deallocated, or
  *                 when the userdata is changed; can be NULL
  * @return 0 on success, nonzero on failure
  */
-int fuse_set_context_node_userdata(void *data, void (*finalize)(void *));
+int fuse_set_context_node_userdata(int parent, void *data, void (*finalize)(void *));
 
 /**
  * Get the current supplementary group IDs for the current request
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -1012,45 +1012,51 @@ int fuse_loop_mt(struct fuse *f, struct fuse_loop_config *config);
  */
 struct fuse_context *fuse_get_context(void);
 
+/** Userdata to be stored against each node.
+ */
+struct fuse_node_userdata {
+	/** Opaque pointer to be used by the caller of
+	 * fuse_context_node_userdata_* methods. */
+	void *data;
+
+	/** Finalizer called by FUSE when the node is being freed.  Users
+	 * should set this if special cleanup for the data field is necessary.
+	 * */
+	void (*finalize)(void *);
+};
+
 /**
- * Get the userdata stored against the node the current filesystem operation is
- * targetting.
+ * Acquire the userdata stored against the node the current filesystem
+ * operation is targetting.  Use fuse_context_node_userdata_release to release
+ * the lock held on the userdata.
  *
  * @param parent get userdata for the parent of the node instead of the node
  *               itself
- * @return the userdata
+ * @return the fuse_node_userdata struct, or NULL if an error occurred (errno
+ *         is non-zero), or NULL if the parent of the root node was requested
+ *         (errno is zero)
  */
-void *fuse_get_context_node_userdata(int parent);
+struct fuse_node_userdata *fuse_context_node_userdata_acquire(int parent);
 
 /**
- * Get the mutex for the userdata for the node the current filesystem operation
- * is targetting.
- *
- * Note that the mutex is not used internally -- it is up to the calling code
- * to use the mutex to synchronize operations on the node's userdata.
- *
- * @param parent get mutex for the parent of the node instead of the node itself
- * @return a pointer to the node's userdata mutex
+ * Like fuse_context_node_userdata_acquire, except a timeout on acquiring the
+ * lock is given, as per pthread_mutex_timedlock.  NULL may be returned with
+ * errno set to ETIMEDOUT if the userdata lock could not be acquired within the
+ * timeout.
  */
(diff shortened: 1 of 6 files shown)


trapexit:
I might be missing something but wouldn't just passing the nodeid through the API be a better design? Allowing the user to treat it as a key into whatever data store they wish thereby keeping the API simpler? Or does the existing high level API not provide callbacks sufficient for the management of the life of the objects?

Then again you could just use the fusepath as a key though not as nice.

kivikakk:
> I might be missing something but wouldn't just passing the nodeid through the API be a better design? 

For my purposes at least, both the nodeid and the parent nodeid are required. One caveat: the nodeid for the target itself often _doesn't exist_ at the point of time of calling into the fs.

For instance, take `mknod`. `fuse_lib_mknod` calls `fuse_fs_mknod`, and only if it succeeds does it then call `lookup_path`, causing the creation of the underlying `struct node`. The API in this PR is designed to be called once you know the target exists (i.e. you wouldn't call `fuse_set_context_node_userdata` until you knew for sure you were returning `0`). If we passed the nodeid into the fs's op function, it implies we already created the `struct node`, even if the fs op might fail and not do anything.

> Or does the existing high level API not provide callbacks sufficient for the management of the life of the objects?

Indeed, I don't think there is (again, unless I've missed something obvious). But we could add a callback in `free_node` (or wherever) to let that happen. It does mean the user code would be duplicating the node hash table, but it would be a smaller change to FUSE.

> Then again you could just use the fusepath as a key though not as nice.

Which path exactly do you mean by "fusepath"? A big benefit of storing this data against the node is that the reference is maintained in the face of renames -- of the node itself, or of any of its parents.

kivikakk:
In 7d0336a I've changed the API from:

```c
void *fuse_get_context_node_userdata(int parent);
pthread_mutex_t *fuse_get_context_node_userdata_lock(int parent);
int fuse_set_context_node_userdata(int parent, void *data, void (*finalize)(void *));
```

To:

```c
struct fuse_node_userdata {
	void *data;
	void (*finalize)(void *);
};

struct fuse_node_userdata *fuse_context_node_userdata_acquire(int parent);
struct fuse_node_userdata *fuse_context_node_userdata_timedacquire(
		int parent, const struct timespec *restrict abs_timeout);
int fuse_context_node_userdata_release(int parent);
```

This has simplified code in both the FUSE and client side.

Nikratio:
Interesting idea. Not yet sure what I think about it.

One thing that would definitely have to change in order for this to go in is that we need to clearly document what a "node" is, when it is created, destroyed, whether it's mutable, etc. It's bad enough that the term is used all over the code internally, but we can no longer get away with that once it's part of the public API. Clearly writing this down may also expose potential issues with this idea, so I think this would be a good next step.

kivikakk:
> Clearly writing this down may also expose potential issues with this idea, so I think this would be a good next step.

I think this is an excellent next step. The properties of a node and how it can be expected to behave are currently implicit in the operations the various library functions do to them (seen most clearly in `rename`) -- making it explicit should be helpful.

> (I have this hunch that if you need something like a node, you be actually be better of using the low-level API)

I had initially started on using the low-level API, but 95% of what we need is in the high-level API, and I didn't really want to copy `fuse.c` into my codebase, instead seeing about implementing what I needed as a patch to FUSE instead. I was inspired by [the following mailing list post](https://sourceforge.net/p/fuse/mailman/message/31925484/) from @mrvn:

> The fuse kernel interface works on inodes and the low level interface reflects this. The high level interface works on paths and has to translate between inodes and path. That takes memory and cpu to do.
>
> BUT, what would be the point of using the low level interface if you then need to translate those inode numbers to paths yourself? Do you think you can translate them faster using less resources? Less bugs?
> 
> IMHO the low level interface is best when implementing an actual filesystem based on inodes where you don't do any mapping from inode to path.
>
> For all the pass through filesystems that translate fuse operations to operations of an underlying filesystem (e.g. unionfs-fuse) the highlevel interface is far better suited.

As what I'm working on is a passthrough filesystem at its core, the argument is persuasive.

Nikratio:
> > For all the pass through filesystems that translate fuse operations to operations of an underlying filesystem (e.g. unionfs-fuse) the highlevel interface is far better suited.
> 
> As what I'm working on is a passthrough filesystem at its core, the argument is persuasive.

Actually, making a truly standard-conforming passthrough filesystem is rather hard (potentially even impossible) with the high-level API. The combination correctly handling hard links and removal of files while they are still open is very difficult to encapsulate in a path-based API. You may want to take a look at the example passthrough_ll filesystem.

kivikakk:
Indeed, we started out with using the low-level API, but with a bunch of TODOs for things like path (re)construction, cache eviction, proper synchronisation, etc. The high-level API provides all these things battle-tested and, at thousands of LoC, isn't something I'm keen to reimplement.

Our filesystem's use case (Linux implementation of [VFS for Git](https://vfsforgit.com/)) doesn't need to handle hard links -- `git` doesn't support them at all. In short, we don't need to be 100% standard conforming, since we only need to support the operations that make sense for a git working tree.

As for removing files while they're still open, that is indeed tricky, and the work the high-level API does here (with node hiding) is helpful, and (I believe) sufficient for our uses. Am I missing something about the downsides of how the high-level API works here? (put another way: what would a better implementation look like?)

trapexit:
The node hiding behavior in the high level API could be more flexible. Having a dedicated "hide" callback would give more control to the user. Some of the misc issues with the high level API IMO is the result of trying to hide too much from the user.

kivikakk:
We've decided for now to go in a different direction that doesn't necessitate this kind of metadata (or reification of `struct node`), so I'm closing the PR.

PR #356, opened 2019-02-11
Silence -Wformat-truncation=/-Wformat-overflow= warnings

kusumi:
sprintf(3)/snprintf(3) destination buffers need to be large enough
so that gcc doesn't warn -Wformat-truncation= or -Wformat-overflow=
when source buffer size is 1024 bytes.

```
../test/test_syscalls.c:1445:47: warning: '%s' directive output may be truncated writing 1 byte into a region of size between 0 and 1023 [-Wformat-truncation=]
 #define PATH(p)  (snprintf(path, sizeof path, "%s/%s", testdir, p), path)
                                               ^~~~~~~
../test/test_syscalls.c:1458:19:
  res = mkdir(PATH("a"), 0755);
                   ~~~
```

### DIFF
--- a/test/test_syscalls.c
+++ b/test/test_syscalls.c
@@ -19,13 +19,13 @@ static char testfile[1024];
 static char testfile2[1024];
 static char testdir[1024];
 static char testdir2[1024];
-static char subfile[1024];
+static char subfile[1280];
 
 static char testfile_r[1024];
 static char testfile2_r[1024];
 static char testdir_r[1024];
 static char testdir2_r[1024];
-static char subfile_r[1024];
+static char subfile_r[1280];
 
 static char testname[256];
 static char testdata[] = "abcdefghijklmnopqrstuvwxyz";
@@ -491,7 +491,7 @@ static int cleanup_dir(const char *path, const char **dir_files, int quiet)
 
 	for (i = 0; dir_files[i]; i++) {
 		int res;
-		char fpath[1024];
+		char fpath[1280];
 		sprintf(fpath, "%s/%s", path, dir_files[i]);
 		res = unlink(fpath);
 		if (res == -1 && !quiet) {
@@ -524,7 +524,7 @@ static int create_dir(const char *path, const char **dir_files)
 		return -1;
 
 	for (i = 0; dir_files[i]; i++) {
-		char fpath[1024];
+		char fpath[1280];
 		sprintf(fpath, "%s/%s", path, dir_files[i]);
 		res = create_file(fpath, "", 0);
 		if (res == -1) {
@@ -1445,7 +1445,7 @@ static int test_rename_dir_loop(void)
 #define PATH(p)		(snprintf(path, sizeof path, "%s/%s", testdir, p), path)
 #define PATH2(p)	(snprintf(path2, sizeof path2, "%s/%s", testdir, p), path2)
 
-	char path[1024], path2[1024];
+	char path[1280], path2[1280];
 	int err = 0;
 	int res;
 


Nikratio:
Thanks! Where does the magic number of 1280 come from? The error message sounds more like an off-by-one error...

kusumi:
It doesn't have to be 1280, but I just aligned with 256 since the original size was 1024.
It can be anything as long as it's larger than potential end result of "%s/%s" that the compiler can find on compile time.

Basically, the first "%s" is from 1024 bytes buffer if you look at callers, hence whatever comes for "/%s" could potentially exceed the existing 1024 bytes destination buffer size, and that's what compiler is warning. Most of the C strings for the second "%s" aren't that long, so 1280 is more than enough, but 1024 or 1025 is not.

The warning I pasted in the commit message is just one example. There are huge number of these from PATH()/PATH2() macros. All the PATH()/PATH2() calls are generating this warning, plus a few more from other places which I fixed also.

Nikratio:
Ah, I see. I think in that case the proper solution is to use PATH_MAX rather than an arbitrary value. What do you think?

kusumi:
OK, I'll re-PR with PATH_MAX.

kusumi:
Actually, PATH_MAX value is platform specific, so using it as a size relative to hard coded buffer size (1024) doesn't always work.
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

e.g. PATH_MAX is 4096 on Linux, but not on *BSD.
```
[root@ tmp]# uname -r
11.2-RELEASE
[root@ tmp]# cat ./test1.c
#include <stdio.h>
#include <limits.h>

int main(void) {
        printf("%d\n", PATH_MAX);
        return 0;
}
[root@ tmp]# clang -Wall -g ./test1.c
[root@ tmp]# ./a.out
1024
```

Nikratio:
I'm sorry, I don't follow. Yes, PATH_MAX is platform specific. Why is that a problem? It will still be the maximum length of the constructed path. I'm not sure what you mean with "relative to hard coded buffer size".

kusumi:
If I change my 1280 to PATH_MAX, it equals changing 1280 to 1024 on *BSD, hence nothing changes on *BSD or possibly some other platforms. gcc still complains in the same way that it did before on those platforms....

The point is that the source buffer for the first "%s" is 1024.

gcc is complaining on possible truncation (snprintf case) or overflow (sprint case) when we store the source string plus some C string literals (in "%s/%s" format), given the destination buffer is also 1024.

Nikratio:
Having a destination buffer larger than PATH_MAX doesn't make sense, since using it would give errors later on. If that gives a compiler warning, then the warning is spurious and we should find a way to suppress it *without* changing the semantics of the code. Does GCC have something like pragmas?

kusumi:
It's not about runtime correctness on filesystem semantics.

It's just to *silence* compiler. This is something that recent versions of gcc complains. gcc is warning on whatever it can *statically* find without knowing filesystem semantics.

If you don't want to change it, I'm okay with that too. I'm just sending it as there are more than 30 of these.

Nikratio:
I think we are in agreement that the compiler should be silenced. However, doing this by replacing 1024 with some higher value at the same time also changes the *runtime semantics*, and not in the right direction. If you can find a way to preserve the runtime semantics but silence the warning, that's fine with me.

kusumi:
If we want to use PATH_MAX (which is fine of course), the source buffer size in this code was supposed to be something based on PATH_MAX.

The real problem is not what the diff is doing, but rather the fact that the source buffer already had 1024. My diff and gcc warnings are both based on that.

kusumi:
`However, doing this by replacing 1024 with some higher value at the same time also changes the _runtime semantics_`

Not really. Not changing destination buffer size (1024) to higher value *is* changing runtime semnantics, because that can cause potential truncation for snprintf(3), and overflow for sprintf(3), which is exactly what the compiler is warning...

Well, not a big deal as we most likely never have that long path on in this case. Feel free to close it.

Nikratio:
Well, I guess one magic number isn't better than the other, and if it does silence the warnings...

PR #359, opened 2019-02-13
Fixed fuse_free_buf() to check each fuse_buf array element instead of…

hselin:
Current fuse_free_buf() code always checks buf->buf[0].flags to free buf->buf[i].mem instead of buf->buf[i].flags.


Nikratio:
Thanks!

PR #372, opened 2019-02-28
Link against libiconv when possible

HazelFZ:
When modules/iconv.c is built , linking needs -liconv. otherwise . In some system, <iconv.h> can be found ,but libiconv can not. Other ,there is failure:

FAILED: lib/libfuse3.so.3.4.1 
cc  -o lib/libfuse3.so.3.4.1 'lib/lib@@fuse3@sha/fuse.c.o' 'lib/lib@@fuse3@sha/fuse_loop.c.o' 'lib/lib@@fuse3@sha/fuse_loop_mt.c.o' 'lib/lib@@fuse3@sha/fuse_lowlevel.c.o' 'lib/lib@@fuse3@sha/fuse_opt.c.o' 'lib/lib@@fuse3@sha/fuse_signals.c.o' 'lib/lib@@fuse3@sha/buffer.c.o' 'lib/lib@@fuse3@sha/cuse_lowlevel.c.o' 'lib/lib@@fuse3@sha/helper.c.o' 'lib/lib@@fuse3@sha/modules_subdir.c.o' 'lib/lib@@fuse3@sha/mount_util.c.o' 'lib/lib@@fuse3@sha/mount.c.o' 'lib/lib@@fuse3@sha/modules_iconv.c.o' -Wl,--no-undefined -Wl,--as-needed -shared -fPIC -Wl,--start-group -Wl,-soname,libfuse3.so.3 -Wl,--version-script,/root/fusermaste/libfuse-master/lib/fuse_versionscript -ldl -lrt -Wl,--end-group -pthread  
lib/lib@@fuse3@sha/modules_iconv.c.o: In function `iconv_new':
/root/fusermaste/libfuse-master/build/../lib/modules/iconv.c:679: undefined reference to `libiconv_open'
/root/fusermaste/libfuse-master/build/../lib/modules/iconv.c:685: undefined reference to `libiconv_open'
/root/fusermaste/libfuse-master/build/../lib/modules/iconv.c:704: undefined reference to `libiconv_close'
/root/fusermaste/libfuse-master/build/../lib/modules/iconv.c:706: undefined reference to `libiconv_close'
lib/lib@@fuse3@sha/modules_iconv.c.o: In function `iconv_convpath':
/root/fusermaste/libfuse-master/build/../lib/modules/iconv.c:68: undefined reference to `libiconv'
/root/fusermaste/libfuse-master/build/../lib/modules/iconv.c:96: undefined reference to `libiconv'
lib/lib@@fuse3@sha/modules_iconv.c.o: In function `iconv_destroy':
/root/fusermaste/libfuse-master/build/../lib/modules/iconv.c:571: undefined reference to `libiconv_close'
/root/fusermaste/libfuse-master/build/../lib/modules/iconv.c:572: undefined reference to `libiconv_close'
[42/63] Compiling C object 'example/example@@poll@exe/poll.c.o'.
ninja: build stopped: subcommand failed.


lib/meson.build
```diff
@@ -10,7 +10,9 @@ else
 endif
 
 if cfg.get('HAVE_ICONV')
+   libiconv = cc.find_library('iconv', required: true)
    libfuse_sources += [ 'modules/iconv.c' ]
+   deps += [ libiconv ]
```
Nikratio: This won't do anything, since it's overriden below

meson.build
```diff
@@ -43,8 +43,10 @@ foreach func : test_funcs
 endforeach
 cfg.set('HAVE_SETXATTR', 
         cc.has_function('setxattr', prefix: '#include <sys/xattr.h>'))
+        
+libiconv = cc.find_library('iconv', required: false)
 cfg.set('HAVE_ICONV', 
-        cc.has_function('iconv', prefix: '#include <iconv.h>'))
+        cc.has_function('iconv', prefix: '#include <iconv.h>') and libiconv.found())
```
Nikratio: This shouldn't be necessary. If libiconv doesn't exist, has_function should fail anyway. On the other hand, systems that don't require an external iconv library will stop working.

lib/meson.build
```diff
@@ -10,7 +10,9 @@ else
 endif
 
 if cfg.get('HAVE_ICONV')
+   libiconv = cc.find_library('iconv', required: true)
```
Nikratio: Not all systems require an external library, so this shouldn't be required

lib/meson.build
HazelFZ: I am sorry,it is my fault,In my work,i had move the define of deps up, but I forget commit it.

meson.build
HazelFZ: I think adding it will be compatible more system. In my system , libiconv.so.2 only exists in /usr/local/lib/, but not /lib64/ . So ,  
has_function do success but libiconv can be found.

HazelFZ:
I have update the meson.build and lib/meson.build. I think the reason of building failure is : 
At first ,gcc searches the iconv.h ,and the order of search path is : 

gcc version 4.8.5 20150623 (Red Hat 4.8.5-4) (GCC) 
COLLECT_GCC_OPTIONS='-v' '-E' '-mtune=generic' '-march=x86-64'
 /usr/libexec/gcc/x86_64-redhat-linux/4.8.5/cc1 -E -quiet -v /dev/null -mtune=generic -march=x86-64
ignoring nonexistent directory "/usr/lib/gcc/x86_64-redhat-linux/4.8.5/include-fixed"
ignoring nonexistent directory "/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../x86_64-redhat-linux/include"
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/gcc/x86_64-redhat-linux/4.8.5/include
 /usr/local/include
 /usr/include

So, if /usr/local/include/iconv.h exists, it will be searched first. And then is /usr/include/iconv.h.
However, the iconv.h in libc exists in /usr/include/, the iconv.h in libiconv in /usr/local/include/
So the libiconv's header will be searched first , if exists. And in libiconv's header ,there is a define: 

 69 #ifndef LIBICONV_PLUG
 70 #define iconv_open libiconv_open
 71 #endif

But , in libc ,the libiconv_open is not defined. So , I think if libiconv exists, it should be linked..

Nikratio:
That sounds right. If libiconf exists, it should be linked. Could you update your pull request to implement that?

Nikratio:
(Currently you are linking against libiconv if HAVE_ICONV is defined and the library exists, but you are not attempting to link when setting HAVE_ICONV - unless that test only compiles without linking?)

HazelFZ:
I have update the pull request.You can merge it. I think complies without linking can do well. Because,if HAVE_ICONV is not set, libiconv should not be linked.  ,If HAVE_ICONV is set ,only if libiconv exist, we could link against it.

PR #374, opened 2019-03-06
Fix the changelog entry for protocol version 7.12

asomers:


Nikratio:
Thanks! For the commit message, could you provide a reference for this? E.g. a link to the CL that added the flag? Otherwise it's hard to confirm that the new version is correct...

asomers:
Looks like the build failure is unrelated.

PR #378, opened 2019-03-08
Improve documentation for the flush method

### LINKED ISSUE #373: Under FreeBSD, flush() return value is ignored
The documentation for the `flush` method says that it can be used "if a filesystem wants to return write errors in close()" while noting that "many applications ignore close(2) errors".  However, the problem is bigger than "many applications ignore close(2) errors", because many kernels don't even deliver them.  

POSIX says that "The close() operation itself need not block awaiting such I/O completion.", so a FUSE filesystem can't assume that an application will ever see an error returned by flush.

http://pubs.opengroup.org/onlinepubs/9699919799/functions/close.html

Indeed, FreeBSD's close implementation does not wait, so it has no opportunity to return flush errors to the application.

Linux's close(2) implementation does appear to return EIO errors (I haven't checked the source), but still the man page says "A  successful  close does not guarantee that the data has been successfully saved to disk", which means that an application programmer can never rely on the absence of errors from close to mean anything.

Given this limitation, a FUSE filesystem can never rely on errors returned by flush to even be delivered.   So I think the documentation should be revised to suggest that the return value will be ignored.  However, in that case there wouldn't seem to be any remaining purpose to flush?  Perhaps the method should be deprecated?

cc @cemeyer

asomers:
Fixes #373

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -475,20 +475,22 @@ struct fuse_operations {
 	 * release which is called on the close of the last file descriptor for
 	 * a file.  One use of flush is to return errors from delayed writes.
 	 * However, such use is non-portable, because POSIX does not require
-	 * [close] to wait for delayed I/O to complete.  Also, many
-	 * applications ignore close() errors.  For these reason, filesystems
-	 * should not assume that errors returned by flush will ever be noticed
-	 * or even delivered.
+	 * [close] to wait for delayed I/O to complete.  Linux's close() does,
+	 * but other operating systems' do not.  Also, many applications ignore
+	 * close() errors.  For these reason, filesystems should not assume
+	 * that errors returned by flush will ever be noticed or even
+	 * delivered.
 	 *
 	 * NOTE: The flush() method may be called more than once for each
 	 * open().  This happens if more than one file descriptor refers to an
-	 * opened file due to dup(), dup2() or fork() calls.  It is not
+	 * opened file, e.g. due to dup(), dup2() or fork() calls.  It is not
 	 * possible to determine if a flush is final, so each flush should be
 	 * treated equally.  Multiple write-flush sequences are relatively
 	 * rare, so this shouldn't be a problem.
 	 *
-	 * Filesystems shouldn't assume that flush will always be called after
-	 * some writes, or that if will be called at all.
+	 * Filesystems shouldn't assume that flush will be called at any
+	 * particular point.  It may be called more times than expected, or not
+	 * at all.
 	 *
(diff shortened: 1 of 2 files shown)


include/fuse.h
```diff
@@ -471,21 +471,26 @@ struct fuse_operations {
...
-	 * open().	This happens if more than one file descriptor refers
-	 * to an opened file due to dup(), dup2() or fork() calls.	It is
-	 * not possible to determine if a flush is final, so each flush
-	 * should be treated equally.  Multiple write-flush sequences are
-	 * relatively rare, so this shouldn't be a problem.
-	 *
-	 * Filesystems shouldn't assume that flush will always be called
-	 * after some writes, or that if will be called at all.
+	 * open().  This happens if more than one file descriptor refers to an
+	 * opened file due to dup(), dup2() or fork() calls.  It is not
```
Nikratio: let's add an "e.g due to dup().." here - there are other ways.

include/fuse.h
```diff
@@ -471,21 +471,26 @@ struct fuse_operations {
...
+	 * possible to determine if a flush is final, so each flush should be
+	 * treated equally.  Multiple write-flush sequences are relatively
+	 * rare, so this shouldn't be a problem.
+	 *
+	 * Filesystems shouldn't assume that flush will always be called after
+	 * some writes, or that if will be called at all.
```
Nikratio: "if"?

Nikratio: "always be called after some" doesn't make sense to me. What about simple saying "shouldn't depend on flush() being called at any particular points when data has been written - it may be called multiple times or not at all"?

include/fuse.h
```diff
@@ -471,21 +471,26 @@ struct fuse_operations {
...
 	 *
-	 * Flush is called on each close() of a file descriptor.  So if a
-	 * filesystem wants to return write errors in close() and the file
-	 * has cached dirty data, this is a good place to write back data
-	 * and return any errors.  Since many applications ignore close()
-	 * errors this is not always useful.
+	 * Flush is called on each close() of a file descriptor, as opposed to
+	 * release which is called on the close of the last file descriptor for
+	 * a file.  One use of flush is to return errors from delayed writes.
+	 * However, such use is non-portable, because POSIX does not require
```
Nikratio: I don't like the term "non-portable" - it sounds scary without giving any useful information. Instead, let's specifically say something like "Under Linux, errors returned by flush() will be passed to userspace as errors from close(), so flush() is a good place to write back any cached dirty data. However, many applications ignore errors on close(), and on non-Linux systems, close() may suceed even if flush() returns an error. For these reasons, [...]"

PR #381, opened 2019-03-10
Defined the (*ioctl)() commands as unsigned int (v2)

jpandre:
On Linux the ioctl() commands are requested as "unsigned long", but fuse
truncates them to 32 bits. Transmitting the commands to user space 
file systems as "unsigned int" is a workaround for processing ioctl
commands which do not fit into a signed int.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,7 @@
+* Changed ioctl commands to "unsigned int" in order to support commands
+  which do not fit into a signed int. Commands issued by applications
+  are still truncated to 32 bits.
+
 libfuse 3.4.2 (2019-03-09)
 ==========================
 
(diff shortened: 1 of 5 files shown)


Nikratio:
Thanks! Could you please also describe the change in Changes.rst, so that people who are using ioctl() know what to do?

trapexit:
And squash the commits?

trapexit:
You don't need to close PRs. Just overwrite your branch and the PR will get updated.

jpandre:
I do not know why it got closed...
How should I overwrite my branch ? It is already pushed.

trapexit wrote:

trapexit:
You already did by force pushing the branch. That's what push -f does.

PR #384, opened 2019-03-13
RFC: max_pages support implementation

scosu:
Hi,

implementation for libfuse of the 4.20 max_pages feature. In this implementation I am coupling it to the max_write parameter. By default the max_pages will be at the maximum on kernels >= 4.20 and at 32 for < 4.20.

Best,
Markus

### DIFF
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -131,3 +131,9 @@ struct fuse *fuse_new_31(struct fuse_args *args, const struct fuse_operations *o
 int fuse_loop_mt_32(struct fuse *f, struct fuse_loop_config *config);
 int fuse_session_loop_mt_32(struct fuse_session *se, struct fuse_loop_config *config);
 
+#define FUSE_MAX_MAX_PAGES 256
+#define FUSE_DEFAULT_MAX_PAGES_PER_REQ 32
+
+/* room needed in buffer to accommodate header */
+#define FUSE_BUFFER_HEADER_SIZE 0x1000
+
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -1909,6 +1909,14 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 			se->conn.capable |= FUSE_CAP_HANDLE_KILLPRIV;
 		if (arg->flags & FUSE_NO_OPENDIR_SUPPORT)
 			se->conn.capable |= FUSE_CAP_NO_OPENDIR_SUPPORT;
+		if (!(arg->flags & FUSE_MAX_PAGES)) {
+			size_t max_bufsize =
+				FUSE_DEFAULT_MAX_PAGES_PER_REQ * getpagesize()
+				+ FUSE_BUFFER_HEADER_SIZE;
+			if (bufsize > max_bufsize) {
+				bufsize = max_bufsize;
+			}
+		}
 	} else {
 		se->conn.max_readahead = 0;
 	}
@@ -1955,10 +1963,10 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 			bufsize);
 		bufsize = FUSE_MIN_READ_BUFFER;
 	}
+	se->bufsize = bufsize;
 
-	bufsize -= 4096;
-	if (bufsize < se->conn.max_write)
-		se->conn.max_write = bufsize;
+	if (se->conn.max_write > bufsize - FUSE_BUFFER_HEADER_SIZE)
+		se->conn.max_write = bufsize - FUSE_BUFFER_HEADER_SIZE;
 
 	se->got_init = 1;
 	if (se->op.init)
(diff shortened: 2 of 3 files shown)


Nikratio:
Apologies for the delay, and thanks for the patch! Could you please:
- rebase on current master?
- confirm that fuse_kernel.h matches the kernel headers exactly (or you changes will get lost on the next sync)
- add a ChangeLog entry

scosu:
sorry for the delay, but I finally updated this patch.
I removed the header synchronization patch as it didn't contain anything significant and added a ChangeLog entry.

lib/fuse_i.h
```diff
@@ -9,6 +9,13 @@
 #include "fuse.h"
 #include "fuse_lowlevel.h"
 
+/**
+ * Fsync flags
+ *
+ * FUSE_FSYNC_FDATASYNC: Sync data only, not metadata
+ */
+#define FUSE_FSYNC_FDATASYNC	(1 << 0)
```
Nikratio: Is this change intentional? It doesn't seem to belong to the feature being added.

lib/fuse_lowlevel.c
```diff
@@ -1835,6 +1835,12 @@ static void do_copy_file_range(fuse_req_t req, fuse_ino_t nodeid_in, const void
 		fuse_reply_err(req, ENOSYS);
 }
 
+#define FUSE_MAX_MAX_PAGES 256
+#define FUSE_DEFAULT_MAX_PAGES_PER_REQ 32
+
+/* room needed in buffer to accommodate header */
+#define HEADER_SIZE 0x1000
+
```
Nikratio: Please put the #ifdefs into `lib/fuse_i.h`, and use a more expressive name for `HEADER_SIZE`

lib/fuse_lowlevel.c
```diff
@@ -1909,6 +1915,14 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 			se->conn.capable |= FUSE_CAP_HANDLE_KILLPRIV;
 		if (arg->flags & FUSE_NO_OPENDIR_SUPPORT)
 			se->conn.capable |= FUSE_CAP_NO_OPENDIR_SUPPORT;
+		if (!arg->flags & FUSE_MAX_PAGES) {
```
Nikratio: Please add parens to make precedence clear

lib/fuse_lowlevel.c
```diff
@@ -1909,6 +1915,14 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
...
+			size_t max_bufsize =
+				FUSE_DEFAULT_MAX_PAGES_PER_REQ * getpagesize()
+				+ HEADER_SIZE;
+			if (bufsize > max_bufsize) {
```
Nikratio: Why do we have to limit the bufsize only if FUSE_MAX_PAGES is not set? I'd think that we always need *some* limit.

lib/fuse_lowlevel.c
```diff
@@ -1955,10 +1969,10 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 			bufsize);
 		bufsize = FUSE_MIN_READ_BUFFER;
 	}
+	se->bufsize = bufsize;
```
Nikratio: If `bufsize` is only modified in the new `if` statement above, then it would make more sense to move the assignment there as well, IMO.

lib/fuse_lowlevel.c
```diff
@@ -1985,6 +1999,14 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 		return;
 	}
 
+	if (se->conn.max_write < bufsize - HEADER_SIZE) {
```
Nikratio: This check is intended to cover the case where init() has *increased* `max_write`, right? If so, let's make that explicit by storing `old_max_write` and comparing with it after init() has run.

lib/fuse_lowlevel.c
```diff
@@ -1955,10 +1969,10 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 			bufsize);
 		bufsize = FUSE_MIN_READ_BUFFER;
 	}
+	se->bufsize = bufsize;
 
-	bufsize -= 4096;
-	if (bufsize < se->conn.max_write)
-		se->conn.max_write = bufsize;
+	if (bufsize - HEADER_SIZE < se->conn.max_write)
```
Nikratio: Let's invert this do `se->conn.max_write > [...]`, since we're imposing a limit on max_write here.

lib/fuse_lowlevel.c
scosu: I only moved this up in this file to use it at this point. But I moved it to fuse_i.h now anyway.

lib/fuse_lowlevel.c
scosu: In fuse_session_new() the bufsize is initialized to the largest possible value. Only if we detect that FUSE_MAX_PAGES is not set we decrease it to the size the kernel uses. I thought it would be better to have a sane default in se->bufsize at all times.

lib/fuse_lowlevel.c
scosu: No, bufsize is also modified two lines above this statement, in case it is lower than FUSE_MIN_READ_BUFFER.

lib/fuse_lowlevel.c
scosu: max_write is at the maximum already. Just before calling init() max_write is adjusted to be the maximum that fits into the buffer. This check is in case it is not using the buffer completely. In that case we can safely remove the total size of the buffer.

PR #391, opened 2019-04-02
Add documentation for opting out of opendir and releasedir

chadaustin:
Add documentation for kernel 5.1's opendir/releasedir opt-out behavior, which also implies readdir caching.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -8,6 +8,7 @@ libfuse 3.5.0 (UNRELEASED)
   on mountpoints within SMB 2.0 filesystems).
 * Added a new `cache_readdir` flag to `fuse_file_info` to enable
   caching of readdir results. Supported by kernels 4.20 and newer.  
+* Add support and documentation for FUSE_CAP_NO_OPENDIR_SUPPORT.
 
 libfuse 3.4.2 (2019-03-09)
 ==========================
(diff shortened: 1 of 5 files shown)


chadaustin:
The relevant kernel commits are:

https://github.com/torvalds/linux/commit/d9a9ea94f748f47b1d75c6c5e33edcf74476c445
https://github.com/torvalds/linux/commit/fabf7e0262d0bd57739d29aeac94c44b0542ff1f

Nikratio:
Thanks, this is great! Could you please also add a definition of `FUSE_CAP_NO_OPENDIR_SUPPORT` to the header files and add an entry to ChangeLog.rst?

chadaustin:
I think I touched all the right places. :)

Note that the kernel now has a handful of flags that aren't currently documented in libfuse: https://github.com/torvalds/linux/blob/master/include/uapi/linux/fuse.h

Specifically FUSE_ABORT_ERROR, FUSE_MAX_PAGES, and, perhaps most interestingly, FUSE_CACHE_SYMLINKS.

chadaustin:
Ah, I see the build failure. Should I add an ifdef to libfuse? Is it expected people will recompile libfuse when upgrading kernels?

Nikratio:
No, you need to sync up include/fuse_kernel.h with the kernel headers (cf. 4ae58ff). In this case, re-basing to current master may already fix the issue.

Nikratio:
Thanks!

PR #392, opened 2019-04-03
Whitelist smb2

lemenkov:
Whitelist SMB2 found on some NAS devices. We have CIFS whitelisted already, so let's allow next-gen CIFS as well.

Also I've sorted out allowed filesystems alphabetically.


Nikratio:
Thanks! Could you please add an entry to `ChangeLog.rst`?

lemenkov:
> Thanks! Could you please add an entry to `ChangeLog.rst`?

Done.

Issue #389, opened 2019-03-30
Advise file system developers not to use file handles in nonportable ways


asomers:
# Background
The FUSE protocol allows a server to assign a different filehandle each and every time that a file is opened.  Multiple filehandles can be concurrently valid for the same file.  Operations like `FUSE_READ` and `FUSE_WRITE` are supposed to include the file handle associated with the particular file descriptor that initiated the operation.  However, that isn't possible in all cases.  The `FUSE_WRITE_CACHE` bit indicates whether the file handle is valid for a `FUSE_WRITE` operation, and the `FATTR_FH` flag indicates whether it's valid for a `FUSE_SETATTR` operation.

# Problem
There are other cases, too, when the file handle can be invalid.  In fact, the entire concept of a file handle is really a Linuxism.  Other operating systems enforce a distinction between the file layer and the vnode layer.  That's why OSX, FreeBSD, OpenBSD, and Illumos can almost *never* guarantee that they're sending the right file handle for any operation.  NetBSD takes a more extreme position; it doesn't allow multiple concurrent fuse file handles for a single file (which leads to other problems).   But since operations other than `FUSE_WRITE` and `FUSE_SETATTR` don't have anything analogous to the `FUSE_WRITE_CACHE` or `FATTR_FH` bits, there's no way for those operating systems to indicate that the file handle is guessed.

# Mitigating Circumstances
The impact of this bug is small simply because few if any fuse servers care about the accuracy of fuse file handles.  I surveyed 42 real-world FUSE filesystems* and _none_ of them ever check the `writepage` bit.  Only _one_ of them (catfs) checks `FATTR_FH`, and it can deal with the absence of that flag.  I can't find any evidence that any FUSE filesystem actually wants to assign distinct file handles on each concurrent `FUSE_OPEN` and requires the FUSE client to keep them straight.

# Solution
Ideally the FUSE protocol would require servers to assign the same file handle for every concurrent open of a single file.  Unfortunately, that's not the current situation and such a change would be backwards-incompatible.  The second best solution would be to update the kernel protocol to include some equivalent of the `FUSE_WRITE_CACHE` bit for every operation that includes a file handle.  However, that would require all clients and servers alike to update their kernel API version, so uptake would be extremely slow.  The path of least resistance would simply be to document the problem.  I recommend the following:

* Advise FUSE server developers to assign the same file handle for every concurrent open of a single file.
[...]

PR #393, opened 2019-04-04
More documentation improvements

asomers:
Clarify the relationship between file handles and file descriptors, add references to O_EXEC and O_SEARCH, and do some proofreading.


jpandre:
Replacing "Thread ID of the calling process" by "Process ID of the calling process" hides the fact that the pid identifies a thread in a multi-thread process. I would rather suggest "Process ID of the calling thread".

asomers:
"Process ID of the calling thread" sounds good.

include/fuse_common.h
```diff
@@ -32,14 +32,22 @@ extern "C" {
...
 
 /**
- * Information about open files
+ * Information about an open file handle
+ *
+ * File Handles are created by the open, opendir, and create methods and closed
+ * by the release and releasedir methods.  Multiple file handles may be
+ * concurrently open for the same file.  Generally, a client will create one
+ * file handle per file descriptor, though in some cases multiple file
+ * descriptors can share a single file handle.
```
Nikratio: Can we omit the last sentence please? I think this is confusing - people will be uncertain what "client" means in this context, and the file handle isn't actually created by the client but may be created by the filesystem.

include/fuse_common.h
```diff
@@ -32,14 +32,22 @@ extern "C" {
 #endif
 
 /**
- * Information about open files
+ * Information about an open file handle
```
Nikratio: Let's keep "open files" here. Filesystems aren't obliged to put anything into fi->fh, in which case there is no file handle (yet there is a fuse_file_info structure).

include/fuse_common.h
```diff
@@ -32,14 +32,22 @@ extern "C" {
...
 struct fuse_file_info {
 	/** Open flags.	 Available in open() and release() */
 	int flags;
 
 	/** In case of a write operation indicates if this was caused by a
-	    writepage */
+	    delayed write from the page cache.  If so, then the context's pid,
+	    uid, and gid fields will not be valid, and the file handle may not
+	    correspond to the file descriptor responsible for the write. */
 	unsigned int writepage : 1;
```
Nikratio: That's much better, thanks! I'd still like to omit the "and the file handle may not correspond to the file descriptor responsible for the write", because FUSE doesn't know anything about file descriptors.

include/fuse_common.h
```diff
@@ -67,8 +75,16 @@ struct fuse_file_info {
...
-	    Available in all other file operations */
+	/** File handle id.  May be filled in by filesystem in create,
+	 * open, and opendir().  Available in most other file operations on the
+	 * same file handle.
+	 *
+	 * Different file handles for the same file may have different file
+	 * handle IDs.  However, since an operation's file handle is not
+	 * guaranteed to be associated with the file descriptor ultimately
+	 * responsible for the operation, it is advised that file systems not
+	 * associated any credential-related information with the file handle.
```
Nikratio: I don't like the second paragraph. The first sentence sound like a tautology, and the second sentence seems to imply that it would a good idea to put credentials into the file handle if e.g. direct_io is used. Can we just omit this paragraph?

include/fuse_common.h
asomers: How about "client will open" instead of "will create"?

include/fuse_common.h
asomers: We seem to be using the word "file handle" to refer to two different things.  There's the 64-bit fh field, of course.  But there's also the thing that you create during `open`.  For every `open` there will be exactly one `release`, which implies that both client and server must be tracking whatever `open` created (even if the server always sets the same value for `fh`).  If that's not a file handle, then what is it?

include/fuse_common.h
asomers: What about that sentence suggests that it would be a good idea to put credentials into the file handle?  I'm trying to suggest the opposite.

include/fuse_common.h
asomers: fuse_kernel.h says "file handle is guessed", but I was trying to come up with something more precise.  I think it's important to be precise here, especially since the relationship between file handles and file descriptors is so close.  It's like the difference between process IDs and thread IDs; on Linux they're almost, but not quite, identical.  So it's very important to carefully document the distinctions or people _will_ get them confused.

include/fuse_common.h
Nikratio: That probably depends on which kernel you use libfuse with. I would say on the libfuse side there is no need to worry about that.

include/fuse_common.h
Nikratio: I think that's still more confusing than helpful... Either people already know how the kernel maps file descriptors to FUSE file handles (in which case they don't need this information), or they don't know (in which case they won't understand it from this sentence either).

include/fuse_common.h
Nikratio: I agree with the goal, but I don't think that one or two extra sentences in the comments are helping with this. I coherent, multi paragraph explanation somewhere else would be a different story...

include/fuse_common.h
Nikratio: You're saying it's not a good idea because an operation's file handle is not guaranteed to be associated with the file descriptor ultimately responsible for the operation. That is true, but it is neither the only reason nor the most important one.

include/fuse_common.h
asomers: Where then?

include/fuse_common.h
asomers: Well, "most important" is a subjective judgement.  What do you think is the most important reason?  I could add it.

include/fuse_common.h
asomers: I think that most libfuse users would prefer their file systems to be portable, so they need to know what the least common denominiator client will do.  From the users' perspective, they need clear rules on they're allowed to do with libfuse.  Is there something I should add here to make it clearer?

include/fuse_common.h
asomers: Kernel aside, there are three distinct things in libfuse that currently share two names:
1) The thing identified by a node_id, called a "file"
2) The thing identified by a node_id and a file handle, created by `open`, `opendir`, or `create`, and described by `fuse_file_info`.  What should it be called?  You've previously said this should be called a "file handle" (https://github.com/libfuse/libfuse/issues/389#issuecomment-478732346).  Are you having second thoughts?
3) The 64-bit value stored in `fuse_file_info.fh` variously called "file handle" or "file handle id".  The latter makes more sense to me if thing 2 will be called a "file handle".

Nikratio:
I'm short on discussion time at the moment, so I've just merged the non-controversial parts for now. Thanks!

Nikratio:
Would you mind rebasing on current master if you have the time?

include/fuse_common.h
Nikratio: I don't want to call any one reason out at all, because this will just tempt people into neglecting the other reasons.

include/fuse_common.h
Nikratio: Maybe create a new doc/writeback.rst file for that? That would be a good place to also store other information related to that.

include/fuse_common.h
Nikratio: Hmm. Yeah, this is messy. Here's my take on it (from general principles without looking at which terms are actually used right now):

(1) should not be a called a file (since it may be a directory). I would refer to it as a "directory entry". 

I would not say that (2) is a thing at all. I would say that `fuse_file_info` provides information about either an "open file" or an "open directory". However, most filesystems will create some internal structure to keep track of open files or directories that libfuse knows nothing about - that is the "file handle".

(3) is the "file handle id"

Nikratio:
Yes, should be fixed now. Sorry about that. Let me know when you want me to take another look!

Nikratio:
Friendly ping. Do you think further changes are needed here? If so, would you mind rebasing to current master?

PR #398, opened 2019-04-17
Add fallocate and use it instead of posix_fallocate if possible

liubogithub:
fuse.ko has supported FALLOC_FL_KEEP_SIZE and FALLOC_FL_PUNCH_HOLE at this
moment and more modes may be supported in the future.

fallocate(2) supports modes while posix_fallocate(2) does not, so this
makes lo_fallocate use fallocate(2) instead.

Signed-off-by: Liu Bo <bo.liu@linux.alibaba.com>


Nikratio:
Thanks!

PR #402, opened 2019-04-30
Enable passthrough_ll on FreeBSD

asomers:


example/passthrough_ll.c
```diff
@@ -336,6 +377,13 @@ static int lo_do_lookup(fuse_req_t req, fuse_ino_t parent, const char *name,
 		inode->fd = newfd;
 		inode->ino = e->attr.st_ino;
 		inode->dev = e->attr.st_dev;
+#ifndef HAVE_AT_EMPTY_PATH
+		struct lo_inode *parent_inode;
+
+		parent_inode = lo_inode(req, parent);
+		snprintf(inode->pathname, sizeof(inode->pathname), "%s/%s",
```
Nikratio: This will get out of sync if the parent directory is renamed, even if the rename comes through the mountpoint.

Could we store only the directory entry name here, and reconstruct the full path on demand?

Nikratio:
Is this PR still up-to-date, or are you planning to rewrite as described in issue #407 ?

asomers:
I'm planning to do #407 first and then rebase this PR.

Nikratio:
Friendly ping. Any update on this? Or should I just close this PR for now?

asomers:
Sorry; I got things mostly working but eventually concluded that it's too hard and perhaps impossible to get true passthrough semantics without `AT_EMPTY_PATH`.  A better example for the low-level API would be to store each FUSE-visible file as two flat files on the underlying file system (one for the inode itself and one for the data), and store FUSE-visible directories in flat files, too.

Nikratio:
No worries, I appreciate your efforts!

PR #413, opened 2019-05-09
passthrough: fix unix-domain sockets on FreeBSD

asomers:
FreeBSD doesn't allow creating sockets using mknod(2).  Instead, you
must use socket(2) and bind(2).  Add appropriate logic to
examples/passthrough.cc.  Also, add a test in test_syscalls.c.  In order
to make the test work, I had to shorten the mountpoint that pytest
generates by abbreviating some stuff in test_examples.py.

### DIFF
--- a/example/passthrough.c
+++ b/example/passthrough.c
@@ -53,6 +53,8 @@
 #include <sys/xattr.h>
 #endif
 
+#include "passthrough_helpers.h"
+
 static void *xmp_init(struct fuse_conn_info *conn,
 		      struct fuse_config *cfg)
 {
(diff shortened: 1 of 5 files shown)


example/passthrough.c
```diff
@@ -144,9 +148,34 @@ static int xmp_mknod(const char *path, mode_t mode, dev_t rdev)
 		res = open(path, O_CREAT | O_EXCL | O_WRONLY, mode);
 		if (res >= 0)
 			res = close(res);
-	} else if (S_ISFIFO(mode))
+	} else if (S_ISFIFO(mode)) {
 		res = mkfifo(path, mode);
-	else
+#ifdef __FreeBSD__
```
Nikratio: Let's put all this code into a separate function, so that the callback remains easy to understand. I'm thinking something like `mknod_wrapper`, which for Linux just calls mknod and for FreeBSD does the extra stuff.

example/passthrough_ll.c
```diff
@@ -391,9 +391,35 @@ static void lo_mknod_symlink(fuse_req_t req, fuse_ino_t parent,
 
 	if (S_ISDIR(mode))
 		res = mkdirat(dir->fd, name, mode);
-	else if (S_ISLNK(mode))
+	else if (S_ISLNK(mode)) {
 		res = symlinkat(link, dir->fd, name);
-	else
+#ifdef __FreeBSD__
```
Nikratio: Same here.. maybe even factor the wrapper out into a bsd_util.c file?

test/test_examples.py
```diff
@@ -81,10 +81,15 @@ def test_hello(tmpdir, name, options, cmdline_builder):
...
-def test_passthrough(tmpdir, name, debug, capfd, writeback):
+@pytest.mark.parametrize("writeback", ("F", "T"))
+@pytest.mark.parametrize("name", ('', 'f', 'l'))
+@pytest.mark.parametrize("debug", ("F", "T"))
+def test_pt(tmpdir, name, debug, capfd, writeback):
+    writeback = {"F": False, "T": True}[writeback]
+    debug = {"F": False, "T": True}[debug]
+    name = {"": "passthrough",
+            "f": "passthrough_fh",
+            "l": "passthrough_ll"}[name]
```
Nikratio: Ugh, that looks terrible. Can we just make the path shorter? I assume the problem is that pytest includes all the parametrisation in the path. I think I'd be fine with defining our own `short_tmdir` fixture that uses `tempfile`.

Nikratio:
Thanks, looks good now - except that it broke some of the unit tests (see CI results) - could you take another look? :-)

asomers:
I don't think those tests are related to my changes.  Are you sure that it isn't just an intermittent test?

Nikratio:
Quite sure, I've never seen them before and I've just merged a pull request that passed: https://travis-ci.org/libfuse/libfuse

asomers:
Ok, it took me way too long to find the problem, but it should be fixed now.  BTW, have you considered rewriting the tests?  They'd be much more maintainable with proper testcase isolation.  I've recently done something very similar using Googletest.  Unfortunately, I don't have time to rewrite all of libfuse's tests myself.

https://svnweb.freebsd.org/base/projects/fuse2/tests/sys/fs/fusefs/

Nikratio:
You mean `test_syscalls.c`? Yeah, that's high on my "list of things that should be done" list, but also very low on my "list of things that I enjoy doing".

PR #421, opened 2019-05-24
Skip install parts that require root when not running as root

DrDaveD:
When running as non-root, for example when building a package, skip the parts of the install process that require root.

This takes the place of a patch that the Fedora build of fuse3 was applying.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -5,7 +5,11 @@ option('udevrulesdir', type : 'string', value : '',
        description: 'Where to install udev rules (if empty, query pkg-config(1))')
 
 option('utils', type : 'boolean', value : true,
-       description: 'Wheter or not to build and install helper programs')
+       description: 'Whether or not to build and install helper programs')
 
 option('examples', type : 'boolean', value : true,
-       description: 'Wheter or not to build example programs')
+       description: 'Whether or not to build example programs')
+
+option('no-root', type : 'boolean', value : false,
+       description: 'Install files without root permissions')
+
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -10,6 +10,6 @@ option('utils', type : 'boolean', value : true,
 option('examples', type : 'boolean', value : true,
        description: 'Whether or not to build example programs')
 
-option('no-root', type : 'boolean', value : false,
-       description: 'Install files without root permissions')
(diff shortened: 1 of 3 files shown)


Nikratio:
Thanks for the patch! I think fixing this issue is a good idea, but this isn't the right way to do it because it means that if you accidentally run as non-root, you get an incomplete install without any warning or error messages.

If you could rework the patch to instead introduce a new build option (e.g. `nonroot`, similar to `utils` or `disable-mtab`), I'd be happy to merge.

util/install_helper.sh
```diff
@@ -22,16 +22,17 @@ else
     DESTDIR="${DESTDIR%/}"
 fi
 
-chown root:root "${DESTDIR}${bindir}/fusermount3"
-chmod u+s "${DESTDIR}${bindir}/fusermount3"
-
 install -D -m 644 "${MESON_SOURCE_ROOT}/util/fuse.conf" \
 	"${DESTDIR}${sysconfdir}/fuse.conf"
 
+if [ `id -u` = 0 ]; then
```
Nikratio: Please replace with a Meson build option as discussed above.

Nikratio:
Are you interested in reworking this as described above? If not, hope it's ok if I close the PR.

DrDaveD:
Sorry, yes I am.  I made an attempt last week that didn't work but I plan to get back to it.

DrDaveD:
Ok this version seems to work, how does it look to you?

meson_options.txt
```diff
@@ -5,7 +5,11 @@ option('udevrulesdir', type : 'string', value : '',
...
-       description: 'Wheter or not to build and install helper programs')
+       description: 'Whether or not to build and install helper programs')
 
 option('examples', type : 'boolean', value : true,
-       description: 'Wheter or not to build example programs')
\ No newline at end of file
+       description: 'Whether or not to build example programs')
+
+option('no-root', type : 'boolean', value : false,
+       description: 'Install files without root permissions')
```
Nikratio: Let's rename this to `useroot` (so you don't need an extra `useroot` variable in `meson.build`), defaut it to `True`, and describe it as *Set owner and setuid bits on installed files*.

PR #425, opened 2019-06-04
Fix build with IGNORE_MTAB

michaelforney:
`IGNORE_MTAB` doesn't seem to be a documented option, but it is referenced in `lib/mount.c`, `lib/mount_util.c`, and `util/fusermount.c`, though only defined in `lib/mount_util.c`.

However, I've been building with `-DIGNORE_MTAB` since my system has `/etc/mtab` as a symlink to `/proc/self/mounts`, and I don't want fuse to try to modify it.

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -208,6 +208,7 @@ static int may_unmount(const char *mnt, int quiet)
 
 	return 0;
 }
+#endif
 
 /*
  * Check whether the file specified in "fusermount3 -u" is really a
@@ -395,6 +396,7 @@ static int chdir_to_parent(char *copy, const char **lastp)
 	return 0;
 }
 
+#ifndef IGNORE_MTAB
 /* Check whether the kernel supports UMOUNT_NOFOLLOW flag */
 static int umount_nofollow_support(void)
 {


Nikratio:
It looks as if you've just moved some code around - is that intentional? If there are changes to the logic, could you separate them out?

michaelforney:
Yes, this just moves the definitions of `chdir_to_parent` and `check_is_mount` (along with dependent functions `clone_newns` and `check_is_mount_child`) outside the `#ifndef IGNORE_MTAB` block, since they are used by `should_auto_unmount`.

There are no logical changes.

An equivalent fix is

```diff
diff --git a/util/fusermount.c b/util/fusermount.c
index 5e0b104..56da6ec 100644
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -208,6 +208,7 @@ static int may_unmount(const char *mnt, int quiet)
 
 	return 0;
 }
+#endif
 
 /*
  * Check whether the file specified in "fusermount3 -u" is really a
@@ -395,6 +396,7 @@ static int chdir_to_parent(char *copy, const char **lastp)
 	return 0;
 }
 
+#ifndef IGNORE_MTAB
 /* Check whether the kernel supports UMOUNT_NOFOLLOW flag */
 static int umount_nofollow_support(void)
 {
```

but I thought it would be better to group the mtab-specific functions in a single preprocessing group.

Nikratio:
Ah, got it. Could you please push the equivalent patch for the benefit of future readers and usability of `git blame`?

michaelforney:
Updated to use the ifdef fix instead.

PR #434, opened 2019-07-10
Add support for stream-like files

navytux:
Starting from 2008 it became possible to implement FUSE filesystems
with pipe/socket-like files. For this a filesystem server should set
file_info->nonseekable and the kernel will correspondingly use
FMODE_NONSEEKABLE for opened file handle. See commit cafdcb253e4c "If
open sets fi->nonseekable, libfuse will tell the kernel that the file is
not seekable" which added nonseekable support to libfuse.

Unfortunately in 2014 a kernel regression was introduced(*): all file IO
switched to be done under lock (file->f_pos_lock) with the idea that
file position changes should be atomic. However for stream-like
files that were using nonseekable open it created possibility of
deadlock in between read and write:

	- file read could be waiting for data from filesystem server
	- filesystem server is generating data in reply to client write to that file
	- write to that file is waiting for f_pos_lock taken by read
	-> deadlock

Unfortunately now the regression is not possible to fix on just kernel
side because a) it is already 5 years after 2014, and b) there are
existing FUSE filesystems that use nonseekable open _and_ position for
corresponding file(+). In other words because of backward compatibility
today we cannot change the kernel to avoid taking f_pos_lock and not to
pass file position for filesystem server for nonseekable FUSE file
handles.

The solution to the deadlock problem, that is already merged to Linux
kernel, is to introduce new FUSE open flag (FOPEN_STREAM) that is
explicitly marking file handle as not having position at all, and thus
for such file handle the kernel will avoid taking f_pos_lock.

	git.kernel.org/linus/10dce8af3422
	git.kernel.org/linus/bbd84f33652f

When implementing stream-like files filesystem servers should be using
both FOPEN_NONSEEKABLE and FOPEN_STREAM:

	- on kernels released before 2014, since in-kernel FUSE client
	  ignores unknown FOPEN flags, this will continue to work ok as before.
	- on kernels with FOPEN_STREAM support, this will avoid
	  read/write deadlock.
	- on kernels released after 2014 but without FOPEN_STREAM
	  support, the deadlock will be there but using FOPEN_STREAM
	  does not make the situation worse.

	- FOPEN_STREAM was merged into Linux 5.2 (see commits ^^^) and
	  was backported to all long-term stable Linux kernel versions
	  that are taking their roots after kernel regression point:

	  Linux 5.1  git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=v5.1.4-110-g5bb2a758f2ed
	  Linux 4.19 git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=v4.19.45-82-gfea685000caf
	  Linux 4.14 git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=v4.14.124-35-g585724f83b0c
	  Linux 4.9  git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=v4.9.180-83-gcfd8d2e79524
	  Linux 4.4  git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=v4.4.180-241-gc9696a8f3e64
	  Linux 3.18 https://android.googlesource.com/kernel/common/+/9225b2f9d042
	  Linux 3.16 git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=v3.16.69-123-g6d654f9d5d1f

Note: we cannot change libfuse to use use both FOPEN_NONSEEKABLE and
FOPEN_STREAM on just file->info_nonseekable=1 for the same backward
compatibility reason why we cannot change the kernel to avoid using file
position for nonseekable opens (see "+" again). Thus let's introduce
new flag file_info->stream and let filesystems that implement
stream-like files set both file_info->stream and
file_info->nonseekable_open.

Even though example/poll.c implements only read, I've changed that
example to use file_info->stream=1 too, because file implemented there
does not use file position at all and because that example is kind of
canonical on how to implement "pipe-like" files with libfuse.

Since this patch is trying to fix regression introduced in 2014, it
should be applied to all libfuse branches covering time after that
moment:

	- master (current 3.x development)
	- fuse_3_0_bugfix
	- fuse-2_9_bugfix
	- fuse_2_8_bugfix

/cc @Nikratio, @szmi

(*) git.kernel.org/linus/9c225f2655e3
(+) e.g. gvfs, see

	https://codesearch.debian.net/search?q=-%3Enonseekable+%3D
	https://gitlab.gnome.org/GNOME/gvfs/blob/1.40.0-6-gcbc54396/client/gvfsfusedaemon.c#L1080
	https://gitlab.gnome.org/GNOME/gvfs/blob/1.40.0-6-gcbc54396/client/gvfsfusedaemon.c#L1247-1346
	https://gitlab.gnome.org/GNOME/gvfs/blob/1.40.0-6-gcbc54396/client/gvfsfusedaemon.c#L1399-1481


navytux:
amended the patch with the following interdiff:

```diff
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -71,10 +71,6 @@ struct fuse_file_info {
            seekable. */
        unsigned int nonseekable : 1;
 
-       /** Can be filled in by open, to indicate that the file is
-           stream-like (no file position at all). */
-       unsigned int stream : 1;
-
        /* Indicates that flock locks for this file should be
           released.  If set, lock_owner shall contain a valid value.
           May only be set in ->release(). */
@@ -86,8 +82,12 @@ struct fuse_file_info {
            nothing when set by open()). */
        unsigned int cache_readdir : 1;
 
+       /** Can be filled in by open, to indicate that the file is
+           stream-like (no file position at all). */
+       unsigned int stream : 1;
+
        /** Padding.  Reserved for future use*/
-       unsigned int padding : 26;
+       unsigned int padding : 25;
 
        /** File handle id.  May be filled in by filesystem in create,
         * open, and opendir().  Available in most other file operations on the
```

appologize for missing that padding and binary compatibility initially.

navytux:
( CI: FreeBSD failure seems to be unrelated to the patch - it [says](https://cirrus-ci.com/task/5672556701417472) it did not start the build job at all: "Failed to start: Read timed out" )

navytux:
Ping, @Nikratio .

include/fuse_common.h
```diff
@@ -82,8 +82,12 @@ struct fuse_file_info {
 	    nothing when set by open()). */
 	unsigned int cache_readdir : 1;
 
+	/** Can be filled in by open, to indicate that the file is
+	    stream-like (no file position at all). */
```
Nikratio: Could you state if this must be specified together with `nonseekable`?

Also, is there a way to *briefly* explain when one should enable this and when one should go with just `nonseekable`? (Maybe with one example of a file of each kind)

Nikratio:
Thanks for the patch! I'm somewhat confused by this. From the description, it seems that at this point in time any reasonable filesystem would set both nonseekable and stream_like. Is that correct? If so, let's state this clearly in the comments (which are used to generate documentation).

Then, I understand that we cannot implicitly set stream_like, because eg gvfs depends on it not being set. Is that correct? However, as I understand this will lead to deadlocks, thus making such filesystems unusable anyway - so why do we have to care? Is the likelihood of the deadlock very small?

Finally, what is it exactly that these file systems depend on? Or, what would go wrong if we were to set stream_like automatically in libfuse? 

It seems to me that we should either make a strong statement that the two flags must always be used together, or need to supply more documentation on when it makes sense to set only nonseekable.

navytux:
@Nikratio, first of all thanks for feedback. Yes it is a bit confusing, mostly due to requirement to keep old "just nonseekable" semantic working. Let me try to clarify:

> From the description, it seems that at this point in time any reasonable filesystem would set both nonseekable and stream_like. Is that correct?

Yes, that's correct. Any reasonable filesystem that implements pipe-like of bidirectional socket-like file must set both `stream=1` and `nonseekable=1`.

> Then, I understand that we cannot implicitly set stream_like, because eg gvfs depends on it not being set. Is that correct?

Yes, that's correct. The reason why we just don't change `nonseekable=1` semantic to drop position is that there are filesystems that use `nonseekable=1` _and_ file position in e.g. read.

> However, as I understand this will lead to deadlocks, thus making such filesystems unusable anyway - so why do we have to care? Is the likelihood of the deadlock very small?

nonseekable=1 and read/write methods that use ppos don't always lead to deadlock. They lead to deadlock only if there is synchronization/waiting in between read and write. For example in kernel there are several files that use `nonseekable_open` (analog of nonseekable=1 in libfuse) but provide plain data buffer to read from or write to. For such cases read method checks the position and uses it to see from where to continue to read. Strictly speaking the file is seekable, because there _is_ a notion of position, and it will be just that `llseek` will fail because `nonseekable_open` was used. There is also no deadlock here because there is no waiting in between read and write.

Contrary to above simple case of `nonseekable=1` with buffer, for real stream files, if there is a dependency connection in between read and write (i.e. read is returning server reply to command that was written) the deadlock is _always_ there and appears 100% of the cases. And even without tx->rx cycle, if just receive operation is blocking (e.g. it receives data from serial port and there is no data coming), the write will be stucked waiting for read to wake up.

In other words for really stream-like files with read/write having blocking semantics the deadlock is always there and stream=1 just has to be used. However there are case where files are really seekable and use position in their read/write handlers that is _needed_ for read/write implementation correctness, and that in turn require the `f_pos_lock` to be there for read/write not to racily step onto each other toes.
And some of those seekable files - both in kernel and in userspace - historically use nonseekable_open e.g. due to being lazy to implement llseek and so `nonseekable=1` semantic, at least for userspace, cannot be changed to avoid providing ppos.

Speaking about gvfs specifically, if I understand it correctly, I see that for streams they are checking that `read(..., pos)` receives position at where last read was stopped. If we won't continue to provide position into that read, the codepath that checks `if (offset != fh->pos)` (https://gitlab.gnome.org/GNOME/gvfs/blob/1.40.0-6-gcbc54396/client/gvfsfusedaemon.c#L1258) will trigger which can lead to read operation failing.
It is indeed better that gvfs change its stream operations to use both `nonseekable=1` and `stream=1`
(/cc @alexlarsson @ondrejholy @rosslagerwall), but we cannot change already released versions, and also there can be other FUSE-based filesystems that depend on `nonseekable=1` still providing position. In other words we cannot change `nonseekable=1` semantic because of backward compatibility, even though in some cases the usage is subject to deadlock.

> Finally, what is it exactly that these file systems depend on? Or, what would go wrong if we were to set stream_like automatically in libfuse?

Please see above - e.g. those filesystem use position provided into read handler to see from where in their buffer they should continue to provide data.

> It seems to me that we should either make a strong statement that the two flags must always be used together, or need to supply more documentation on when it makes sense to set only nonseekable.

I suggest to go the strong statement way. Because nonseekable=1 + position looks like a bit of hack to me because people were lazy to either implement llseek returning -ENOTSUPP or to simply implement it.

I suggest to apply the following interdiff:

```diff
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -68,7 +68,8 @@ struct fuse_file_info {
        unsigned int flush : 1;
 
        /** Can be filled in by open, to indicate that the file is not
-           seekable. */
+           seekable. For pipe-like and socket-like files both nonseekable=1
+           and stream=1 must be set. */
        unsigned int nonseekable : 1;
 
        /* Indicates that flock locks for this file should be
@@ -83,7 +84,16 @@ struct fuse_file_info {
        unsigned int cache_readdir : 1;
 
        /** Can be filled in by open, to indicate that the file is
-           stream-like (no file position at all). */
+         * stream-like (no file position at all).
+         *
+         * Both stream=1 and nonseekable=1 must be set for pipe-like and
+         * socket-like files. See
+         *
+         *     git.kernel.org/linus/10dce8af3422, and
+         *     git.kernel.org/linus/bbd84f33652f
+         *
+         * for details.
+         */
        unsigned int stream : 1;
 
        /** Padding.  Reserved for future use*/
```

and add a bold warning in next libfuse release notes about the issue.

What do you think?

include/fuse_common.h
navytux: thanks, this makes sens. I suggest to go "both must be set way" + see interdiff proposed in main reply.

Nikratio:
Thanks for the explanations! I'm on vacation right now, will take a closer look in 2 weeks.

Quick impression: I feel like the steam like vs non seekable distinction is not useful, i'd rather call the new flag "truly_nonseekable"  or something like that and have or imply the other (so that we can retire that at dinyner point). And in addition to the mailing list link, I'd add a short summary along the lines of "if truly non seekable, offset values are undefined. If legacy non seekable, offset values will point to current position but deadlocks may occur"

navytux:
@Nikratio thanks for feedback and looking forward to your closer look when you are back. On "nonseekable" vs "trully nonseekable" I suggest to use the "stream" terminology as this is what the kernel is using. Being stream also implies being nonseekable (the kernel enforces it) and the same can be done at client side by libfuse (selecting nonseekable=1 if stream was set =1 just in case we are running on older kernel). Stream is also more clear (to me) compared to "trully nonseekable".

> I'd add a short summary along the lines of "if truly non seekable, offset values are undefined. If legacy non seekable, offset values will point to current position but deadlocks may occur"

Good idea to add such explanation. Only I would suggest to change "undefined" to something exact - there is no current position and pos passed to read/write handlers will be 0.

Anyway, looking forward to hear from you after vacation.

Nikratio:
I see. Ok, let's do this:

- Let's rename `nonseekable` to `nonseekable_broken`. This means that filesystems linked against old libfuse versions will continue to work, but attempting to compile against a new version will break. Prominently mention this in `ChangeLog.rst`

- Let's add a comment stating that (1) this flag causes deadlocks, (2) that `streamlike` should be used instead, and that (3) the difference is that with `streamlike` no offset is passed to read/write. Clearly mark the linux-kernel link as "background reading".

- Let's add a comment to `streamlike` that explains that write/read handler will always receive a zero offset.

- Do not require setting both `streamlike` and `nonseekable_broken` - if necessary, libfuse should internally set `nonseekable_broken`.

Sounds good?

Nikratio:
Are you still interested in working on this?

navytux:
Appologize for getting stuck. I was trying to finish something else and got completely swamped. I will try to provide feedback here in a day or two hopefuly.

navytux:
@Nikratio, first of all I appologize once again for the long delay with replying.

> - Let's rename nonseekable to nonseekable_broken. This means that filesystems linked against old libfuse versions will continue to work, but attempting to compile against a new version will break. Prominently mention this in ChangeLog.rst


I'm not sure that breaking libfuse backward compatibility - even source one - is acceptable.
However if backward compatibility break is acceptable I would expose only `stream` setting and not expose `nonseekable` setting at all, i.e. don't expose even `nonseekable_broken` or anything like that.
Of course if such breakage is decided to be made, it has to be done with incrementing FUSE API number in major way. I agree that making a bold statement in the changelog is required in any way - if, or if not - backward compatibility is lost.

By the way, we are talking not only about new libfuse releases, but also about backports for older libfuse releases too - see e.g. https://github.com/libfuse/libfuse/pull/435 and https://github.com/libfuse/libfuse/pull/436 as, from my point of view, it is not good to leave users of those older libfuse release out of luck to avoid the deadlock regression fixes.

> - Let's add a comment stating that (1) this flag causes deadlocks, (2) that streamlike should be used instead, and that (3) the difference is that with streamlike no offset is passed to read/write. Clearly mark the linux-kernel link as "background reading".

Agree.

> Let's add a comment to streamlike that explains that write/read handler will always receive a zero offset.

Agree.

> Do not require setting both streamlike and nonseekable_broken - if necessary, libfuse should internally set nonseekable_broken.

Agree.

----

I have to clarify one thing: personally I'm not using libfuse for my project (which is based on go-fuse). However because libfuse is kind of canonical library for FUSE-based filesystems (at least it was started from that) with this libfuse fixes I just wanted to be a good "citizen" by not leaving libfuse not notified and not adapted to the FUSE kernel fixes that got merged. Unfortunately it is a very tough time for me at present and I do not have the time resources to continue working on libfuse bits for at least several months. I appologize for this inconvenience, but unfortunately it means that either libfuse people pick this work up and carry it on, or it has to wait until I could have at least some spare cycles to continue working on this topic. Of course I would prefer to finish this myself if I have the chance, but at present I cannot afford that. I appologize for this inconvenience and hope for understanding...

Sorry once again for the delay with replying. Best,
Kirill

PR #438, opened 2019-07-16
fuse-lowlevel: set pipe size to max

giuseppe:
on failure to set the pipe size, set it to the maximum allowed by the
kernel.

If the first request required more than the maximum allowed, the
can_grow flag would be reset thus preventing any further resize.

Grow the pipe to the maximum allowed to increase the likelihood of
using splice for successive requests instead of falling back to
read/write.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -611,6 +611,35 @@ static int read_back(int fd, char *buf, size_t len)
 	return 0;
 }
 
+static int grow_pipe_to_max(int pipefd)
+{
+	int max;
+	int res;
+	int maxfd;
+	char buf[32];
+
+	maxfd = open("/proc/sys/fs/pipe-max-size", O_RDONLY);
+	if (maxfd < 0)
+		return -errno;
+
+	res = read(maxfd, buf, sizeof(buf) - 1);
+	if (res < 0) {
+		int saved_errno;
+
+		saved_errno = errno;
+		close(maxfd);
+		return -saved_errno;
+	}
+	close(maxfd);
+	buf[res] = '\0';
+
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks! Looking at this, wouldn't it be easier to get rid of the pipe growing logic and set it to the maximum size right from the beginning? Are there any disadvantages of having a large pipe size?

giuseppe:
> Thanks! Looking at this, wouldn't it be easier to get rid of the pipe growing logic and set it to the maximum size right from the beginning? Are there any disadvantages of having a large pipe size?

I don't think there are disadvantages beside wasting some memory.  AFAICS the main advantage for the lazy allocation that is used now is that on a host with `/proc/sys/fs/pipe-max-size` set to a big value, we will always use the biggest value even in the case there won't be any request that needs it.
With the current code instead we are sure we set the maximum value only if there was already a request that needed more than the maximum size.

giuseppe:
@Nikratio anything more to move this PR forward?

Nikratio:
Sorry for the delay. I suspect that just requesting a large size will not consume any memory unless someone actually writes that much data, and that it would therefore be okay to simplify the code. However, I don't have time to confirm so will do the conservative choice and merge as-is. Thanks!

bsbernd:
@yangyun50 Sorry had missed your question, yes this patch is wrong, kernel `fuse_dev_do_read` checks the max buffer size and if not provided it fails

`fuse_dev_do_read
```
			   fc->max_write))
		return -EINVAL;
```

bsbernd:
https://github.com/libfuse/libfuse/pull/1465

PR #445, opened 2019-08-20
Clarify bitfield padding issue

AsumFace:
The current situation is evidently devious and the next person adding another named bit field will probably be fooled by it (again) and shorten the padding field. This change should avert that while maintaining backwards compatibility.

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -83,7 +83,9 @@ struct fuse_file_info {
 	unsigned int cache_readdir : 1;
 
 	/** Padding.  Reserved for future use*/
-	unsigned int padding : 26;
+	unsigned int padding : 25;
+	/* the following int was once allocated by accident. kept for backwards compatibility */
+	unsigned int padding2 : 32;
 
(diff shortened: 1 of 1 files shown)


Nikratio:
I'm sorry, but what would be the problem with adding another flag and reducing the padding by one?

AsumFace:
It very likely does not do what the programmer intends. The current padding field is not located inside the same 32 bit block where the flags are due to being oversized by one bit, which is not apparent without closer inspection. Luckily, as long as the padding remains too large to fit, no breakage will occur, but who knows what breakage *might* occur based on wrong assumptions about the struct layout later on.

Nikratio:
Got you, thanks! But there is no reason to talk about backwards compatibility, etc for the second 32 bit, is there? It can be used in the future just like the remaining bits in the preceding byte.

AsumFace:
Yeah, the second 32 bit can be used up for additional flags as you'd expect. The change effectively only affects the naming of the padding that's already there after all.

Nikratio:
Could you change the comment appropriately then?

PR #448, opened 2019-09-03
Add fuse_log.h APIs for custom log message handlers

stefanha:
Introduce an API for custom log handler functions.  This allows libfuse
applications to send messages to syslog(3) or other logging systems.
See include/fuse_log.h for details.

Convert libfuse from fprintf(stderr, ...) to log_fuse(level, ...).  Most
messages are error messages with FUSE_LOG_ERR log level.  There are also
some debug messages which now use the FUSE_LOG_DEBUG log level.

Note that lib/mount_util.c is used by both libfuse and fusermount3.
Since fusermount3 does not link against libfuse, we cannot call
fuse_log() from lib/mount_util.c.  This file will continue to use
fprintf(stderr, ...) until someone figures out how to split it up.

Signed-off-by: Stefan Hajnoczi <stefanha@redhat.com>

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,10 @@
+Unreleased Changes
+==================
+
+* Added custom log message handler function support so that libfuse
+  applications can direct messages to syslog(3) or other logging systems.
+  stderr remains the default.  See `fuse_log.h` for the new API.
+
 libfuse 3.6.2 (2019-07-09)
 ==========================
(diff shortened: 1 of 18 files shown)


Nikratio:
Thanks, this is great! Could you please add an entry to `Changelog.rst` as well?

stefanha:
@Nikratio Done!

PR #449, opened 2019-09-09
passthrough_ll: fix fallocate variant ifdefs

dagrh:
If fallocate isn't available we incorrectly check for the value of
HAVE_POSIX_FALLOCATE rather than it being defined.
We also fail to initialise 'err' in the case where neither are defined.

Fixes: 5fc562c90d7925963467 ("Add fallocate and use it instead of ...")

Signed-off-by: Dr. David Alan Gilbert <dgilbert@redhat.com>


Nikratio:
Thanks!

PR #450, opened 2019-09-10
Export log api

stefanha:
Oops, I forgot to add the fuse_log.h API to lib/fuse_versionscript so that the symbols are actually public.

Also export the fuse_log() function so applications can do unified logging through it and demonstrate this in passthrough_ll.c.


Nikratio:
Thanks! Could you please squash the first two commits? Looks like the second is just a fixup of the first.

stefanha:
@Nikratio Thanks for the feedback, I have squashed the two first commits.

Issue #473, opened 2019-11-19
Add FUSE_EXPLICIT_INVAL_DATA support


hselin:
Wire FUSE_EXPLICIT_INVAL_DATA support into libfuse

PR: #474

PR #474, opened 2019-11-19
Add support for FUSE_EXPLICIT_INVAL_DATA

hselin:
Enable only invalidate cached pages on explicit request.
Let me know if anything needs to be changed. Thanks!

Issue: #473

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -359,6 +359,29 @@ struct fuse_loop_config {
 #define FUSE_CAP_NO_OPENDIR_SUPPORT    (1 << 24)
 
 /**
+ * Indicates support for invalidating cached pages only on explicit request.
+ *
+ * If this flag is set in the `capable` field of the `fuse_conn_info` structure,
+ * then the FUSE kernel module supports invalidating cached pages only on
+ * explicit request by the filesystem through fuse_lowlevel_notify_inval_inode()
+ * or fuse_invalidate_path().
+ *
+ * By setting this flag in the `want` field of the `fuse_conn_info` structure,
+ * the filesystem is responsible for invalidating cached pages through explicit
+ * requests to the kernel.
+ *
+ * Note that setting this flag does not prevent the cached pages from being
+ * flushed by OS itself and/or through user actions.
+ *
+ * Note that if both FUSE_CAP_EXPLICIT_INVAL_DATA and FUSE_CAP_AUTO_INVAL_DATA
+ * are set in the `capable` field of the `fuse_conn_info` structure then
+ * FUSE_CAP_AUTO_INVAL_DATA takes precedence.
+ *
+ * This feature is disabled by default.
+ */
+#define FUSE_CAP_EXPLICIT_INVAL_DATA    (1 << 25)
+
+/**
  * Ioctl flags
  *
  * FUSE_IOCTL_COMPAT: 32bit compat ioctl on 64bit machine
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,9 @@
+Unreleased Changes
+==========================
+
+* Added support for FUSE_EXPLICIT_INVAL_DATA to enable
+  only invalidate cached pages on explicit request.
+
 libfuse 3.8.0 (2019-11-03)
 ==========================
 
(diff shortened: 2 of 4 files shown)


Nikratio:
Thanks! Could you please also update ChangeLog.rst? (if necessary, add a new heading "Unreleased Changes").

hselin:
@Nikratio 
Updated ChangeLog.rst.
Let me know if it is okay with you. Thanks!

include/fuse_common.h
```diff
@@ -358,6 +358,16 @@ struct fuse_loop_config {
  */
 #define FUSE_CAP_NO_OPENDIR_SUPPORT    (1 << 24)
 
+/**
+ *
+ * Indicates support for only invalidate cached pages on explicit request.
+ * If both FUSE_CAP_EXPLICIT_INVAL_DATA and FUSE_CAP_AUTO_INVAL_DATA are set,
+ * then FUSE_CAP_AUTO_INVAL_DATA takes precedence.
```
Nikratio: Could you clarify if this refers to the `supported` or `capable` field?

include/fuse_common.h
```diff
@@ -358,6 +358,16 @@ struct fuse_loop_config {
  */
 #define FUSE_CAP_NO_OPENDIR_SUPPORT    (1 << 24)
 
+/**
+ *
+ * Indicates support for only invalidate cached pages on explicit request.
```
Nikratio: Could you elaborate on what this means for the filesystem? How would the filesystem make such a request?

hselin: Agreed on the grammar. I will fix that.
I will also add a snippet on the usage and how it will not protect from OS flushing the cache.

How does the following sound?

```
/**
 * Indicates support for invalidating cached pages only on explicit request.
 *
 * If this flag is set in the `capable` field of the `fuse_conn_info` structure,
 * then the FUSE kernel module supports invalidating cached pages only on explicit
 * request by the filesystem.
 * 
 * By setting this flag in the `want` field of the `fuse_conn_info` structure,
 * the filesystem is responsible for invalidating cached pages through explicit
 * requests to the kernel.
 *
 * Note that setting this flag does not prevent the cached pages from being flushed
 * by OS itself and/or through user actions.
 *
 * Note that if both FUSE_CAP_EXPLICIT_INVAL_DATA and FUSE_CAP_AUTO_INVAL_DATA
 * are set in the `capable` field of the `fuse_conn_info` structure then
 * FUSE_CAP_AUTO_INVAL_DATA takes precedence.
 *
 * This feature is disabled by default.
 */
```

include/fuse_common.h
hselin: Will do. I've pasted an updated comment section below. Please take a look.

include/fuse_common.h
Nikratio: SGTM, but could you clarify how the filesystem would make such an explicit request to the kernel? (i.e. which function needs to be called)

hselin: Ah... I misunderstood your point of 'how would the filesystem make such a request'.
I thought you were asking how would the filesystem enable this 'explicit invalidation mode', but you were asking how would the filesystem send the 'explicit invalidation request' :)

I have been using fuse_invalidate_path() function for this purpose.
I think you can also write into '/proc/sys/vm/drop_caches' to trigger a cache flush.
Do you know any other functions?

Nikratio: I am not trying to test you here. I am asking you how the filesystem needs to create such an "explicit invalidation request" because I do not know what the kernel expects. From your answer, I think the answer is that a standard notify_inval_inode is sufficient. So let's just write that ("..only on explicit request by the file system (through fuse_lowlevel_notify_inval_inode() or fuse_invalidate_path()").

hselin: No worries, I didn't think you are testing me :)
Sorry if I suggested otherwise.

I agree it is important to understand and document the "explicit invalidation request".
I've been using fuse_invalidate_path() and '/proc/sys/vm/drop_caches' but I genuinely don't know if these are the only way to do this. That is why I asked the question about other functions.

Thanks! I'll modify the comments to what you've suggested above.

hselin: @Nikratio sorry for the delay. Here is the updated comment section.
Let me know if we should change anything else. Thanks!

PR #485, opened 2020-01-03
State GPL version in comment

dagrh:
In a bunch of comments we say 'under the terms of the GNU GPL', make
it clear this is GPLv2 (as LICENSE says).

Signed-off-by: Dr. David Alan Gilbert <dgilbert@redhat.com>

### DIFF
--- a/example/cuse.c
+++ b/example/cuse.c
@@ -3,7 +3,7 @@
   Copyright (C) 2008-2009  SUSE Linux Products GmbH
   Copyright (C) 2008-2009  Tejun Heo <tj@kernel.org>
 
-  This program can be distributed under the terms of the GNU GPL.
+  This program can be distributed under the terms of the GNU GPLv2.
   See the file COPYING.
 
 */
--- a/example/cuse_client.c
+++ b/example/cuse_client.c
@@ -3,7 +3,7 @@
   Copyright (C) 2008       SUSE Linux Products GmbH
   Copyright (C) 2008       Tejun Heo <teheo@suse.de>
 
-  This program can be distributed under the terms of the GNU GPL.
+  This program can be distributed under the terms of the GNU GPLv2.
   See the file COPYING.
 */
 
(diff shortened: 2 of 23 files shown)


Nikratio:
I'm not sure about this one. I do not know what the original authors intended for all these files. LICENSE itself was written by me in 2018 as a best guess to at least distinguish better between GPL and LGPL. To me, changing all these files feels like pretending that there is more certainty than there actually is.

dagrh:
If there's uncertainty about the actual intent of the license we have a bigger problem! People (i.e. me!) might be attempting to take the files and use them in a GPLv2 project - and that's how this came up, because a reviewer pointed out they were uncomfortable with the 'GPL' in passthrough_ll.c.
Now, looking at passthrough_ll.c,  it was created in f55e489 by Miklos in 2014, and it said:

  This program can be distributed under the terms of the GNU GPL.
  See the file COPYING.

and at the time the COPYING file said:
                    GNU GENERAL PUBLIC LICENSE
                       Version 2, June 1991
..
so that seems consistent to me; (COPYING has been there since Nov 2001) and there are some older files.

We've currently not got a COPYING file - since they got renamed; which is a bit confusing since we now have all the headers saying they're GPL and pointing to a non-existent file for clarification.

There's also at least one inconsistency with teh LICENSE file, which is example/passthrough_helpers.h which has, what I think isa  BSD license.

Nikratio:
Yeah, that's what I'm saying - there is a bigger problem :-).

Maybe we can tackle the issue file by file? If we have specific evidence of the intention (as you dug up for passthrough_ll.c), then I'll happily apply patches to the headers.

szmi:
The project predates GPLv3 and the intention has definitely been to refer to the then actual version (GPLv2).

Nikratio:
@szmi, are you the original author of all the files in the distribution (other than those later added by me)?

szmi:
All except:

 - bsd mount support (Csaba Henk)
 - cuse and ioctl/poll examples (Tejun Heo)

PR #490, opened 2020-01-16
install_helper: Install init_script in sysconfdir

### LINKED ISSUE #489: Install script ignores prefix variable when installing fuse3 init script
# How to reproduce
```
meson --prefix="${PWD}/prefix" -Duseroot=false -Dudevrulesdir="${PWD}/prefix/lib/udev/rules.d" libfuse build
ninja -C build install
```
```
Installing /home/ubuntu/libfuse/doc/mount.fuse3.8 to /home/ubuntu/prefix/share/man/man8
Installing /home/ubuntu/build/meson-private/fuse3.pc to /home/ubuntu/prefix/lib/x86_64-linux-gnu/pkgconfig
Running custom install script '/home/ubuntu/libfuse/util/install_helper.sh /home/ubuntu/prefix/etc /home/ubuntu/prefix/bin /home/ubuntu/prefix/lib/udev/rules.d false'
Failed to run install script '/home/ubuntu/libfuse/util/install_helper.sh /home/ubuntu/prefix/etc /home/ubuntu/prefix/bin /home/ubuntu/prefix/lib/udev/rules.d false'
FAILED: meson-install 
```
# Problem
install_helper.sh is trying to install `util/init_script` into `/etc/init.d/fuse3` even if prefix is specified. It is possible to solve problem by adding DESTDIR environment variable but the installation of `init.d/fuse3` becomes inconsitent with the installation of other configuration files. All other configuration files are installing into sysconfdir directory and according to [GNU coding standarts](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) I assume that `init.d/fuse3` is also should do this.

bobrofon:
fuse3 init script was always installing to /etc directory even if prefix
variable is defined. Now it is installing to sysconfdir directory.
Also update-rc.d was running even if using root permissins was
explicitly disabled via useroot definition.

Fixes #489

### DIFF
--- a/util/install_helper.sh
+++ b/util/install_helper.sh
@@ -40,14 +40,14 @@ install -D -m 644 "${MESON_SOURCE_ROOT}/util/udev.rules" \
         "${DESTDIR}${udevrulesdir}/99-fuse3.rules"
 
 install -D -m 755 "${MESON_SOURCE_ROOT}/util/init_script" \
-        "${DESTDIR}/etc/init.d/fuse3"
+        "${DESTDIR}${sysconfdir}/init.d/fuse3"
 
 
-if test -x /usr/sbin/update-rc.d && test -z "${DESTDIR}"; then
+if test -x /usr/sbin/update-rc.d && test -z "${DESTDIR}" && $useroot; then
     /usr/sbin/update-rc.d fuse3 start 34 S . start 41 0 6 . || /bin/true
 else
     echo "== FURTHER ACTION REQUIRED =="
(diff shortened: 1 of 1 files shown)


Nikratio:
I believe with this change the default directory for the init script is /usr/local/etc. Is that correct? 

Having the default be /etc even if libfuse is installed to /usr/local is a feature, not a bug (libraries are readily usable when installed into /usr/local, but an initscript there will do nothing)

Nikratio:
Close per comment above.

PR #491, opened 2020-01-17
meson_options: Add initscriptsdir option

### LINKED ISSUE #489: Install script ignores prefix variable when installing fuse3 init script

bobrofon:
Add meson option 'initscriptsdir' to explicitly define location of
system initialization scripts. 'initscriptsdir' is '/etc/init.d' by
default.
Option usage:
meson --prefix=/usr/local -Dinitscriptsdir=/custom/init.d libfuse-src-dir build-dir
DESTDIR="${PWD}/install-dir" ninja -C build-dir

As a result, init script will be installed into
"${PWD}/install-dir/custom/init.d" location.

Fixes #489

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -13,3 +13,5 @@ option('examples', type : 'boolean', value : true,
 option('useroot', type : 'boolean', value : true,
        description: 'Set owner and setuid bits on installed files')
 
+option('initscriptsdir', type : 'string', value : '/etc/init.d',
+       description: 'Where to install init scripts')
--- a/util/install_helper.sh
+++ b/util/install_helper.sh
@@ -10,6 +10,7 @@ sysconfdir="$1"
 bindir="$2"
 udevrulesdir="$3"
 useroot="$4"
+initscriptsdir="$5"
 
 # Both sysconfdir and bindir are absolute paths (since they are joined
 # with --prefix in meson.build), but need to be interpreted relative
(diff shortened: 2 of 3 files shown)


Nikratio:
I do not think the extra complexity justifies the benefits here, sorry.

PR #495, opened 2020-01-30
Optimize high level id/name hash efficiency

tianshan:
Signed-off-by: Tianshan Qu <qutianshan@gmail.com>

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -516,8 +516,8 @@ static void free_node_mem(struct fuse *f, struct node *node)
 
 static size_t id_hash(struct fuse *f, fuse_ino_t ino)
 {
-	uint64_t hash = ((uint32_t) ino * 2654435761U) % f->id_table.size;
-	uint64_t oldhash = hash % (f->id_table.size / 2);
+	uint64_t hash = ((uint32_t) ino * 2654435761U) & (f->id_table.size - 1);
+	uint64_t oldhash = hash & (f->id_table.size / 2 - 1);
 
 	if (oldhash >= f->id_table.split)
 		return oldhash;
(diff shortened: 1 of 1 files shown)


Nikratio:
Why is this more efficient?

tianshan:
@Nikratio bitwise AND is faster than mod operator

trapexit:
Yes, but 1) have you ensured that the table is always a power of 2? 2) Unless the hash is called very often it won't really matter in practice on modern machines.

tianshan:
@trapexit  
1) the initialize value is hard coded 8192 (NODE_TABLE_MIN_SIZE) assign in _node_table_init_,  table grow is always double in _node_table_resize_, table reduce is halved in _node_table_reduce_ 
2) id_hash and name_hash are called each request in lookup phase.
but indeed, the optimize is so little to observe in test

jpandre:
Consider the code below :

unsigned int mod(unsigned int w)
{
        return (w % 8192);
}

gcc compiles it as :
        movl    %edi, %eax
        andl    $8191, %eax
        ret

So gcc does the optimization when relevant.

tianshan:
@jpandre I think you are right, after enable compile optimization, the two ways are almost the same.
I will close this pr, thanks for yours time.

Nikratio:
I'm not surprised that GCC optimizes division by a hardcoded value, but I doubt that it will do the same if the divisor is variable (as in the code). It's rather unlikely that static analysis can tell that `f->id_table.size` will always be a power of two, and checking this at runtime would incur an extra penalty.

The reason I haven't applied this yet is that I it makes the code *depend* on the table size being a power of two. This is currently true, but people may change it without realizing that this is a hidden requirement.

If someone wants to push this, then the best way forward would probably be to store just the exponent of the table size, which would make the requirement explicit.

PR #505, opened 2020-02-24
Fix issue preventing using splice with reads

jyvet:
Context: SPLICE_WRITE is not used with regular buffers
(i.e. when they are not file-descriptor backed buffers).
There is a bug which assumes file descriptors are used.
If the amount of data associated with those FD is lower
than twice the page size, SPLICE_WRITE is not utilized.
With regular buffers the aggregated size is always 0.

This patch prevents to fallback to non-splice enabled
copies if FD total size is equal to 0 (basically no
copy involving FDs).

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -650,7 +650,7 @@ static int fuse_send_data_iov(struct fuse_session *se, struct fuse_chan *ch,
 	struct fuse_ll_pipe *llp;
 	int splice_flags;
 	size_t pipesize;
-	size_t total_fd_size;
+	size_t total_buf_size;
 	size_t idx;
 	size_t headerlen;
 	struct fuse_bufvec pipe_buf = FUSE_BUFVEC_INIT(len);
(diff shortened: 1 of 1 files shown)


Nikratio:
I don't quite understand. If the source is a regular buffer, splicing doesn't make any sense. As splice(2) says: "splice()  moves data between two file descriptors without copying between kernel address space and user address space". 

Am I missing something here?

jyvet:
Thank you for your comment. It works in that case because of the vmsplice (splice user pages to/from a pipe) call in lib/fuse_lowlevel.c.
We are using this capability with IME burst buffer. 

I'll resubmit another version today. I think we should keep the 2 pages limit (as using splice calls on small sizes may decrease performance).

jyvet:
Sorry for the delay. Here are some results with a parallel burst buffer using FUSE (vmsplice + splice from/to user space buffers).

IOR with 10 MPI tasks, 40s runs, 5 iterations (single client node with fuse on Centos 7.6, NUMA affinity set): 
```

4K random with splice:
Max Write: 681.28 MiB/sec (714.38 MB/sec)
Max Read:  331.25 MiB/sec (347.34 MB/sec)

4K random without splice:
Max Write: 584.35 MiB/sec (612.74 MB/sec)
Max Read:  321.47 MiB/sec (337.09 MB/sec)

512B random with splice:
Max Write: 60.41 MiB/sec (63.34 MB/sec)
Max Read:  40.81 MiB/sec (42.79 MB/sec)

512B random without splice:
Max Write: 87.39 MiB/sec (91.64 MB/sec)
Max Read:  47.51 MiB/sec (49.82 MB/sec)
```

It confirms using splice brings overheads on small sizes.

PR #514, opened 2020-05-14
passthrough_ll: remove symlink fallbacks

szmi:
Path lookup in the kernel has special rules for looking up magic symlinks
under /proc.  If a filesystem operation is instructed to follow symlinks
(e.g. via AT_SYMLINK_FOLLOW or lack of AT_SYMLINK_NOFOLLOW), and the final
component is such a proc symlink, then the target of the magic symlink is
used for the operation, even if the target itself is a symlink.  I.e. path
lookup is always terminated after following a final magic symlink.

I was erronously assuming that in the above case the target symlink would
also be followed, and so workarounds were added for a couple of operations
to handle the symlink case.  Since the symlink can be handled simply by
following the proc symlink, these workardouds are not needed.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -121,7 +121,6 @@ typedef std::unordered_map<SrcId, Inode> InodeMap;
 
 struct Inode {
     int fd {-1};
-    bool is_symlink {false};
     dev_t src_dev {0};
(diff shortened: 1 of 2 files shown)


Nikratio:
Would you have time to also update `examples/passthrough_hp.cc`?

PR #525, opened 2020-07-02
fuse_lowlevel: Remove useless NULL assert for se

winndows:
Remove useless NULL assert for se.

Signed-off-by: Liao Pingfang <liao.pingfang@zte.com.cn>

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -189,8 +189,6 @@ static int fuse_send_msg(struct fuse_session *se, struct fuse_chan *ch,
 	int err = errno;
 
 	if (res == -1) {
-		assert(se != NULL);
-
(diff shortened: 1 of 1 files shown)


Nikratio:
Could you explain why this is useless?

winndows:
If it is useful, line 171: "if (se->debug) {" in this file will dereference NULL pointer.

winndows:
> Could you explain why this is useless?

If it is useful, line 171: "if (se->debug) {" in this file will dereference NULL pointer.

Nikratio:
That's actually a bug above, then. Probably the assert should go there, or the check for se->debug should first check for se !+= NULL.

winndows:
> Could you explain why this is useless?

If it is useful, line 171: "if (se->debug) {" in this file will dereference NULL pointer.

PR #532, opened 2020-07-11
Define fuse_session_loop_mt as a macro on uclibc and MacOS

asafkahlon:
On uclibc and MacOS we don't use versioned symbols. Hence,
there's no definition for fuse_session_loop_mt on those cases
and the linker won't be able to resolve calls to fuse_session_loop_mt()

Signed-off-by: Asaf Kahlon <asafka7@gmail.com>

### DIFF
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -471,7 +471,7 @@ struct fuse_lowlevel_ops {
 	 *  - When writeback caching is disabled, the filesystem is
 	 *    expected to properly handle the O_APPEND flag and ensure
 	 *    that each write is appending to the end of the file.
-	 *
+	 * 
          *  - When writeback caching is enabled, the kernel will
 	 *    handle O_APPEND. However, unless all changes to the file
 	 *    come through the kernel this will not work reliably. The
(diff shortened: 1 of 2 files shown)


include/fuse_lowlevel.h
```diff
@@ -471,7 +471,7 @@ struct fuse_lowlevel_ops {
 	 *  - When writeback caching is disabled, the filesystem is
 	 *    expected to properly handle the O_APPEND flag and ensure
 	 *    that each write is appending to the end of the file.
-	 * 
+	 *
```
yann-morin-1998, line 474: Spurious, unrelated space cleanup. Ditto below.

lib/fuse_versionscript
```diff
@@ -146,6 +146,7 @@ FUSE_3.2 {
 	global:
 		fuse_session_loop_mt;
 		fuse_session_loop_mt_31;
+                fuse_session_loop_mt_32;
```
yann-morin-1998, line 149: The lines around use leading TABs, not spaces.

asafkahlon:
@yann-morin-1998 fixed your comments, thanks.

PR #533, opened 2020-07-11
Add an option to disable building tests

asafkahlon:
Signed-off-by: Asaf Kahlon <asafka7@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -96,13 +96,17 @@ thread_dep = dependency('threads')
 #
 # Read build files from sub-directories
 #
-subdirs = [ 'lib', 'include', 'test' ]
+subdirs = [ 'lib', 'include']
 if get_option('utils') and not platform.endswith('bsd') and platform != 'dragonfly'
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks for the patch! Can you explain the use-case in which not compiling (and thus running) the tests is desirable?

asafkahlon:
The case is not compiling and not running the tests, which is currently not possible when just invoking a regular build with `meson`.

If, for example, you compile an embedded linux system and you want to include `libfuse` in it, you don't want to compile the tests because they won't be installed on the target board.

Nikratio:
Not sure I follow. Tests will never be installed, no matter if embedded or not. They should still be run before installation though, to ensure that everything works.

asafkahlon:
@Nikratio Yeah, I know the tests won't be installed, but why would you want to compile them over and over again unless explicitly asked?
Anyway, you need to explicitly run another command in order to run the tests, so I don't think we should compile them if we don't have an intention to actually run them too.

PR #545, opened 2020-09-08
Implement GCC 10 style symbol versioning

spotrh:
Previously, libfuse was using top level ASM constructs to implement symbol versioning. These constructs are incompatible with Link Time Optimization (LTO). GCC v10 introduces a new mechanism for symbol versioning which works with LTO.

This PR adds a meson check to verify that the build compiler supports the symver attribute and if not, falls back to the ASM constructs.

I am submitting this change on behalf of Jeff Law <law@redhat.com> with his permission.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: xenial
+dist: bionic
 
 language:
     - c
@@ -13,8 +13,10 @@ addons:
     - valgrind
     - clang
     - libstdc++-7-dev 
+    - libstdc++-10-dev
     - gcc
     - gcc-7
+    - gcc-10
     - python3-pip
     - python3-setuptools
     - ninja-build
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -4569,7 +4569,7 @@ int fuse_loop(struct fuse *f)
 	return fuse_session_loop(f->se);
 }
 
-FUSE_SYMVER(".symver fuse_loop_mt_32,fuse_loop_mt@@FUSE_3.2");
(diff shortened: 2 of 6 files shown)


Nikratio:
Thanks! Could you maybe update the CI script (tests/travis-build.sh IIRC) to attempt a build with link time optimization to make sure this works as expected?

spotrh:
Your travis environment is using Ubuntu 16.04 LTS (Xenial). I do not believe there are gcc-10 packages (even in the compiler test ppa) for Xenial. If you move to Ubuntu 18.04 LTS (Bionic), I can add a PPA to .travis.yml and include gcc-10 in the packages list. Once that happens, I can easily update the CI script.

My testing has been on Fedora 33, but Bionic + ppa:ubuntu-toolchain-r/test should be equivalent.

Are you okay with that change?

Nikratio:
Please feel free to update as needed. The only reason for using Xenial is probably because this read up to date when the CI was set up.

Nikratio:
Thanks!

Nikratio:
Hmm. Looks like this doesn't quite work (not sure why the failing CI didn't show up above):

https://travis-ci.org/github/libfuse/libfuse

Could you take another look?

spotrh:
I can... but it doesn't look like Travis generated any useful logging here:

https://travis-ci.com/github/libfuse/libfuse/builds/183868601

Nikratio:
That's odd - the last time I looked at it, the logs were there.

The error was something about gcc-7 complaining about `-Wattributes`.

PR #547, opened 2020-09-09
Updated example code to work with new API

AKowshik:
Updated the example programs that would fail on updating FUSE_USE_VERSION to 34. Used the newer versions of certain functions that would cause the program to fail.


Nikratio:
Thanks!

PR #549, opened 2020-09-11
Adjust .travis.yml to use ppa syntax

spotrh:
I think this is what is confusing travis... difficult for me to test this locally though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: xenial
+dist: bionic
 
 language:
     - c
@@ -13,8 +13,10 @@ addons:
     - valgrind
     - clang
     - libstdc++-7-dev 
+    - libstdc++-10-dev
     - gcc
     - gcc-7
+    - gcc-10
     - python3-pip
     - python3-setuptools
     - ninja-build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 dist: bionic
 
-before_script:
+before_install:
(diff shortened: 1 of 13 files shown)


spotrh:
Okay, so Travis is working again, looks like the meson check needs to be improved.

spotrh:
Well, that took longer than I thought it would. There's a few things here:

1. The original meson.build test was incorrect, it was reporting that it supported the symver attribute in gcc 7 when it most certainly did not. That is now fixed with a more correct sample code.
2. Moving from xenial to bionic broke some things. Specifically, there was a warning about an undefined marker (uses_fuse), which I defined.
3. The version of valgrind in bionic was throwing a WARNING because it did not know how to handle amd64 syscall 326 (copy_file_range). Upstream added that support in valgrind 3.16.1, but there was no available build of 3.16.1 for bionic... so I made a build and made a PPE for it to live in.
4. When I added the valgrind PPE to the list of sources for the .travis.yml, Travis started to throw errors pulling down the corresponding public key from the Ubuntu keyserver. This is an itinerant failure, sometimes it would succeed, but it looked like it was failing more often than not. Best guess is that the Ubuntu keyserver is having issues (there are lots of other people who have this issue with the Ubuntu keyserver and Travis)
5. So... I added some logic to manually install the public key and the updated valgrind and that seems to work all of the time.
6. Oh, and I also spent far too much time trying to convince the travis buld bash script to pass two options as an array to meson. Eventually gave up and just used the most important one.

Apologies for the commit spam.

Nikratio:
Thank you very much for figuring all this out!

PR #551, opened 2020-09-16
Allow caching symlinks in kernel page cache.

eduble:
This commit defines a new capability called `FUSE_CAP_CACHE_SYMLINKS`.
It is off by default but you can now enable it by setting this flag in
in the `want` field of the `fuse_conn_info` structure.

When enabled, the kernel will save symlinks in its page cache,
by making use of the feature introduced in kernel 4.20:
https://github.com/torvalds/linux/commit/5571f1e65486be025f73fa6aa30fb03725d362a2

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,7 +1,7 @@
 Unreleased Changes
 ==================
 
-* (none so far)
+* Add FUSE_CAP_CACHE_SYMLINKS: allow caching symlinks in kernel page cache.
 
 libfuse 3.9.4 (2020-08-09)
 ==========================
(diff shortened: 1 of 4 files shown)


Nikratio:
Thank you!

Could you please add an entry to `ChangeLog.rst` and expand the comment in `fuse_common.h` to explain how to invalidate the cached links? (I presume it's either with `fuse_inval_entry` or `fuse_inval_ino`?)

eduble:
Hi,
I amended the commit, let me know if it is OK.
Thanks
Etienne

Nikratio (approved):

PR #574, opened 2020-12-24
util/fusermount.c: Assume the kernel supports UMOUNT_NOFOLLOW

sargun:
UMOUNT_NOFOLLOW was added in Kernel 2.6.34. It's been 10 years since it's been added
Kernel 5.9, and 5.10 break this check mechanism[1]. Let's deprecate it.

[1]: https://lore.kernel.org/linux-fsdevel/20201223102604.2078-1-sargun@sargun.me/

Signed-off-by: Sargun Dhillon <sargun@sargun.me>

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -397,26 +397,12 @@ static int chdir_to_parent(char *copy, const char **lastp)
 }
 
 #ifndef IGNORE_MTAB
-/* Check whether the kernel supports UMOUNT_NOFOLLOW flag */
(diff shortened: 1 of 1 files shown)


sargun:
This should be backported to 2.9 as well. What's the process for that?

Nikratio:
Thanks!

The backport "process" is to send a pull request for the 2.9 branch and cross your fingers that at some point there will be a release again (very unlikely).

PR #582, opened 2021-01-28
Ignore "-o nonempty"

skitt:
Commit 0bef21e8543d removed "-o nonempty" since mounting over
non-empty directories is always allowed. But this broke tools which
specify "-o nonempty". Since the expected behaviour is the same
anyway, ignoring the "nonempty" option seems safe, and allows programs
specifying "-o nonempty" to continue working with fusermount3.

This would fix https://bugs.debian.org/939767

Signed-off-by: Stephen Kitt <steve@sk2.org>

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,10 @@
+Unreleased Changes
+==================
+
+* Allow "nonempty" as a mount option, for backwards compatibility with
+  fusermount 2. The option has no effect since mounting over non-empty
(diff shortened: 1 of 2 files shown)


Nikratio:
Could you also add an entry to `ChangeList.rst`?

skitt:
> Could you also add an entry to `ChangeList.rst`?

Of course; I followed the model used previously, with an “Unreleased Changes” section.

Issue #427, opened 2019-06-13
fuse3 init script is still installed in /usr/local/etc instead of /etc


iamorphen:
I believe this is identical to #178, but the issue persists in libfuse 3.6.1.
The manual fix in #178 can be used to correctly install the init script. Unfortunately, I do not have a corresponding PR at this time.

OS: Ubuntu 16

```sh
$ tar xf fuse-3.6.1.tar.xz
$ cd fuse-3.6.1/
$ mkdir build; cd build
build $ ../../meson-0.50.1/meson.py .. # native meson version is too old
build $ ninja
build $ sudo python3 -m pytest test/ # no issues
build $ sudo ninja install
```
The last line of output from the install attempt indicates `update-rc.d: error: unable to read /etc/init.d/fuse3`.

PR #598, opened 2021-05-03
install_helper: fix sysconf dir for init.d script

paralin:
Fixes the following build error in Buildroot as a host package:

```
Running custom install script 'install_helper.sh /host/etc /host/bin /host/lib/udev/rules.d false
+ sysconfdir=/host/etc
+ bindir=/host/bin
+ udevrulesdir=/host/lib/udev/rules.d
+ useroot=false
+ '[' -z '' ']'
+ DESTDIR=
+ install -D -m 644 /build/host-libfuse3-3.10.3/util/fuse.conf /host/etc/fuse.conf
+ false
+ install -D -m 644 /build/host-libfuse3-3.10.3/util/udev.rules /host/lib/udev/rules.d/99-fuse3.rules
+ install -D -m 755 /build/host-libfuse3-3.10.3/util/init_script /etc/init.d/fuse3
FAILED: install script 'install_helper.sh /host/etc /host/bin /host/lib/udev/rules.d false' exit code 1, stopped
FAILED: meson-install
```

As far as I can tell, this is correct - and the current implementation is wrong - because it should be installing the init.d script to sysconfdir.

### DIFF
--- a/util/install_helper.sh
+++ b/util/install_helper.sh
@@ -38,16 +38,3 @@ fi
 
 install -D -m 644 "${MESON_SOURCE_ROOT}/util/udev.rules" \
         "${DESTDIR}${udevrulesdir}/99-fuse3.rules"
-
-install -D -m 755 "${MESON_SOURCE_ROOT}/util/init_script" \
-        "${DESTDIR}/etc/init.d/fuse3"
-
-
-if test -x /usr/sbin/update-rc.d && test -z "${DESTDIR}"; then
-    /usr/sbin/update-rc.d fuse3 start 34 S . start 41 0 6 . || /bin/true
-else
-    echo "== FURTHER ACTION REQUIRED =="
-    echo "Make sure that your init system will start the ${DESTDIR}/etc/init.d/fuse3 init script"
-fi
-
-


Nikratio:
Please take a look at the history of this file. Both settings are causing problems in different circumstances (e.g. your change would re-open https://github.com/libfuse/libfuse/issues/427).

It is not clear to me what the right way to fix this is.

paralin:
@Nikratio while an appropriate solution is found, could you possibly add an option to disable installing the init.d script?

Nikratio:
Pull requests for making init script installation optional are welcome :-).

I'll this pull request, since it seems it's not really making things better than the status quo.

eli-schwartz:
> Both settings are causing problems in different circumstances

From the sound of it, the answer is no.

Some people don't *want* it to use anything other than /etc because if you configure with a prefix of /usr/local, the *default* sysconfdir is /usr/local/etc while the init script expects that sysconfdir is /etc.

Some people want it to respect sysconfdir, because they don't want the init script at all, and cannot write to /etc.

Simply removing it altogether was probably not a perfectly ideal attempt at a solution tbh.

paralin:
Probably not ideal, but I certainly was alarmed to find that the install script was writing to my host machine's /etc, while I was trying to get it to write to a "target" FS in Buildroot.

PR #600, opened 2021-05-06
Fix doxygen warnings.

dancerj:
Some parameters were undocumented.

tested with `doxygen doc/Doxyfile`

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -667,6 +667,10 @@ struct fuse_operations {
 	 */
 	int (*bmap) (const char *, size_t blocksize, uint64_t *idx);
 
+#if FUSE_USE_VERSION < 35
+	int (*ioctl) (const char *, int cmd, void *arg,
+		      struct fuse_file_info *, unsigned int flags, void *data);
+#else
 	/**
 	 * Ioctl
 	 *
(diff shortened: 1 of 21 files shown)


dancerj:
Not sure if I want to remove the source code from individual pages or fix the 2 pages to include the source code.

include/fuse.h
```diff
@@ -664,6 +667,10 @@ struct fuse_operations {
 	 */
 	int (*bmap) (const char *, size_t blocksize, uint64_t *idx);
 
+#if FUSE_USE_VERSION < 35
```
Nikratio, line 670: What's the rationale for this change? It doesn't seem related to Doxygen to me...

include/fuse.h
```diff
@@ -984,6 +987,10 @@ int fuse_loop(struct fuse *f);
  */
 void fuse_exit(struct fuse *f);
 
+#if FUSE_USE_VERSION < 32
```
Nikratio, line 990: Same question here - why is this needed?

dancerj: the inline diff is a bit weird, but I've moved the comment line just before the function definition.

Before, the comment block was decorating 
int fuse_loop_mt_31(struct fuse *f, int clone_fd);

and there was a warning that `clone_fd` isn't described, and `config` isn't defined.

dancerj:
I realized that I fixed the warnings but some descriptions disappeared. I figured out what I did wrong and here's a pull request to fix it: https://github.com/libfuse/libfuse/pull/608

PR #606, opened 2021-05-13
Add CMake support - plus !

Smit-tay:
This is a rather substantial change which adds CMake build support.
Almost all of the changes are additional CMakeLists.txt files to support that.
It also adds a Jenkins file which can be used in place of Travis.
There are minor modifications to a very few source code files to eliminate warnings.

I supply this in the hopes that someone might find it useful.

You mileage may vary !


Nikratio:
Thanks for submitting this!

The drawback of merging this is that it will incur a continuous maintenance overhead - now everyone who makes changes that affect the build needs to be familiar with two different build systems.

Unless there are good reasons for supporting both, I'd therefore leave this open as a pull request, so that interested people can find it and apply it on their own.

Smit-tay:
Understood.

I will do my best to maintain my fork (Smit-tay/libfuse-cmake) - keeping it
current with upstream libfuse/libfuse at least a few times a year.

I'm wondering about adding some kind of link in the readme.md to direct
CMake users to my fork, though.  That might be a good strategy.

wdlkmpx:
I often cross compile static apps, and just hit the Meson wall

A "low level" C llb using Meson ... it just feels wrong

eli-schwartz:
Can you elaborate on what feels wrong about it? :)

Smit-tay:
It shouldn't make any difference if you use CMake or Meson.
That being said, I have never used Meson for cross compiling.

Nikratio:
I'll close this pull request per discussion above. Please feel free to submit a PR just for README to point at the separately maintained branch.

PR #608, opened 2021-05-17
Define FUSE_USE_VERSION in Doxygen.

dancerj:
We currently do not pass anything in PREDEFINED and that means
FUSE_USE_VERSION is undefined.

Add that definition so that Doxygen built-in C pre-processor can use
FUSE_USE_VERSION value to find the correct comment to parse.

tested with doxygen doc/Doxyfile and now ioctl definitions appear.

### DIFF
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -990,6 +990,11 @@ struct fuse_lowlevel_ops {
 	void (*bmap) (fuse_req_t req, fuse_ino_t ino, size_t blocksize,
 		      uint64_t idx);
 
+#if FUSE_USE_VERSION < 35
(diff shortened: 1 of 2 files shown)


include/fuse_lowlevel.h
```diff
@@ -990,6 +990,11 @@ struct fuse_lowlevel_ops {
 	void (*bmap) (fuse_req_t req, fuse_ino_t ino, size_t blocksize,
 		      uint64_t idx);
 
+#if FUSE_USE_VERSION < 35
```
dancerj, line 993: I've only updated the non-lowlevel version in https://github.com/libfuse/libfuse/pull/600 but I think lowlevel version should be updated or the doc is pointing to the old signature.

PR #610, opened 2021-05-31
fuse3: revert "fuse_daemonize(): chdir to '/' even if not running in the background"

lixiaokeng:
The nvme-cuse in SPDK 21.01 uses cuse in foreground mode. If
fuse_daemonize function calls chdir to '/', then all functions
call with relative-path parameter will report error due to
the working directory changed. The bugfix in SPDK can't fix
the all sence, like: apps call chdir self or multi-thread
race condition etc. Prefer revert patch in fuse3 to fix the
bug in cuse foreground mode.

Signed-off-by: Lixiaokeng<lixiaokeng@huawei.com>
Signed-off-by: Weifeng Su<suweifeng1@huawei.com>

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -270,8 +270,6 @@ int fuse_daemonize(int foreground)
 		(void) write(waiter[1], &completed, sizeof(completed));
 		close(waiter[0]);
 		close(waiter[1]);
-	} else {
-		(void) chdir("/");
 	}
 	return 0;
 }


Nikratio:
Thanks for the report! Could you elaborate on why this is not a problem when running in daemon mode?

lixiaokeng:
> Thanks for the report! Could you elaborate on why this is not a problem when running in daemon mode?

 We can't make sure there is no problem in daemon mode. We find this problem in foreground mode and don't know the impact of deleting chdir in daemon mode, so just revert c52c3256.

Nikratio:
I think it makes more sense for a program that is intended to run in the background to not make assumptions about the current path. If you truly need a specific working directory, why not chdir into it after calling fuse_daemonize?

lixiaokeng:
We have tried it. However, we can't lock all chdir() when it is multi-thread , because we just provide a link library. I agree that  “a program run in the background should not make assumptions about the current path”.  But the cuse always runs in foreground mode, so some functions are called with relative path.  The commit (c52c325) that we want to  revert just changes the directory in foreground mode.

Nikratio:
Sorry, I still do not understand. What's preventing you from running `chdir` after `fuse_daemonize`, but before starting any threads? What's preventing you from not using relative paths at all by prepending your desired "working directory" to any relative path that you encounter?

lixiaokeng:
The cuse_thread will created from time to time in process running. If we want to  solve the problem, we should lock every chdir() and position of using relative path. As I said before， we provide a library to users  and we can't prevent the users from using relative path which was valid in past.  
Here is another question. Does the commit (c52c325) fix some bug? I don't known what effect there will be without it.

Nikratio:
I still do not understand why you cannot run `chdir` after `fuse_daemonize`. Any threads created afterwards will see the effect of this and can use relative paths.

PR #612, opened 2021-06-09
Test reused inodes

amir73il:
To test this kernel fix https://lore.kernel.org/linux-fsdevel/20210609181158.479781-1-amir73il@gmail.com/

This test passes on all the example passthrough filesystems in libfuse without the kernel fix,
because none of these fs can reuse inode numbers of non evicted inodes.

Specifically, all the ll passthrough examples keep open fd's for all live inodes in real fs,
so real fs cannot reuse those inode numbers.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -123,6 +123,8 @@ struct Inode {
     int fd {-1};
     dev_t src_dev {0};
     ino_t src_ino {0};
+    int generation {0};
+    uint64_t nopen {0};
     uint64_t nlookup {0};
     std::mutex m;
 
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks for the patch! Could you provide a more detailed description of what the patch does? Looking at the diff, this mostly looks like a re-factoring to me (passing around `struct stat` instead of `char path`). 

Also, if the test is guaranteed to always succeed, what exactly are the benefits of making the change?

Might it be possible to change e.g. `examples/hello_ll.c` to trigger the behavior that we want to test?

amir73il:
Re: what the patch does?
Isn't the description in commit message of the top commit clear about the WHAT?
As to the WHY, that is explained in the (not so easy to understand) commit message of the linked kernel patch.
In a nut shell, test keeps open O_PATH fd for some files and then tries to use those fd to stat those files again at the end of the test.
As the commit message explains, getting an error on those fstat() calls is perfectly valid, but observing the wrong file via that O_PATH fd is very much wrong and can lead to serious security breach.

RE: change an example to trigger the test failure
Hard question.
Short answer:
Maybe I change passthrough_hp to close the unlinked inode.fd on
unlink() and it would not break other tests, but fail this one. I will try.

Long answer:
I have a branch in my github with an example fs that fails the test (cachegwfs-tests).
The example is based on passthrough_hp, but it is not a trivial change and
it only works over specific src fs (xfs and ext4).
I can contribute a simpler version of this fs to examples, but I have bigger plans -
I plan to factor out and contribute libfuse passthrough library from my fs to help
consolidate the development efforts on different passthrough fs. 
It's still in very early stages, but you can see the general idea on my fuse_passthrough branch. 

Bottom line:
There is no rush to merge this test.
I posted it mostly as a reference test for the kernel fix.
I will see about making the test fail on passthrough_hp

amir73il:
All right. I managed to change passthrough_hp in a way that it:
1. Passes all upstream libfusse tests
2. Fails the modified test_syscalls in this PR on upstream kernel
3. Passes  the modified test_syscalls in this PR on patched kernel

PR #613, opened 2021-06-16
Fix: a potential crash on failure to setlocale

lixiaokeng:
setlocale() can fail, returning NULL, which will lead
to a crash in iconv_new(). Fix it like in iconv_help().

Signed-off-by: Lixiaokeng <lixiaokeng@huawei.com>


lib/modules/iconv.c
```diff
@@ -672,7 +672,7 @@ static struct fuse_fs *iconv_new(struct fuse_args *args,
 {
 	struct fuse_fs *fs;
 	struct iconv *ic;
-	char *old = NULL;
```
Nikratio, line 675: Why remove the `const`?

PR #614, opened 2021-06-16
Fix: fd and memory leak in mount.fuse.c

lixiaokeng:
The command isn't freed and the fuse_fd isn't
closed if execl failed. Fix it.

Signed-off-by: Lixiaokeng <lixiaokeng@huawei.com>

### DIFF
--- a/util/mount.fuse.c
+++ b/util/mount.fuse.c
@@ -243,6 +243,7 @@ int main(int argc, char *argv[])
 	int dev = 1;
 	int suid = 1;
 	int pass_fuse_fd = 0;
+	int fuse_fd = 0;
 	int drop_privileges = 0;
 	char *dev_fd_mountpoint = NULL;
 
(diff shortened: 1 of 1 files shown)


util/mount.fuse.c
```diff
@@ -441,5 +441,9 @@ int main(int argc, char *argv[])
 	execl("/bin/sh", "/bin/sh", "-c", command, NULL);
 	fprintf(stderr, "%s: failed to execute /bin/sh: %s\n", progname,
 		strerror(errno));
+	
+	if (pass_fuse_fd)
+		close(fuse_fd);
```
Nikratio, line 446: As far as I can see, fuse_fd isn't even defined here. Have you tried if this compiles?

Nikratio (changes requested):

util/mount.fuse.c
lixiaokeng: I'm sorry for no compile. Now I move definition of fuse_fd to the begin of main.

PR #615, opened 2021-06-17
passthrough_ll: Don't allow file creation with FUSE_OPEN

gkurz:
A well behaved FUSE client uses FUSE_CREATE to create files. It isn't
supposed to pass O_CREAT along a FUSE_OPEN request, as documented in
the "fuse_lowlevel.h" header :

    /**
     * Open a file
     *
     * Open flags are available in fi->flags. The following rules
     * apply.
     *
     *  - Creation (O_CREAT, O_EXCL, O_NOCTTY) flags will be
     *    filtered out / handled by the kernel.

But if the client happens to do it anyway, the server ends up passing
this flag to open() without the mandatory mode_t 4th argument. Since
open() is a variadic function, glibc will happily pass whatever it
finds on the stack to the syscall. If this file is compiled with
-D_FORTIFY_SOURCE=2, glibc will even detect that and abort:

*** invalid openat64 call: O_CREAT or O_TMPFILE without mode ***: terminated

Specifying O_CREAT with FUSE_OPEN is a protocol violation. Check this
in lo_open() and return an error to the client : EINVAL since this is
already what glibc returns with other illegal flag combinations.

The FUSE filesystem doesn't currently support O_TMPFILE, but the very
same would happen if O_TMPFILE was passed in a FUSE_OPEN request. Check
that as well.

Signed-off-by: Greg Kurz <groug@kaod.org>

### DIFF
--- a/example/passthrough_ll.c
+++ b/example/passthrough_ll.c
@@ -790,6 +790,12 @@ static void lo_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
 		fuse_log(FUSE_LOG_DEBUG, "lo_open(ino=%" PRIu64 ", flags=%d)\n",
 			ino, fi->flags);
 
+	/* File creation is handled by lo_create() */
+	if (fi->flags & (O_CREAT | O_TMPFILE)) {
+		fuse_reply_err(req, EINVAL);
+		return;
+	}
+
 	/* With writeback cache, kernel may send read requests even
 	   when userspace opened write-only */
 	if (lo->writeback && (fi->flags & O_ACCMODE) == O_WRONLY) {


gkurz:
Travis CI is failing on [`test/test_examples.py::test_passthrough_hp[False] FAILED`](https://travis-ci.com/github/libfuse/libfuse/builds/229592337#L650). The test name seems to indicate it is for passthrough_hp while my patch is for passthrough_ll... so I guess something else is breaking the test.

Nikratio:
Thanks for the patch!

The "FUSE Client" in this context is the Linux kernel, right? 

If we return EINVAL, then this signals to the userspace process who has called `open` that it did something wrong - which isn't the case. So I think this is sub-optimal. I agree that it makes sense to check for this, but I think we should be handling it differently. At the very least print an explanatory message, and maybe just terminate the process instead of returning an error.

Have you checked if `passthrough_hp` may have the same problem?

Wouldn't such a check be placed better in FUSE itself rather than in one of the examples? Or is there a situation where a FUSE filesystem would want to accept this flag in its open handler?

gkurz:
> Thanks for the patch!
> 
> The "FUSE Client" in this context is the Linux kernel, right?
> 

Yes.

> If we return EINVAL, then this signals to the userspace process who has called `open` that it did something wrong - which isn't the case. 

You're right that userspace didn't do anything wrong. The culprit is the linux kernel : O_CREAT should be converted to a FUSE_CREATE or FUSE_MKNOD request but not FUSE_OPEN.

> So I think this is sub-optimal. I agree that it makes sense to check for this, but I think we should be handling it differently. At the very least print an explanatory message,  and maybe just terminate the process instead of returning an error.
> 

I have mixed feelings about this. This patch was originaly made for QEMU's virtiofsd passthough_ll.c. In our case, the kernel FUSE driver is untrusted code and we need to harden virtiofsd against buggy guests. Replacing an abort() by an exit() doesn't make much sense in this respect. Another solution could be to ignore O_CREAT (i.e. flags &= ~O_CREAT) when processing a FUSE_OPEN request : ENOENT would be returned if the file doesn't exist, which is definitely something that userspace should
be ready to handle when calling open(), no matter O_CREAT is specified or not.

I guess printing an error message is fine if it doesn't allow the guest to flood some log files on the host.

> Have you checked if `passthrough_hp` may have the same problem?
> 

No.

> Wouldn't such a check be placed better in FUSE itself rather than in one of the examples? Or is there a situation where a FUSE filesystem would want to accept this flag in its open handler?

Since we already have FUSE_MKNOD and FUSE_CREATE to handle open(O_CREAT), I don't see
a case where we'd also handle this with FUSE_OPEN.

So, yes, this should probably be handled in fuse_lowlevel.c:do_create() rather than lo_create().

Nikratio:
Are you still interested in working on this? If not, I'll close the pull request for now.

gkurz:
> Are you still interested in working on this? If not, I'll close the pull request for now.

I've switched to another project and I won't be able to go ahead with this. Feel free to close this PR.

PR #616, opened 2021-06-22
Do not run unlinked files test on passthrough_hp with old kernels

amir73il:
test_syscalls unlinked testfiles check fails on passthourhg_hp without
the kernel commit "fuse: fix illegal access to inode with reused nodeid"
queued for kernel 5.14 [1].

Make this check opt-in and do not run it with kernel version before 5.14.

[1] https://lore.kernel.org/linux-fsdevel/CAJfpegtGKjeK8E5QsHKF0=re1J9wVHuGRVgg9NDJ_OOwQdCUNg@mail.gmail.com/

Signed-off-by: Amir Goldstein <amir73il@gmail.com>

### DIFF
--- a/test/test_examples.py
+++ b/test/test_examples.py
@@ -17,6 +17,8 @@ import tempfile
 import time
 import errno
 import sys
+import platform
+from distutils.version import LooseVersion
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks!

PR #619, opened 2021-07-24
util/ulockmgr_server.c: conditionally define closefrom (fix glibc-2.34+)

thesamesam:
util/ulockmgr_server.c: conditionally define closefrom (fix glibc-2.34+)

closefrom(3) has joined us in glibc-land from *BSD and Solaris. Since
it's available in glibc 2.34+, we want to detect it and only define our
fallback if the libc doesn't provide it.

Bug: https://bugs.gentoo.org/803923
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,6 +55,7 @@ fi
 
 AC_CHECK_FUNCS([fork setxattr fdatasync splice vmsplice utimensat])
 AC_CHECK_FUNCS([posix_fallocate])
+AC_CHECK_FUNCS([closefrom])
 AC_CHECK_MEMBERS([struct stat.st_atim])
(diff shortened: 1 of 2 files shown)


roastduck:
Hi, there. It seems this fix has been merged to `fuse_2_9_bugfix` for 3 years, but there is no more 2.9 release since then. Should we make a new 2.9 release including this fix?

bsbernd:
I think we should fork that branch into a new ulockmgr repo and then remove all unrelated files. I.e. that would leave history, but would also make ulockmgr independent of the libfuse 2.9 branch.
For our internal ulockmgr version I had also fixed several severe bugs (list handling issues, as far as I remember) several years back. Just hard to push the changes upstream in the current state.

PR #623, opened 2021-08-20
Minor cleanup

a1346054:


Nikratio:
Thank you! The problem with white-space fixes is that they make `git blame` much harder to use and there is nothing that prevents trailing whitespace from being introduced again in the future. So I would prefer if you could revert those.

a1346054:
I added the .editorconfig file, to make conforming editors remove trailing whitespace by default, so that it doesn't come back.

Trailing whitespace makes a lot of editors give annoying warnings.

Nikratio:
As for the SSHFS changes, I don't think the .editorconfig file will be sufficient to ensure that this does not recur. There's too many editors that are not using it (including Emacs).

Can we split this into one pull request for whitespace + ignore file + CI changes, and a second one for the other changes?

a1346054:
Split those commits into #624 where the whitespace discussion can happen.

PR #630, opened 2021-11-07
WIP: Fix dlsym usage

some00:
"fuse_module_<name>_factory" is a function pointer but is loaded like a
function, which results an crash for any external module.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -252,7 +252,7 @@ static int fuse_load_so_module(const char *module)
 	int ret = -1;
 	char *tmp;
 	struct fusemod_so *so;
-	fuse_module_factory_t factory;
+	fuse_module_factory_t* factory;
 
 	tmp = malloc(strlen(module) + 64);
 	if (!tmp) {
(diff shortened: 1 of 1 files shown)


lib/fuse.c
```diff
@@ -280,7 +280,11 @@ static int fuse_load_so_module(const char *module)
 			tmp, dlerror());
 		goto out_dlclose;
 	}
-	ret = fuse_register_module(module, factory, so);
+	if (*factory == NULL) {
```
Nikratio, line 283: How does this check differ from the one in line 278?

Nikratio, line 283: Isn't this dangerous to do if tmp == NULL? factory has not been initialized at this point.

some00: If tmp == NULL the functions returns and this line is not reached.
For some reason github displays the unchanged lines in PRs with an extra indentation in this repository.

some00:
Changed the title of the PR to WIP

1. There is a further issue at cleanup when using modules. I found a solution for it.
2. I would like to extend it with a test case.

Nikratio:
Closing this for now, please re-open when ready for review.

PR #631, opened 2021-11-17
test/test_syscalls.c: clean-up unlinked files from fds list

luis-henrix:
Test test/test_examples.py::test_passthrough_hp[False] was failing due to lack
of file descriptors clean-up when the test file is unlinked.  This would result
in errors such as:

  3 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
  4 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
  5 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
  9 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
  ...

This patch simply replaces all the unlink() calls to files created with
create_testfile() by a new helper unlink_testfile() that will also set the ->fd
to -1.

Signed-off-by: Luís Henriques <lhenriques@suse.de>

### DIFF
--- a/test/test_syscalls.c
+++ b/test/test_syscalls.c
@@ -277,7 +277,8 @@ static int fcheck_stat(int fd, int flags, struct stat *st)
 		if (flags & O_PATH) {
 			// With O_PATH fd, the server does not have to keep
 			// the inode alive so FUSE inode may be stale or bad
-			if (errno == ESTALE || errno == EIO || errno == ENOENT)
+			if (errno == ESTALE || errno == EIO ||
+			    errno == ENOENT || errno == EBADF)
 				return 0;
 		}
 		PERROR("fstat");


amir73il:
This completely breaks check_unlinked_testfile()
The only purpose of the ->fd field is to keep an open O_PATH
fd for unlinked files.

Need to figure out why getting EBADF for those fds that
should be valid O_PATH fds.

luis-henrix:
> This completely breaks check_unlinked_testfile() The only purpose of the ->fd field is to keep an open O_PATH fd for unlinked files.
> 
> Need to figure out why getting EBADF for those fds that should be valid O_PATH fds.

Ok, I may have something.  Here's the sfs_unlink() (from example/passthrough_hp.cc):
```
                    if (inode.fd > 0 && !inode.nopen) {
                            if (fs.debug)
                                    cerr << "DEBUG: unlink: release inode " << e.attr.st_ino
                                            << "; fd=" << inode.fd << endl;
                            lock_guard<mutex> g_fs {fs.mutex};
                            close(inode.fd);
                            inode.fd = -ENOENT;
                            inode.generation++;
                    }
```
This `close(inode.fd)` will effectively cause the test to fail; commenting it (and the next statement) makes the failure go away.  But to be honest I'm not really sure that's the correct way to fix.  Could we simply remove this whole `if` statement?  @amir73il can you advise please?

amir73il:
@luis-henrix the code in sfs_unlink() is intentional and correct.
The problem is with the code in fcheck_stat().
fstat() should not be used to check an O_PATH fd.
The EBADF error is the result of that misuse. 

Instead, for O_PATH fd , the function should use fstatat(fd, "", &stbuf, AT_EMPTY_PATH)
I am not sure how the test worked when I tested. I cannot check it right now, but thought I'd give you the suggested fix anyway.

@Nikratio I guess none of the CI machines run with kernel >= 5.14 where this test is run?
I remember we discussed this but don't remember the answer.

luis-henrix:
> @luis-henrix the code in sfs_unlink() is intentional and correct. The problem is with the code in fcheck_stat(). fstat() should not be used to check an O_PATH fd. The EBADF error is the result of that misuse.
> 
> Instead, for O_PATH fd , the function should use fstatat(fd, "", &stbuf, AT_EMPTY_PATH) I am not sure how the test worked when I tested. I cannot check it right now, but thought I'd give you the suggested fix anyway.

Ok, to be honest I'm not sure why using fstatat() instead of fstat() should make a difference.  According to the man page, they both should be OK with an O_PATH file descriptor.  But the truth is that I still see the same error using fstatat().

But I still think the issue is on the example/passthrough_hp.cc code.  When the test unlinks a file, sfs_unlink() is called and runs the above code snippet (which does `inode.fd = -ENOENT`).  Later on, when the test does the fstat() in fcheck_stat(), function sfs_getattr() will eventually get executed and will execute fstatat() with an invalid (-ENOENT) fd.

amir73il:
@luis-henrix yes you are right. fstat() should work.
close(inode.fd) is essential to testing inode number reuse, but leaving an invalid inode.fd is not good.
I think a possible solution is to do:
`inode.fd = open("/dev/null", O_PATH);` in sfs_unlink after closing inode.fd.
If this works for you, please verify that the test fails on kernel < 5.14 (before "fuse: fix illegal access to inode with reused nodeid"). You will need to remove the test requirement LooseVersion(platform.release()) >= '5.14' in test_examples.py.
The expectation of a good test result is that fstat() returns with ESTALE/ENONET because the fuse inode held by the O_PATH fd is no longer valid and that fstat() will NOT return the attributes of "/dev/null".

luis-henrix:
>If this works for you, please verify that the test fails on kernel < 5.14 (before "fuse: fix illegal access to inode with reused nodeid"). You will need to remove the test requirement LooseVersion(platform.release()) >= '5.14' in test_examples.py. The expectation of a good test result is that fstat() returns with ESTALE/ENONET because the fuse inode held by the O_PATH fd is no longer valid and that fstat() will NOT return the attributes of "/dev/null".

So, I've tested the code with
```diff
@@ -525,7 +525,7 @@ static void sfs_unlink(fuse_req_t req, fuse_ino_t parent, const char *name) {
                                            << "; fd=" << inode.fd << endl;
                            lock_guard<mutex> g_fs {fs.mutex};
                            close(inode.fd);
-                           inode.fd = -ENOENT;
+                           inode.fd = open("/dev/null", O_PATH);
                            inode.generation++;
                    }
            }
```
and
```diff
@@ -245,8 +245,8 @@ def test_passthrough_hp(short_tmpdir, cache, output_checker):
             # unlinked testfiles check fails without kernel fix
             # "fuse: fix illegal access to inode with reused nodeid"
             # so opt-in for this test from kernel 5.14
-            if LooseVersion(platform.release()) >= '5.14':
-                syscall_test_cmd.append('-u')
+            #if LooseVersion(platform.release()) >= '5.14':
+            syscall_test_cmd.append('-u')
             subprocess.check_call(syscall_test_cmd)
     except:
         cleanup(mount_process, mnt_dir)
```
The test passes both with a v5.13 (without commit 15db16837a35 ("fuse: fix illegal access to inode with reused nodeid")) and with a recent kernel (5.16-rc1).  The reason seems to be that fcheck_stat() masks ESTALE, EIO and ENOENT (I checked that errno is set to EIO with both kernels) .

Looking deeper, I found that the reason for getting EIO there is because, when doing that fstat(), the kernel function fuse_do_getattr() will do:
```
                if (fuse_invalid_attr(&outarg.attr) ||
                    inode_wrong_type(inode, outarg.attr.mode)) {
                        fuse_make_bad(inode);
                        err = -EIO;
                } else {
```
due to wrong inode mode.

Sooo... changing the inode.fd to /dev/null isn't probably a good idea.

amir73il:
How about trying with some regular file like "/proc/version"?
Thanks for the accurate analysis. Sorry I cannot test this right now.

luis-henrix:
> How about trying with some regular file like "/proc/version"?

No, that won't work either because fcheck_stat() will then call st_check_type(), st_check_mode(), st_check_size() and st_check_nlink(), and they will all fail.

**However** it still doesn't seem to fail on kernel 5.13 when I comment out these extra checks in fcheck_stat().  I can't see a reason for it not failing, as you mentioned that it should without commit 15db16837a35.  But I'll have to leave that for tomorrow.

amir73il:
@luis-henrix the whole point of the test is for st_check_type() and friends to fail on 5.13, because the kernel driver did not recognize the generation change of the inode.
The kernel bug fix should result in EIO/STALE error from fstat(), because inode_wrong_type() was replaced with fuse_stale_inode() that will make the test pass, because it will not get to st_check_type() and friends.
You see, passthrough_hp is emulating a reused inode number (nodeid) with different generation, which is a situation that can happen and kernel needs to deal with it correctly.

luis-henrix:
Thanks @amir73il I think I understand the point of the st_check_*() functions and that they rightfully fail for 5.13.

Unfortunately, the test also fails with a 5.16 kernel in these checks (when using "/proc/version" of the inode.fd):
```
...
  3 [check_unlinked_testfile] st_check_mode() - mode 0444 instead of 0644             
  3 [check_unlinked_testfile] st_check_size() - length 0 instead of 26                
  3 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0                 
  4 [check_unlinked_testfile] st_check_mode() - mode 0444 instead of 0644             
  4 [check_unlinked_testfile] st_check_size() - length 0 instead of 26                
  4 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0                 
  5 [check_unlinked_testfile] st_check_mode() - mode 0444 instead of 0644             
...
```
I believe this is because fuse_do_getattr() is not returning -EIO.  This kernel function should probably use fuse_stale_inode() instead of inode_wrong_type() to also check the inode generation.  However, from what I could see, there is no easy way for doing that without changing the fuse protocol to also include the generation in struct fuse_attr_out.

amir73il:
Ok @luis-henrix I think I understand what's going on. I think all you need to fix the test is add EBADF to accepted errors for passing the test.
I guess when I tested I tested over ext4 or xfs and inodes got reused.
I am guessing you are testing over btrfs where inode numbers do not get reused.
The test will not verify the kernel fix over btrfs but at least it will not fail.
Are my assumptions correct? Can you test over xfs or ext4?

luis-henrix:
> Are my assumptions correct? Can you test over xfs or ext4?

No, I am not using btrfs.  My workstation runs on ext4 and I was using `virtme` to run the tests (to easily pick the kernel versions I wanted).  But to double-check I've just tried to run the test on bare-metal (running the distro 5.14.14-based kernel).  Same result as above.

For the record, I'm using the follow to run the test:
```
python3 -m pytest test/test_examples.py::test_passthrough_hp[False]
```

amir73il:
Ok so to sum it up. I think it is fine to fix the test by accepting EBADF as correct outcome.
I think that xfs ext4 are supposed to get success from fstat() for some of the unlinked files some of the times and on old kernel and then fail the other checks but the nature of these failures is random so if you do not observe those failures I will investigate it later.
For new kernel fstat() should always fail for some error, so the test should pass.

amir73il (approved):

amir73il:
I can confirm that on my system with kernel v5.9.6 and underlying ext4 I can see the EBADF:
` 14 [check_unlinked_testfile] st_check_size() - length 45 instead of 0
 14 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 15 [check_unlinked_testfile] st_check_size() - length 45 instead of 0
 15 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 16 [check_unlinked_testfile] st_check_size() - length 45 instead of 13
 16 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 17 [check_unlinked_testfile] st_check_size() - length 45 instead of 26
 17 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 18 [check_unlinked_testfile] st_check_size() - length 45 instead of 126
 18 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 19 [check_unlinked_testfile] st_check_mode() - mode 0644 instead of 0600
 19 [check_unlinked_testfile] st_check_size() - length 45 instead of 0
 19 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 20 [check_unlinked_testfile] st_check_mode() - mode 0644 instead of 0600
 20 [check_unlinked_testfile] st_check_size() - length 45 instead of 13
 20 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 21 [check_unlinked_testfile] st_check_mode() - mode 0644 instead of 0600
 21 [check_unlinked_testfile] st_check_size() - length 45 instead of 26
 21 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 22 [check_unlinked_testfile] st_check_mode() - mode 0644 instead of 0600
 22 [check_unlinked_testfile] st_check_size() - length 45 instead of 126
 22 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 23 [check_unlinked_testfile] st_check_mode() - mode 0644 instead of 0400
 23 [check_unlinked_testfile] st_check_size() - length 45 instead of 0
 23 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 24 [check_unlinked_testfile] st_check_mode() - mode 0644 instead of 0200
 24 [check_unlinked_testfile] st_check_size() - length 45 instead of 0
 24 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 25 [check_unlinked_testfile] st_check_mode() - mode 0644 instead of 00
 25 [check_unlinked_testfile] st_check_size() - length 45 instead of 0
 25 [check_unlinked_testfile] st_check_nlink() - nlink 1 instead of 0
 49 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
 50 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
 51 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
 52 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
 53 [check_unlinked_testfile] fcheck_stat() - fstat: Bad file descriptor
36 unlinked testfile checks failed
FAILED
`
 and that the test still fails as expected after whitelisting EBADF.

PR #635, opened 2022-01-03
Add support for FOPEN_NOFLUSH flag

amir73il:
The support for the flag was added by commit a390ccb316be ("fuse: add FOPEN_NOFLUSH") to kernel v5.16-rc1

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,8 @@
+Unreleased Changes
+==================
+
+* Add support for flag FOPEN_NOFLUSH for avoiding flush on close.
+
 libfuse 3.10.5 (2021-09-06)
 ===========================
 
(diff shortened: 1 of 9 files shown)


Nikratio:
Thanks, looks good! Please just document the new flag in ChangeLog.rst.

Nikratio:
This seems to break a unit test...?

amir73il:
Interesting.
The test was not supposed to change logically for writeback=True, but there are those mutex locks that may have changed timing. I will remove them from the writeback=True case.

However, the assertion that failed (ERROR: Expected 8192 bytes, got 4096) is not really guaranteed with writeback cache enabled - it is only very probable...

Nikratio:
The idea with the assertion is probably to test that writes are indeed cached without relying on timing. So maybe it is sufficient to reduce the size of userspace writes and/or change the assertion to not-equal to userspace write size...

aakefbs:
It is there a reason this is not merged yet?

aakefbs:
Thanks for merging @Nikratio !

Issue #640, opened 2022-02-06
Returning an error in ioctl(2) with the high level of fuse


jpandre:
An unexpected behavior has been reported when using lsattr(1) on ntfs-3g (see https://github.com/tuxera/ntfs-3g/issues/28).

lsattr(1) is not supported by ntfs-3g. An ioctl(2) is received to which ntfs-3g returns with -EINVAL to the high level interface (or with fuse_reply_err() in the low level interface). EOPNOTSUPP might be more appropriate, but this is not a valid return to ioctl(2).

When using the low interface, the error is properly propagated :

```
unique: 18, opcode: OPEN (14), nodeid: 87, insize: 48, pid: 5081
   unique: 18, success, outsize: 32
unique: 20, opcode: IOCTL (39), nodeid: 87, insize: 72, pid: 5081
   unique: 20, error: -22 (Invalid argument), outsize: 16
```

But when using the high level, the error is not propagated :

```
unique: 20, opcode: IOCTL (39), nodeid: 3, insize: 72, pid: 5105
ioctl[0] 0x80086601 flags: 0x0
   unique: 20, success, outsize: 36
```

Now, in fuse.c, in the function fuse_lib_ioctl(), a negative returned value from the file system is just sent to fuse_reply_ioctl() instead of using reply_err(). As a consequence the error is not visible and causes some confusion to the caller.

Is this intended ? should not a negative return from the file system imply reply_err() ?

If the issue is confirmed, I will do a pull request.

Tuupertunut:
Just a note that this bug breaks Wine (the Windows compatibility layer).

The Windows-style file discovery in Wine relies on reading directory iflags with `ioctl`. On fuse filesystems with broken version of libfuse, `ioctl` always succeeds and returns random data as iflags.

I highly recommend that anyone who uses Wine updates to libfuse 3.11 which fixes this issue. It's a pity that Ubuntu 22.04 LTS has settled on using libfuse 3.10 and they don't update often.

rafalh:
In Ubuntu 22.04 LTS it is probably easier and better to use `ntfs3` driver that was added in Linux kernel 5.15 and does not use fuse at all.
But still it's a pity a fix for this issue was not included :( I wonder if it has any implications for security...

PR #641, opened 2022-02-09
Fixed returning an error condition to ioctl(2)

jpandre:
When returning a negative error code by ->ioctl() to the high level interface, no error is propagated to the low level, and the reply message to the kernel is shown as successful.

A negative result is however returned to kernel, so the kernel can detect the bad condition, but this appears to not be the case since kernel 5.15.

The proposed fix is more in line with the usual processing of errors in fuse, taking into account that ioctl(2) always returns a non-negative value in the absence of errors.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,5 @@
+* Fixed returning an error condition to ioctl(2)
+
 libfuse 3.10.5 (2021-09-06)
 ===========================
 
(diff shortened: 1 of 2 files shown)


jpandre:
See the context of this proposal in
https://github.com/tuxera/ntfs-3g/issues/28
and
https://github.com/libfuse/libfuse/issues/640

ollpu:
Should this be backported to the 2.9 branch?

PR #652, opened 2022-03-28
fuse-loop/fuse_do_work: Avoid lots of thread creations/destructions

aakefbs:
On benchmarking metadata operations with a single threaded bonnie++
and "max_idle_threads" limited to 1, 'top' was showing suspicious
160% cpu usage.
Profiling the system with flame graphs showed that an astonishing
amount of CPU time was spent in thread creation and destruction.

After verifying the code it turned out that fuse_do_work() was
creating a new thread every time all existing idle threads
were already busy. And then just a few lines later after processing
the current request it noticed that it had created too many threads
and destructed the current thread. I.e. there was a thread
creation/destruction ping-pong.

Code is changed to only create new threads if the max number of
threads is not reached.

Furthermore, thread destruction is disabled, as creation/destruction
is expensive in general.

With this change cpu usage of passthrough_hp went from ~160% to
~80% (with different values of max_idle_threads). And bonnie
values got approximately faster by 90%. This is a with single
threaded bonnie++
bonnie++ -x 4 -q -s0  -d <path> -n 30:1:1:10 -r 0

Without this patch, using the default max_idle_threads=10 and just
a single bonnie++ the thread creation/destruction code path is not
triggered.  Just one libfuse and one application thread is just
a corner case - the requirement for the issue was just
n-application-threads >= max_idle_threads.

Signed-off-by: Bernd Schubert <bschubert@ddn.com>

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -43,7 +43,7 @@
  * \include passthrough_hp.cc
  */
 
-#define FUSE_USE_VERSION 35
+#define FUSE_USE_VERSION FUSE_MAKE_VERSION(3, 12)
 
 #ifdef HAVE_CONFIG_H
 #include "config.h"
@@ -1228,6 +1228,8 @@ static void maximize_fd_limit() {
 
 int main(int argc, char *argv[]) {
 
+    struct fuse_loop_config *loop_config = NULL;
+
     // Parse command line options
     auto options {parse_options(argc, argv)};
 
@@ -1274,15 +1276,15 @@ int main(int argc, char *argv[]) {
     umask(0);
 
     // Mount and run main loop
-    struct fuse_loop_config loop_config;
-    loop_config.clone_fd = 0;
-    loop_config.max_idle_threads = 10;
+    loop_config = fuse_loop_cfg_create();
+
     if (fuse_session_mount(se, argv[2]) != 0)
         goto err_out3;
     if (options.count("single"))
         ret = fuse_session_loop(se);
     else
-        ret = fuse_session_loop_mt(se, &loop_config);
+        ret = fuse_session_loop_mt(se, loop_config);
+
 
     fuse_session_unmount(se);
 
@@ -1291,6 +1293,8 @@ err_out3:
 err_out2:
     fuse_session_destroy(se);
 err_out1:
+
+    fuse_loop_cfg_destroy(loop_config);
     fuse_opt_free_args(&args);
 
     return ret ? 1 : 0;
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -1041,7 +1041,7 @@ int fuse_loop_mt_32(struct fuse *f, struct fuse_loop_config *config);
  * in the callback function of fuse_operations is also thread-safe.
  *
  * @param f the FUSE handle
- * @param config loop configuration
+ * @param config loop configuration, may be NULL and defaults will be used then
  * @return see fuse_session_loop()
  *
  * See also: fuse_loop()
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -1011,6 +1011,9 @@ void fuse_exit(struct fuse *f);
 #if FUSE_USE_VERSION < 32
 int fuse_loop_mt_31(struct fuse *f, int clone_fd);
 #define fuse_loop_mt(f, clone_fd) fuse_loop_mt_31(f, clone_fd)
+#elif FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
+int fuse_loop_mt_32(struct fuse *f, struct fuse_loop_config *config);
+#define fuse_loop_mt(f, config) fuse_loop_mt_32(f, config)
 #else
 /**
  * FUSE event loop with multiple threads
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -857,13 +857,19 @@ struct fuse_loop_config *fuse_loop_cfg_create(void);
 void fuse_loop_cfg_destroy(struct fuse_loop_config *config);
 
 /**
- * fuse_loop_config2 setter to set the number of max idle threads.
+ * fuse_loop_config setter to set the number of max idle threads.
  */
 void fuse_loop_cfg_set_idle_threads(struct fuse_loop_config *config,
 				    unsigned int value);
 
 /**
- * fuse_loop_config2 setter to enable the clone_fd feature
+ * fuse_loop_config setter to set the number of max threads.
+ */
+void fuse_loop_cfg_set_max_threads(struct fuse_loop_config *config,
+				   unsigned int value);
+
+/**
+ * fuse_loop_config setter to enable the clone_fd feature
  */
 void fuse_loop_cfg_set_clone_fd(struct fuse_loop_config *config,
 				unsigned int value);
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -23,7 +23,7 @@
 #define FUSE_MAJOR_VERSION 3
 
 /** Minor version of FUSE library interface */
-#define FUSE_MINOR_VERSION 11
+#define FUSE_MINOR_VERSION 12
 
 #define FUSE_MAKE_VERSION(maj, min)  ((maj) * 100 + (min))
 #define FUSE_VERSION FUSE_MAKE_VERSION(FUSE_MAJOR_VERSION, FUSE_MINOR_VERSION)
@@ -104,11 +104,20 @@ struct fuse_file_info {
 	uint32_t poll_events;
 };
 
+
+
 /**
  * Configuration parameters passed to fuse_session_loop_mt() and
  * fuse_loop_mt().
+ * Deprecated and replaced by a newer private struct in FUSE API
+ * version 312 (FUSE_MAKE_VERSION(3, 12)
  */
+#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
+struct fuse_loop_config_v1; /* forward declarition */
 struct fuse_loop_config {
+#else
+struct fuse_loop_config_v1 {
+#endif
 	/**
 	 * whether to use separate device fds for each thread
 	 * (may increase performance)
@@ -127,6 +136,7 @@ struct fuse_loop_config {
 	 */
 	unsigned int max_idle_threads;
 };
+
 
 /**************************************************************************
  * Capability bits for 'fuse_conn_info.capable' and 'fuse_conn_info.want' *
@@ -833,6 +843,39 @@ int fuse_set_signal_handlers(struct fuse_session *se);
  * fuse_set_signal_handlers()
  */
 void fuse_remove_signal_handlers(struct fuse_session *se);
+
+/**
+ * Create and set default config for fuse_session_loop_mt and fuse_loop_mt.
+ *
+ * @return anonymous config struct
+ */
+struct fuse_loop_config *fuse_loop_cfg_create(void);
+
+/**
+ * Free the config data structure
+ */
+void fuse_loop_cfg_destroy(struct fuse_loop_config *config);
+
+/**
+ * fuse_loop_config2 setter to set the number of max idle threads.
+ */
+void fuse_loop_cfg_set_idle_threads(struct fuse_loop_config *config,
+				    unsigned int value);
+
+/**
+ * fuse_loop_config2 setter to enable the clone_fd feature
+ */
+void fuse_loop_cfg_set_clone_fd(struct fuse_loop_config *config,
+				unsigned int value);
+
+/**
+ * Convert old config to more recernt fuse_loop_config2
+ *
+ * @param config current config2 type
+ * @param v1_conf older config1 type (below FUSE API 312)
+ */
+void fuse_loop_cfg_convert(struct fuse_loop_config *config,
+			   struct fuse_loop_config_v1 *v1_conf);
 
 /* ----------------------------------------------------------- *
  * Compatibility stuff					       *
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -1968,26 +1968,29 @@ int fuse_session_mount(struct fuse_session *se, const char *mountpoint);
 int fuse_session_loop(struct fuse_session *se);
 
 #if FUSE_USE_VERSION < 32
-int fuse_session_loop_mt_31(struct fuse_session *se, int clone_fd);
-#define fuse_session_loop_mt(se, clone_fd) fuse_session_loop_mt_31(se, clone_fd)
+	int fuse_session_loop_mt_31(struct fuse_session *se, int clone_fd);
+	#define fuse_session_loop_mt(se, clone_fd) fuse_session_loop_mt_31(se, clone_fd)
+#elif FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
+	int fuse_session_loop_mt_32(struct fuse_session *se, struct fuse_loop_config *config);
+	#define fuse_session_loop_mt(se, config) fuse_session_loop_mt_32(se, config)
 #else
-#if (!defined(__UCLIBC__) && !defined(__APPLE__))
-/**
- * Enter a multi-threaded event loop.
- *
- * For a description of the return value and the conditions when the
- * event loop exits, refer to the documentation of
- * fuse_session_loop().
- *
- * @param se the session
- * @param config session loop configuration 
- * @return see fuse_session_loop()
- */
-int fuse_session_loop_mt(struct fuse_session *se, struct fuse_loop_config *config);
-#else
-int fuse_session_loop_mt_32(struct fuse_session *se, struct fuse_loop_config *config);
-#define fuse_session_loop_mt(se, config) fuse_session_loop_mt_32(se, config)
-#endif
+	#if (!defined(__UCLIBC__) && !defined(__APPLE__))
+		/**
+		 * Enter a multi-threaded event loop.
+		 *
+		 * For a description of the return value and the conditions when the
+		 * event loop exits, refer to the documentation of
+		 * fuse_session_loop().
+		 *
+		 * @param se the session
+		 * @param config session loop configuration
+		 * @return see fuse_session_loop()
+		 */
+		int fuse_session_loop_mt(struct fuse_session *se, struct fuse_loop_config *config);
+	#else
+		int fuse_session_loop_mt_312(struct fuse_session *se, struct fuse_loop_config *config);
+		#define fuse_session_loop_mt(se, config) fuse_session_loop_mt_312(se, config)
+	#endif
 #endif
 
 /**
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -1868,6 +1868,11 @@ void fuse_cmdline_help(void);
  * Filesystem setup & teardown                                 *
  * ----------------------------------------------------------- */
 
+/**
+ * Note: Any addition to this struct needs to create a compatibility symbol
+ *       for fuse_parse_cmdline(). For ABI compatibility reasons it is also
+ *       not possible to remove struct members.
+ */
 struct fuse_cmdline_opts {
 	int singlethread;
 	int foreground;
@@ -1877,7 +1882,11 @@ struct fuse_cmdline_opts {
 	int show_version;
 	int show_help;
 	int clone_fd;
-	unsigned int max_idle_threads;
+	unsigned int max_idle_threads; /* discouraged, due to thread
+	                                * destruct overhead */
+
+	/* Added in libfuse-3.12 */
+	unsigned int max_threads;
 };
 
 /**
@@ -1898,8 +1907,20 @@ struct fuse_cmdline_opts {
  * @param opts output argument for parsed options
  * @return 0 on success, -1 on failure
  */
+#if (!defined(__UCLIBC__) && !defined(__APPLE__))
 int fuse_parse_cmdline(struct fuse_args *args,
 		       struct fuse_cmdline_opts *opts);
+#else
+#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
+int fuse_parse_cmdline_30(struct fuse_args *args,
+			   struct fuse_cmdline_opts *opts);
+#define fuse_parse_cmdline(args, opts) fuse_parse_cmdline_30(args, opts)
+#else
+int fuse_parse_cmdline_312(struct fuse_args *args,
+			   struct fuse_cmdline_opts *opts);
+#define fuse_parse_cmdline(args, opts) fuse_parse_cmdline_312(args, opts)
+#endif
+#endif
 
 /**
  * Create a low level session.
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -87,6 +87,50 @@ struct fuse_module {
 	int ctr;
 };
 
+/**
+ * Configuration parameters passed to fuse_session_loop_mt() and
+ * fuse_loop_mt().
+ *
+ * Internal API to avoid exposing the plain data structure and
+ * causing compat issues after adding or removing struct members.
+ *
+ */
+#if FUSE_USE_VERSION >= FUSE_MAKE_VERSION(3, 12)
+struct fuse_loop_config
+{
+	/* verififier that a correct struct was was passed. This is especially
+	 * needed, as versions below (3, 12) were using a public struct
+	 * (now called  fuse_loop_config_v1), which was hard to extend with
+	 * additional parameters, without risking that file system implementations
+	 * would not have noticed and might either pass uninitialized members
+	 * or even too small structs.
+	 * fuse_loop_config_v1 has clone_fd at this offset, which should be either 0
+	 * or 1. v2 or even higher version just need to set a value here
+	 * which not conflicting and very unlikely as having been set by
+	 * file system implementation.
+	 */
+	int version_id;
+
+	/**
+	 * whether to use separate device fds for each thread
+	 * (may increase performance)
+	 */
+	int clone_fd;
+	/**
+	 * The maximum number of available worker threads before they
+	 * start to get deleted when they become idle. If not
+	 * specified, the default is 10.
+	 *
+	 * Adjusting this has performance implications; a very small number
+	 * of threads in the pool will cause a lot of thread creation and
+	 * deletion overhead and performance may suffer. When set to 0, a new
+	 * thread will be created to service every operation.
+	 * The special value of -1 means that this parameter is disabled.
+	 */
+	int max_idle_threads;
+};
+#endif
+
 /* ----------------------------------------------------------- *
  * Channel interface (when using -o clone_fd)		       *
  * ----------------------------------------------------------- */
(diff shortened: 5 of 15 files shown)


lib/fuse_loop_mt.c
```diff
@@ -155,7 +155,7 @@ static void *fuse_do_work(void *data)
 
 		if (!isforget)
 			mt->numavail--;
-		if (mt->numavail == 0)
+		if (mt->numavail == 0 && mt->numworker < mt->max_idle)
```
Nikratio, line 158: This seems to contradict the description of the option - max *idle* threads. We are not creating an idle thread here, but a busy one.

lib/fuse_loop_mt.c
```diff
@@ -164,6 +164,14 @@ static void *fuse_do_work(void *data)
 		pthread_mutex_lock(&mt->lock);
 		if (!isforget)
 			mt->numavail++;
+#if 0
```
Nikratio, line 167: If it doesn't make sense, then let's just remove the code rather than #ifdef it out.

Nikratio (changes requested):
Thank you for investigating this! I think you have discovered an issue that should be fixed, but I am not sure if your fix is the right approach since it changes the code to contradict the documentation and name of the `max_idle` option.

Do I understand correctly that the behavior that gives you better performance could also be achieved by simply setting `max_idle` to a very high number? In that case, perhaps what is needed is not a code change but a modification of the default value? However, I am a bit wary about tuning a default to optimize a benchmark suite. Do you know if there are any practical applications that  benefit from a high number of threads?

lib/fuse_loop_mt.c
aakefbs: Fine we me as well. But  we cannot just delete or comment out this code here before updating the condition above, as libfuse would other create an indefinite number of threads.

aakefbs:
> Thank you for investigating this! I think you have discovered an issue that should be fixed, but I am not sure if your fix is the right approach since it changes the code to contradict the documentation and name of the `max_idle` option.
> 
> Do I understand correctly that the behavior that gives you better performance could also be achieved by simply setting `max_idle` to a very high number? In that case, perhaps what is needed is not a code change but a modification of the default value? However, I am a bit wary about tuning a default to optimize a benchmark suite. Do you know if there are any practical applications that benefit from a high number of threads?

Sorry I had first missed your introduction text. Increasing max_idle to a very high value is not a good idea, imho. My final is for sure not to optimize bonnie, but optimize a real live file system. Bonnie over passthrough_{ll,hp} is just a simple test that can be done locally or on a single system and also can be demonstrated to kernel devs. We have tested different max_idle parameters in the past and the default of 10 was not too bad for us. Increasing the number of threads resulted in worse results. 

Test kernel patches and to demonstrate it to Miklos/fsdevel I also need to restrict the number of threads.

aakefbs:
Also, there might be applications spawning hundreds or thousands of threads - I don't think that increasing the default number of threads to area of thousands is a good idea.
I simple test for this is here https://github.com/aakefbs/fs-heavy-load-stress-tests

lib/fuse_loop_mt.c
Nikratio: I do not think it's a good idea to change the meaning of max_idle (or to drop it) for backwards compatibility reason.

Introducing a new `max_threads` variable seems fine to me. I do not think that introducing a default limit would cause issues with backwards compatibility. I'm less sure about what's a reasonable value though - any suggestions?

(At the same time, I think we should default `max_idle` to max_threads to prevent the expensive destruction/creation cycle by default).

aakefbs: Ok, fine with max_threads. I would set it by default to 10 - same default as 

. But now we cannot simply add another variable to struct fuse_loop_config(), because it is exposed to everything and not versioned - making it larger would cause memory access beyond struct for existing applications not recompiled against the new libfuse. I.e, a new fuse_session_loop_mt_33 is needed. At least from my point of view that also allows to deprecate the max_idle_threads variable.

aakefbs: Hmm, so this is going to be rather big. I stop for now and will finish it in the morning. Posting an unfinished patch here so that you can get an idea (logic part completed, but not even compilation tested...).

aakefbs:
@Nikratio When you review it, please pay attention to FUSE_SYMVER  - the usage of "@@" and "@" is not well documented. I spent ages trying to understand why linking failed and in the end it was the double "@" I was initially using.

Please also double verify lib/fuse_versionscript:  "} FUSE_3.4".

lib/fuse_loop_mt.c
aakefbs: Please let me know what you think about this. I increased the version as this introduces a new API.

include/fuse.h
```diff
@@ -1004,6 +1004,8 @@ void fuse_exit(struct fuse *f);
 #if FUSE_USE_VERSION < 32
 int fuse_loop_mt_31(struct fuse *f, int clone_fd);
 #define fuse_loop_mt(f, clone_fd) fuse_loop_mt_31(f, clone_fd)
+#elif FUSE_USE_VERSION < 311
```
Nikratio, line 1007: Same here, let's always use FUSE_MAKE_VERSION instead of literal numbers.

example/passthrough_hp.cc
```diff
@@ -43,7 +43,7 @@
  * \include passthrough_hp.cc
  */
 
-#define FUSE_USE_VERSION 35
+#define FUSE_USE_VERSION 311
```
Nikratio, line 46: Let's instead use FUSE_MAKE_VERSION(3, 11) here. This will make sure that in the future, no one sets FUSE_USE_VERSION = 40 and gets confused because it sorts as less than 311.

include/fuse_common.h
```diff
@@ -834,6 +847,45 @@ int fuse_set_signal_handlers(struct fuse_session *se);
  */
 void fuse_remove_signal_handlers(struct fuse_session *se);
 
+/**
+ * Create and set default config for fuse_session_loop_mt and fuse_loop_mt.
```
Nikratio, line 851: Great idea, we really should have always had dedicated setters and getters :-).

lib/fuse_i.h
```diff
@@ -87,6 +87,41 @@ struct fuse_module {
...
 
+/**
+ * Configuration parameters passed to fuse_session_loop_mt() and
+ * fuse_loop_mt().
+ *
+ * Internal API to avoid exposing the plain data structure and
+ * causing compat issues after adding or removing struct members.
+ *
+ */
+struct fuse_loop_config2 {
```
Nikratio, line 98: Let's keep the name as is (`fuse_loop_config`), and instead introduce `fuse_loop_config_pre311` for the old struct.

include/fuse_lowlevel.h
```diff
@@ -1968,27 +1970,37 @@ int fuse_session_mount(struct fuse_session *se, const char *mountpoint);
 int fuse_session_loop(struct fuse_session *se);
 
 #if FUSE_USE_VERSION < 32
```
Nikratio, line 1972: I think this is not necessary. If you recompile a program from source, then we can always use the new struct fuse_loop_config (since all code that works with it will be recompiled to take into account the new size). We just have to make sure that the additional member is properly initialized.

The backwards compatibility problem occurs when an existing binary uses a newer version of the libfuse shared object. In that case, fuse_lowlevel.h is never evaluated, so you can't handle this through use of FUSE_USE_VERSION. This will instead be handled through the symver declarations that you already added below.

Nikratio (changes requested):

lib/fuse_i.h
aakefbs: Yes, I had considered this, but I have a concern here. Basically this will break API compatibility. So applications that are using this struct and that were compiling fine against libfuse <= 3.10 would now fail to compile against libfuse >= 3.11. Take for example Ubuntu/Debian, sometimes all packages get recompiled because package format changes or new package options are set - once libfuse is updated to 3.11 quite some old packages would fail cause mass bug reports. Same for arbitrary non-dev users that just want to compile a random fuse file system that didn't get a recent update. So from my point of view and even though ABI compatibility is kept, keeping "fuse_loop_config" for the new struct would be an libfuse-4 change. 
Hmm, I have an idea how to solve this with ifdef on the version. I need to think about it a bit.

Nikratio: Why would they fail to compile if the struct gains an additional member?

include/fuse_lowlevel.h
aakefbs: So lets assume we keep "struct fuse_loop_config" with the additional member and now a program initializes members of that struct without using the constructor and setter - how do we ensure without these changes here that any new struct members are initialized?
I'm more tempted to make the condition for UCLIBC and APPLE more generic (with a bit code dup), to make it better readable.

lib/fuse_i.h
aakefbs: Well the problem I have is that I really wouldn't like to expose the struct details - I had made struct fuse_loop_config2 private for the reason that we need to ensure that members are correctly initialized. As soon as you just add members there is no enforcement anymore that people use fuse_loop_config_create(), even though the current version 311 is defined - all bad and expected things might come up.

include/fuse_lowlevel.h
Nikratio: Hmm, right, we could end up with an undefined value in `max_threads`.

How about this though? Since use of `max_idle_threads` is discouraged, re-purpose it as a flag value. If `max_idle_threads == INT_MAX`, assume that `max_threads` is defined and use it and ignore `max_idle_threads`. If `max_idle_threads` has any other value, assume that `max_thread` is undefined and use some suitable default while continuing to honor `max_idle_threads`.

aakefbs: We would still need to expose the struct, instead of having it private. And with that there is no way to enforce users to set whatever value.
Let me try something else, I _think_ we can keep the old name with a more ifdefs on the version. I'm right now on kernel thread changes, will take care of this patch here a bit later today or tomorrow.
Thanks for your reviews and discussion!

Nikratio: While ideally the struct would be private, I do not consider it worth the extra headache that it causes here. If I'm not mistaken, the above approach means no need for `#ifdefs` and no need for `symver` declarations. This makes the code much easier to understand and maintain.

We can still provide accessor functions and with an opaque pointer type, and recommend to use this interface, without hiding the struct completely.

If someone really wants to mess with internals, then making the struct private is not going to stop them either. They can look at the fuse source code, copy over the private definition, and still mess with it :-).

aakefbs: Well, all the headache is because the struct was not private before. 

I also don't mean that user mess on purpose with internals Someone not carefully looking at the code (that could be easily me at 5 am in the morning after a long long day) might miss to initialize one or multiple values. That is basically were relying on flag values does not help.

And the other thing is, once you make it public (and allow manual settings with that) - any new extension of the struct means that a new FUSE_SYMVER is needed for each and every new parameter - really inconvenient from my point of view. Example, I will probably need another option for my current kernel changes - we really want to have another API definition for that? I'm also trying to get time (for me or a colleague) to work on ioctl threads with shared memory access - similar to zufs and fuse2. More options - another FUSE_SYMVERS needed?

aakefbs: Also, please consider that the fuse version further increases for other reasons. Now file systems increase their version for those other reasons. And without noticing the file system passes an uninitialized struct.
From my point of view passing a struct that can change at any time a severe API bug. I will split the patches once we come to a conclusion how to handle this - the API bug needs a separate commit message. 
From my point of view we are left with 3 solutions
1) As in my patch, introduce a new private struct
2) Try to handle the public < 311 struct with ifdefs. From what I have in my mind it might work, but I have not written the code yet.
3) Give up on the struct an go back to the scheme that passes all parameters to the function directly.

All these 3 schemes enforce the right usage by giving compilation errors.

Nikratio: I agree that having a private struct is better design, and we should have done this from the get go. However, we did not. We are having a public struct, and my worry is that replacing it with a better API will just mean that a large number of filesystems will never migrate to the new API because it's too much trouble. It would be nice if we could offer these filesystems a "minimum effort" route to get rid of the thread creation overhead.

So, in other words, I am not opposed to adding a proper API, I would just prefer if we could also offer another, easier way for filesystems to avoid the thread creation/destruction overhead.

aakefbs: As i wrote, your proposal using flags does not work once the fuse version is increased for other reasons and then file systems do set the new version for that reason - the file system developer will not get automatically informed that there is an API issue. With the hundeds or thousands of different fuse file systems out there this will cause accidental read beyond end of struct - I would be scared of security CVEs...
And as I wrote, increasing the API version for every new option is not good either. 

Now, regarding thread creation overhead - from my point of view we could just ignore that idle parameter and document that it gets ignored in 3.11 . Or we add another parameter and the parameter is only not ignored if that new parameter is given. I personal do not see where thread destruction is needed, but who knows. For my upcoming kernel patches I will need at least one stable thread that will never get destructed... zufs and the unmaintained fuse2 branch from Miklos require to create a thread per core...

aakefbs: There another way - security-by-obscurity. 

```
#define CONFIG_VERSION_VERIFIER1 <some-big-random-number>
#define CONFIG_VERSION_VERIFIER2 <another-big-random-number>
#define CONFIG_VERSION_VERIFIER_STR "I-know-that-I-shall-not-use-this-struct-muanlly-and-I-know-might-cause-read-of-uninitialized-variables-and-security-issues-by-doing-so"

/**
 * beginning with  FUSE_MAKE_VERSION(3, 11)  this struct should be treated a private/opaque
 * and only be accessed through fuse_loop_config_create() and fuse_loop_config_set_*() setters
 * outside of libfuse internals.
 */
struct fuse_loop_config {
...
    unsigned int max_threads;
    
    /* FUSE_MAKE_VERSION(3, 11) starts below  */
    
    uint64_t verifier1;
    uint64_t verifier2;
    char      verifier3[sizeof(CONFIG_VERSION_VERIFIER_STR )+ 1];
}
```

And with that  fuse_loop_config_create() would set these values and we would need another function like fuse_loop_config_verify() that would verify these values.

aakefbs: @BenBE Thanks. This might work, although I'm not entirely sure about "code compiled with the new headers will use the new definitions.". We basically want that code compiled with FUSE_USE_VERSION < 311 gets compiled with the old definition of struct fuse_loop_config. Otherwise recompilation of a file system that was compiling fine before 3.11 would now fail - might cause a mass debian bug report (same for other distributions). And I _think_ that requirement brings in the need for #ifdefs, to define struct fuse_loop_config as the old version for that code. 
I guess the result is that a versioned struct is then not so different to an internal private struct, with setters and getters. Yours has the advantage that a conversion function is not needed, with the disadvantage of the need for more future versioning.

aakefbs: Well, we could reserve some space into fuse_loop_config_versioned for future options, to avoid the need for more FUSE_SYMVER on each and every struct increase.

BenBE: Okay, didn't see that requirement for explicit targeting of old header versions. though could possibly be accomplished by overlaying `fuse_loop_config` as a `#define fuse_loop_config fuse_loop_config_v1`. Absolutely not a situation with a good solution.

Possibly the least messiest idea is to just drop providing `fuse_loop_config` as a struct type itself, but just `fuse_loop_config_versioned`, `fuse_loop_config_v1`, `fuse_loop_config_v2`. Depending on targeted API version `fuse_loop_config` could map to `fuse_loop_config_v1` for pre-3.11 and `fuse_loop_config_versioned` after. You could choose to just embed extra fields directly in `fuse_loop_config_versioned`, but at least for the two fields of `fuse_loop_config_v1` you want an union so typecasting/copying the data between the legacy version and the new one stays straight-forward.

Re conversion:
On entry to any function you basically have:

```c
struct fuse_loop_config_v1 *cfg_unk = (struct fuse_loop_config_v1*)cfg;
const bool have_v1 = cfg_unk->clone_fd != magic;
struct fuse_loop_config_versioned cfg_v = have_v1 ? (struct fuse_loop_config_versioned){ .version = magic, .v1 = *cfg_unk, .v2 = default_v2, … } : *(struct fuse_loop_config_versioned*)cfg;
```

To pass this on you'd do:
```c
#define pass_on(cfg) ((struct fuse_loop_config*)(have_v1 ? &cfg.v1 : &cfg))
```

Both could be further optimized to hide some gory details, but starting from there it shouldn't introduce too much complexity apart from this versioning transition check, which can mostly be hidden when working with things internally. You might even get away with just rerouting the passed config pointer only when you receive a _v1 structure and otherwise using things as is.

aakefbs: Yeah that works, but we still need to have reserved fields in the struct for future extensions (to avoid access beyond passed memory) to avoid FUSE_SYMVER() for every new config  option. So why shouldn't we do the special work only for FUSE_USE_VERSION < 311 and from 311 on we can use the API as in my initial patch? Well, I'm fine with anything as long as I can move ahead - my other patches are depending on this patch :)
The discussion with @Nikratio happend, because Nikolaus disagreed on adding an struct ifdef for < 311. I think your suggestions are more complicated that that macro. Well, I need to upload a new patch to let you have look.

Nikratio:
I think I may have poorly expressed some of my previous comments. I am not fundamentally opposed to anything that you're proposing. I may have different thoughts on some of your suggestions, but I am open to be convinced that your way is the better one. So please do not rule anything out just because I said I didn't like it on first sight.

How about you rework the pull request along the lines that you consider best on all the discussions above, and we take a fresh look at that?

include/fuse_lowlevel.h
aakefbs: Just uploaded an updated version. Basically the old struct is available in fuse_common.h

```
#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 11)
struct fuse_loop_config_v1; /* forward declarition */
struct fuse_loop_config {
#else
struct fuse_loop_config_v1 {
#endif
	/**
```

and the new struct is in fuse_i.h

```
#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3,11)
struct fuse_loop_config_v2
#else
struct fuse_loop_config
#endif
{
```

What do you think?

Nikratio:
Just wondering about the current status of this.. are you still working on this? Or is this waiting for my review?

aakefbs:
@Nikratio Please see my comment from 24 days ago - I had asked about your opinion (@BenBE is welcome to comment as well) about the  apprroach using ifdef. The patch is complete, but I need to double check if binary and api compatibility works.

include/fuse_lowlevel.h
```diff
@@ -1968,26 +1970,29 @@ int fuse_session_mount(struct fuse_session *se, const char *mountpoint);
...
- * @param se the session
- * @param config session loop configuration 
- * @return see fuse_session_loop()
- */
-int fuse_session_loop_mt(struct fuse_session *se, struct fuse_loop_config *config);
-#else
-int fuse_session_loop_mt_32(struct fuse_session *se, struct fuse_loop_config *config);
-#define fuse_session_loop_mt(se, config) fuse_session_loop_mt_32(se, config)
-#endif
+	#if (!defined(__UCLIBC__) && !defined(__APPLE__))
```
Nikratio, line 1979: There seem to be  a lot of spurious changes here (maybe space vs tab). Could you get rid of those?

Nikratio:
I'm not sure why we need the `version_id` in the struct if the struct is private to FUSE. But other than that, this looks fine to me.

include/fuse_lowlevel.h
aakefbs: Hmm, it is a bit hard to read the conditions without any indentation - I had added that on purpose.

```
#if FUSE_USE_VERSION < 32
	int fuse_session_loop_mt_31(struct fuse_session *se, int clone_fd);
	#define fuse_session_loop_mt(se, clone_fd) fuse_session_loop_mt_31(se, clone_fd)
#elif FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 11)
	int fuse_session_loop_mt_32(struct fuse_session *se, struct fuse_loop_config *config);
	#define fuse_session_loop_mt(se, config) fuse_session_loop_mt_32(se, config)
#else
	#if (!defined(__UCLIBC__) && !defined(__APPLE__))
	/**
	 * Enter a multi-threaded event loop.
	 *
	 * For a description of the return value and the conditions when the
	 * event loop exits, refer to the documentation of
	 * fuse_session_loop().
	 *
	 * @param se the session
	 * @param config session loop configuration
	 * @return see fuse_session_loop()
	 */
		int fuse_session_loop_mt(struct fuse_session *se, struct fuse_loop_config *config);
	#else
		int fuse_session_loop_mt_311(struct fuse_session *se, struct fuse_loop_config *config);
		#define fuse_session_loop_mt(se, config) fuse_session_loop_mt_311(se, config)
	#endif
#endif
```
vs

```
#if FUSE_USE_VERSION < 32
int fuse_session_loop_mt_31(struct fuse_session *se, int clone_fd);
#define fuse_session_loop_mt(se, clone_fd) fuse_session_loop_mt_31(se, clone_fd)
#elif FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 11)
int fuse_session_loop_mt_32(struct fuse_session *se, struct fuse_loop_config *config);
#define fuse_session_loop_mt(se, config) fuse_session_loop_mt_32(se, config)
#else
#if (!defined(__UCLIBC__) && !defined(__APPLE__))
/**
 * Enter a multi-threaded event loop.
 *
 * For a description of the return value and the conditions when the
 * event loop exits, refer to the documentation of
 * fuse_session_loop().
 *
 * @param se the session
 * @param config session loop configuration
 * @return see fuse_session_loop()
 */
int fuse_session_loop_mt(struct fuse_session *se, struct fuse_loop_config *config);
#else
int fuse_session_loop_mt_311(struct fuse_session *se, struct fuse_loop_config *config);
#define fuse_session_loop_mt(se, config) fuse_session_loop_mt_311(se, config)
#endif
#endif
```

aakefbs:
@Nikratio  I kept the 'spurious' ifdef indentation for, as I personally have hard time to decipher it without that. Please let me know what you think - I can remove it. 
In any way, please don't merge yet, I didn't manage to run some manual tests yet and I'm not sure if abi/api compatibility is covered already by the existing automatic tests.

Nikratio:
It's a bit odd to have top level function definitions indented, especially with the long multi-line comment. But I guess it's okay.

aakefbs:
@Nikratio  Don't merge it yet please, I eventually found the time to test today and there is another ABI/API issue. I had extended struct fuse_cmdline_opts, but old binaries obviously only pass an old smaller version and that causes a buffer overflow. Looks like we need versioning for that as well. Not sure yet if we want to make it private or follow another way to detect the struct version.

aakefbs:
I hope that I fixed all issues now, although I'm not entirely sure about the ulibc/apple fuse_parse_cmdline() compatible symbol. I don't have a non-linux system - hard to test for me.

aakefbs:
I don't see github tests running anymore?

aakefbs:
@Nikratio I'm just reviewing a bit my own changes and see this 

```
[lib/meson.build](https://github.com/aakefbs/libfuse/commit/db734ebaae5203f1bac333525bbdb44992e027ad?diff=unified#diff-f95a8510ab53a5e21d9e4a306b12f76fc79086143aa8989c13effed889bec144)
@@ -37,7 +37,7 @@ libfuse = library('fuse3', libfuse_sources, version: meson.project_version(),
                  soversion: '3', include_directories: include_dirs,
                  dependencies: deps, install: true,
                  link_depends: 'fuse_versionscript',
                  c_args: [ '-DFUSE_USE_VERSION=35',
                  c_args: [ '-DFUSE_USE_VERSION=312',
```

This is with your 3.11 patch applied - why is part of libfuse compiled so far with `-DFUSE_USE_VERSION=35`?

aakefbs:
So I'm lost here what to do about travis - I don't know what "Because some parts of FUSE haven't been updated (or confirmed not to need changes) to use the most recent API." means.
I just set up travis for my own github account and updated patches triggered travis there.

https://app.travis-ci.com/github/aakefbs/libfuse/builds/250710225

I was search for "github does not run travis" and at best I find that one should switch to github actions, hmm.

I think missing in the tests are ABI compat checks - build an unpatched version, store the example/ binaries, then add in the patches, compile and run old binaries with the library. I don't know how to automate this with github/travis, though, as it means one needs to compile a different commit than currently provided. I had manually done this, except of ULIBC/APPLE.

Nikratio:
> So I'm lost here what to do about travis 

Uhm, what's wrong with Travis? And is this related to the pull request?

> - I don't know what "Because some parts of FUSE haven't been updated (or confirmed not to need changes) to use the most recent API." means. 

This is unrelated to Travis, I think?

It means that some parts of the FUSE codebase (eg the examples) have not been updated to use the most recent libfuse API. We can leave this as-is (it's not looking great from a maintenance point of view, but there is nothing technically wrong with it). Or, someone would need to look how the API has changed and - where required - update the code to make use of the new API.

Makes sense?

aakefbs:
May question was why Travis does not run on the current patches. So you say that because I didn't update everything to the new API it does not run travis? How would it know that?

I have no issue to update the all the other examples, but I think we should add some tests for the old API - I had basically used non-converted examples to test that I didn't break anything (and initially found some issues).
Any objections to merge and I work on API tests during the next days?

Nikratio:
> May question was why Travis does not run on the current patches. 

Ah, got question. Looks like an issue on the Travis side - there are "no credits left". But as an open-source project we shouldn't need any credits. I've emailed support.

> So you say that because I didn't update everything to the new API it does not run travis?

No, I'm saying this is totally unrelated.

> I have no issue to update the all the other examples, but I think we should add some tests for the old API - I had basically used non-converted examples to test that I didn't break anything (and initially found some issues). Any objections to merge and I work on API tests during the next days?

Let's wait until Travis is back-up (hopefully support will respond soon), just to make sure we don't break CI by accident.

Nikratio:
Travis should be running again, it seems you need to reapply for open source credits once a year

aakefbs:
Thanks for fixing it! Looks like Travis verified the request. Any objections to merge?

Nikratio:
It looks good. Please just add an entry to `ChangeLog.rst` describing the change and how filesystems need to be changed correspondingly.

aakefbs:
@Nikratio Any objections to merge?

hswong3i:
It is now broken with legacy OS older then Ubuntu 20.04, with missing `fuse_parse_cmdline@FUSE_3.0` symbol, see #696.

Since I am using fuse together with fuse-overlayfs, where they default with `#define FUSE_USE_VERSION 32` (see https://github.com/containers/fuse-overlayfs/blob/main/main.c#L22), shall I also compile fuse 3.12.0 with `-DFUSE_USE_VERSION=35` or even `-DFUSE_USE_VERSION=32`?

aakefbs:
@hswong3i I'm going to look into it now. Thanks for reporting it.

hswong3i:
@aakefbs thank you very much, https://github.com/libfuse/libfuse/issues/696#issuecomment-1242892812 show that following OS are now failed:

> I could figure out following OS are FAILED with this missing `fuse_parse_cmdline@FUSE_3.0` symbol (most likely, OS with basic libraries older or equal than Ubuntu 20.04 could fail):
> * Ubuntu 20.04
> * Ubuntu 18.04
> * Debian 10
> * CentOS 8 Stream
> * CentOS 7
> * openSUSE Leap 15.4
> * openSUSE Leap 15.3

PR #653, opened 2022-03-28
Passthrough hp

aakefbs:
Two fixes - disable splice and and inode ref leak.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -357,7 +357,14 @@ static int do_lookup(fuse_ino_t parent, const char *name,
             cerr << "DEBUG: lookup(): inode " << e->attr.st_ino
                  << " (userspace) already known; fd = " << inode.fd << endl;
         lock_guard<mutex> g {inode.m};
+
         inode.nlookup++;
+        if (fs.debug)
+            cerr << "DEBUG:" << __func__ << ":" << __LINE__ << " "
+                 <<  "inode " << inode.src_ino
+                 << " count " << inode.nlookup << endl;
+
+
         close(newfd);
     } else { // no existing inode
         /* This is just here to make Helgrind happy. It violates the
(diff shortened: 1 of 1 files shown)


example/passthrough_hp.cc
```diff
@@ -194,13 +194,11 @@ static void sfs_init(void *userdata, fuse_conn_info *conn) {
...
         conn->want |= FUSE_CAP_FLOCK_LOCKS;
 
-    // Use splicing if supported. Since we are using writeback caching
-    // and readahead, individual requests should have a decent size so
-    // that splicing between fd's is well worth it.
-    if (conn->capable & FUSE_CAP_SPLICE_WRITE && !fs.nosplice)
-        conn->want |= FUSE_CAP_SPLICE_WRITE;
-    if (conn->capable & FUSE_CAP_SPLICE_READ && !fs.nosplice)
-        conn->want |= FUSE_CAP_SPLICE_READ;
+    // splice is enabled in libfuse3 by default
```
Nikratio, line 197: What's the intent behind this change? Note that conn->capable checks the capabilities of the *kernel*, not libfuse.

aakefbs: You mean the comment update? I'm aware that conn->capable are for kernel features - this is not changed here, but conn->want disables splice if splice is not enabled. The purpose is to be able to disable splice, as splice has quite some overhead for small data. Right now this cannot be demonstrated with passthrough_hp, as enabled splice by default create bug and broke its '--nosplice' option. This is what the comment update is about.

aakefbs: Also see your commit 3ff500f9f25a37616ee69d2f876ac087c728895b, which sets FUSE_CAP_SPLICE_READ by default.

@@ -1928,6 +1928,7 @@ static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
        if ((cond) && (f->conn.capable & (cap))) \
                f->conn.want |= (cap)
        LL_SET_DEFAULT(1, FUSE_CAP_ASYNC_READ);
+       LL_SET_DEFAULT(f->op.write_buf, FUSE_CAP_SPLICE_READ);
        LL_SET_DEFAULT(f->op.getlk && f->op.setlk,
                       FUSE_CAP_POSIX_LOCKS);
        LL_SET_DEFAULT(f->op.flock, FUSE_CAP_FLOCK_LOCKS);

I.e. FUSE_CAP_SPLICE_READ is set by default.

aakefbs: Actually FUSE_CAP_SPLICE_WRITE was not enabled by default, fixed in a patch update.

Nikratio:
Looks good, thanks. Will merge when CI run is complete.

aakefbs:
Did I overload travis and it does not want to run anymore?

aakefbs:
Thank you!

PR #654, opened 2022-03-28
passthrough_hp: Add options for clone_fd and number of threads

bsbernd:
This is useful for benchmarking fuse.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -695,7 +695,7 @@ static bool is_dot_or_dotdot(const char *name) {
 
 
 static void do_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
-                    off_t offset, fuse_file_info *fi, int plus) {
+                    off_t offset, fuse_file_info *fi, const int plus) {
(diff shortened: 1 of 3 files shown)


example/passthrough_hp.cc
```diff
@@ -1241,10 +1257,12 @@ int main(int argc, char *argv[]) {
     // Don't apply umask, use modes exactly as specified
     umask(0);
 
+    fuse_daemonize(fs.foreground);
+
     // Mount and run main loop
     struct fuse_loop_config loop_config;
     loop_config.clone_fd = 0;
-    loop_config.max_idle_threads = 10;
+    loop_config.max_idle_threads = fs.num_threads;
```
Nikratio, line 1265: There is some confusion here. This is the maximum number of idle threads, not the number of threads that is being used.

Other than that, adding the extra options looks good to me.

aakefbs: Well, we can postpone this patch until we are done with #652 (thread creation and destruction issue).

aakefbs:
Sorry, for the late update, I was rather busy during the past months.

PR #655, opened 2022-04-08
Parallel writes

d-hans:
Allow parallel direct writes on the same file


d-hans:
closing this pull request as it got combined with other patches whereas wants to raise it for parallel write only

PR #659, opened 2022-04-08
Enable parallel direct writes on the same file

d-hans:
This patch allows parallel direct writes on the same file to be enabled with the help of a flag FOPEN_PARALLEL_DIRECT_WRITES. FUSE implementations which want to use this feature
can set this flag during fuse init. Default behavior remains same i.e no parallel direct writes on
the same file.

Corresponding fuse kernel patch(Merged).
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v6.2&id=153524053bbb0d27bb2e0be36d1b46862e9ce74c

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -59,7 +59,7 @@ struct fuse_file_info {
 	    requests if write caching had been disabled. */
 	unsigned int writepage : 1;
 
-	/** Can be filled in by open, to use direct I/O on this file. */
+	/** Can be filled in by open/create, to use direct I/O on this file. */
 	unsigned int direct_io : 1;
 
 	/** Can be filled in by open and opendir. It signals the kernel that any
@@ -67,6 +67,10 @@ struct fuse_file_info {
 	    last time the file/directory was open) need not be invalidated when
 	    the file/directory is closed. */
 	unsigned int keep_cache : 1;
+
+	/** Can be filled by open/create, to allow parallel direct writes on this
+         *  file */
+        unsigned int parallel_direct_writes : 1;
 
 	/** Indicates a flush operation.  Set in flush operation, also
 	    maybe set in highlevel lock operation and lowlevel release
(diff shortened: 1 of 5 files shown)


Nikratio:
Thanks! Could you please ping this request once the necessary patches have been merged into the kernel?

bsbernd (approved):
Support for parallel writes is released in linux/6.2 (commit 153524053bbb0d27bb2e0be36d1b46862e9ce74c).

lib/fuse.c
```diff
@@ -4669,6 +4675,7 @@ static const struct fuse_opt fuse_lib_opts[] = {
 	FUSE_LIB_OPT("noforget",              remember, -1),
 	FUSE_LIB_OPT("remember=%u",           remember, 0),
 	FUSE_LIB_OPT("modules=%s",	      modules, 0),
+	FUSE_LIB_OPT("parallel_direct_write=%d",parallel_direct_writes, 0),
```
bsbernd, line 4678: Nit: There should be a space after the comma.

bsbernd (approved):
I guess we can enable this for passthrough_ll, passthrough_hp, etc, but can be done in a separate commit.

bsbernd:
@d-hans there is a merge conflict, can you please rebase?

lib/fuse.c
d-hans: done

include/fuse_common.h
```diff
@@ -68,6 +68,10 @@ struct fuse_file_info {
 	    the file/directory is closed. */
 	unsigned int keep_cache : 1;
 
+	/** Can be filled by open/create, to allow parallel direct writes on this
+         *  file */
+        unsigned int parallel_direct_writes;
```
bsbernd, line 73: : 1 is missing

bsbernd (changes requested):

include/fuse_common.h
d-hans: fixed

bsbernd (approved):

PR #661, opened 2022-04-11
Add missing kernel flags up to 1ULL << 33

aakefbs:
Just a further sync with the in-kernel flags.


aakefbs:
@d-hans Please review.

PR #662, opened 2022-04-11
Set FUSE_INIT_EXT in fuse_init_out::flags

aakefbs:
It is better to tell the kernel that libfuse knows
about the 64 bit flag extension.


aakefbs:
Patch for the kernel side also follows.

aakefbs:
@d-hans Please review.

PR #671, opened 2022-05-02
Implement Atomic lookup + create

d-hans:
There are certain places where we can avoid costly lookup calls
into libfuse from fuse kernel. One of such place is
1) When we go for opening the file with O_CREAT flags.

Since we are going to create a file (passing O_CREAT in open),
its very much likely that file does not exist yet so lookup
performed before creating a file can be avoided. Instead this
lookup can be performed as part of create call itself.

Now with these changes, user space when receiving atomic create is
supposed to create the file if it does not exist/perform lookup on
the file if already existing, fill in the attributes, set 'file_created'
bit in 'struct fuse_file_info' and open the file. These filled in
attributes are used by fuse kernel to make inode stand/revalidate.

Link to the kernel patch is
https://lore.kernel.org/linux-fsdevel/20220502054628.25826-1-dharamhans87@gmail.com/T/#t

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -633,6 +633,30 @@ struct fuse_operations {
 	int (*create) (const char *, mode_t, struct fuse_file_info *);
 
 	/**
+	 *
+	 * Note: This call is an optimization to avoid lookusp before create.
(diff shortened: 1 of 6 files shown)


d-hans:
duplicate

PR #682, opened 2022-06-20
Increase meson min version and avoid get_pkgconfig_variable warning

bsbernd:
meson was complaining:

Build targets in project: 27
NOTICE: Future-deprecated features used:
 * 0.56.0: {'Dependency.get_pkgconfig_variable'}

So change to .get_variable(pkgconfig : 'type' and also increase
the meson minimal version to be able to handle it.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('libfuse3', ['c'], version: '3.11.0',
-        meson_version: '>= 0.42',
+        meson_version: '>= 0.51',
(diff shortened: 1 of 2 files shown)


aakefbs:
@Nikratio I think you merged a bit early - I guess we need to update the travis configuration - it is a bit old and looks like it doesn't have the required meson version.

aakefbs:
I will try to look into it later on today

PR #686, opened 2022-07-02
fix use after free in iconv module

nm004:
After realloc, newpath shouldn't be used if newpath != tmp.

### DIFF
--- a/lib/modules/iconv.c
+++ b/lib/modules/iconv.c
@@ -82,7 +82,8 @@ static int iconv_convpath(struct iconv *ic, const char *path, char **newpathp,
 			if (!tmp)
 				goto err;
 
-			p = tmp + (p - newpath);
+			if (tmp != newpath)
+				p = tmp;
 			plen += inc;
 			newpath = tmp;
 		}


lib/modules/iconv.c
```diff
@@ -82,7 +82,8 @@ static int iconv_convpath(struct iconv *ic, const char *path, char **newpathp,
 			if (!tmp)
 				goto err;
 
-			p = tmp + (p - newpath);
+			if (tmp != newpath)
+				p = tmp;
```
aakefbs, line 86: Are you sure this is right? "p - newpath" looks like a pointer offset - calculate this before the realloc?

ptrdiff_t ptr_off = p - newpath;
tmp = realloc(newpath, newpathlen + 1);
...

p = tmp + ptr_off;

aakefbs (changes requested):
I think the pointer arithmetic should be kept.

Nikratio:
I'm resolving this for now, please feel free to re-open when you had a chance to look at the comments.

PR #698, opened 2022-09-13
Fix build failure with uclibc using HAVE_SYMVER_ATTRIBUTE macro and

giuliobenetti:
moving handling to helper.c file

Building with uclibc leads to failure:
```
FAILED: lib/libfuse3.so.3.12.0.p/helper.c.o.
/home/giuliobenetti/git/upstream/test-libfuse3/bootlin-armv5-uclibc/host/bin/arm-linux-gcc -Ilib/libf
In file included from ../lib/fuse_i.h:10,
                 from ../lib/helper.c:14:
../include/fuse_lowlevel.h:1921:40: error: redefinition of ‘fuse_parse_cmdline_312’
 1921 | #define fuse_parse_cmdline(args, opts) fuse_parse_cmdline_312(args, opts)
      |                                        ^~~~~~~~~~~~~~~~~~~~~~
../lib/helper.c:258:5: note: in expansion of macro ‘fuse_parse_cmdline’
  258 | int fuse_parse_cmdline(struct fuse_args *args,
      |     ^~~~~~~~~~~~~~~~~~
../lib/helper.c:208:5: note: previous definition of ‘fuse_parse_cmdline_312’ was here
  208 | int fuse_parse_cmdline_312(struct fuse_args *args,
```
This happens because uclibc, depending on version, can support symver, so if symver is supported and uclibc is used function fuse_parse_cmdline_312() will be defined twice:
1. the function itself with symver
2. fuse_parse_cmdline() as the #define of fuse_parse_cmdline_312() and its prototype
This leads to have the redefinition of ‘fuse_parse_cmdline_312’.

To solve this let's check against HAVE_SYMVER_ATTRIBUTE instead of __UCLIBC__ and __APPLE__ and move all the checks of FUSE_USE_VERSION to helper.c file instead of fuse_lowlevel.h with #define fuse_parse_cmdline that defines fuse_parse_cmdline_312() twice in helper.c leading to the error above.

Signed-off-by: Giulio Benetti <giulio.benetti@benettiengineering.com>

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -251,18 +251,6 @@ int fuse_parse_cmdline_30(struct fuse_args *args,
 	return rc;
 }
 
-/**
- * Compatibility ABI symbol for systems that do not support version symboling
- */
-#if (defined(__UCLIBC__) || defined(__APPLE__))
-int fuse_parse_cmdline(struct fuse_args *args,
-		       struct fuse_cmdline_opts *opts)
-{
-	return fuse_parse_cmdline_30(args, out_opts);
-}
-#endif
-
-
 int fuse_daemonize(int foreground)
 {
 	if (!foreground) {
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -1914,11 +1914,19 @@ int fuse_parse_cmdline(struct fuse_args *args,
 #if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
 int fuse_parse_cmdline_30(struct fuse_args *args,
 			   struct fuse_cmdline_opts *opts);
-#define fuse_parse_cmdline(args, opts) fuse_parse_cmdline_30(args, opts)
+static int fuse_parse_cmdline(struct fuse_args *args,
+			      struct fuse_cmdline_opts *opts)
+{
+	return fuse_parse_cmdline_30(args, opts);
+}
 #else
 int fuse_parse_cmdline_312(struct fuse_args *args,
 			   struct fuse_cmdline_opts *opts);
-#define fuse_parse_cmdline(args, opts) fuse_parse_cmdline_312(args, opts)
+static int fuse_parse_cmdline(struct fuse_args *args,
+			      struct fuse_cmdline_opts *opts)
+{
+	return fuse_parse_cmdline_312(args, opts);
+}
 #endif
 #endif
 


giuliobenetti:
Travis CI fails due a warning that is not in the scope of this patch.

aakefbs:
@giuliobenetti Why did you close it? I'm not sure if your patch is right. I agree it should test for HAVE_SYMVER_ATTRIBUTE, but if this is not defined, shouldn't it still test for apple and ulibc? That way it would also work on older systems that do not have the HAVE_SYMVER_ATTRIBUTE? I guess we should create a meson test here.

aakefbs:
Can you please re-open or create an issue?

giuliobenetti:
Hi @aakefbs,

> @giuliobenetti Why did you close it? I'm not sure if your patch is right. I agree it should test for HAVE_SYMVER_ATTRIBUTE, but if this is not defined, shouldn't it still test for apple and ulibc?

Yes, that's the point I've finally found. I was mixing up 2 changes in 1 patch.

> That way it would also work on older systems that do not have the HAVE_SYMVER_ATTRIBUTE?

It doesn't indeed, you've caught it :-)

> I guess we should create a meson test here.

Yes, but that IMHO should be another patch. So here I've updated the patch that only move the fuse_parse_cmdline() handling in helper.c depending on FUSE_USE_VERSION and let the fuse_parse_cmdline() only in header file. This fixes the bug for 44 different builds with random arch-gcc/libc(uclibc/glibc/musl).

The meson test is already there but incomplete:
https://github.com/libfuse/libfuse/blame/master/meson.build#L93
But that can be improved later and is not mandatory to fix the bug.

The only "drawback" of this new patch is that uclibc never uses symver even if it's supported, while __APPLE__ never supports it.

aakefbs:
Thanks for the update and thanks for fixing my patch. I will check in the morning.

aakefbs (approved):
Thanks for the fix and updates!

lib/helper.c
```diff
@@ -252,17 +252,20 @@ int fuse_parse_cmdline_30(struct fuse_args *args,
...
 /**
- * Compatibility ABI symbol for systems that do not support version symboling
+ * Compatibility symbol for systems that do not support version symboling
  */
 #if (defined(__UCLIBC__) || defined(__APPLE__))
 int fuse_parse_cmdline(struct fuse_args *args,
 		       struct fuse_cmdline_opts *opts)
 {
-	return fuse_parse_cmdline_30(args, out_opts);
+#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
```
Nikratio, line 261: This is evaluated at compile time of the fuse library. However, the FUSE_USE_VERSION macro is defined at compile time of the application that links to FUSE. So I think that this will not have the intended effect, applications will get the most recent API even if they declare wanting to use an older one.

aakefbs:
@Nikratio Yes sure, but what is the way out for libc's that do not have symbol versioning?

Nikratio:
I do not know. But it seems to me that breaking API compatibility for *every* daemon to avoid problems with *some* libcs is not the right trade-off.

aakefbs:
This is only with "#if (defined(__UCLIBC__) || defined(__APPLE__))"?

giuliobenetti:
Then, what about in fuse_lowlevel.h
```
#if (!defined(__UCLIBC__) && !defined(__APPLE__))
int fuse_parse_cmdline(struct fuse_args *args,
                       struct fuse_cmdline_opts *opts);
#else
#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
int fuse_parse_cmdline_30(struct fuse_args *args,
                           struct fuse_cmdline_opts *opts);
static int fuse_parse_cmdline(struct fuse_args *args,
                       struct fuse_cmdline_opts *opts)
{
        return fuse_parse_cmdline_30(args, out_opts);
}
#else
int fuse_parse_cmdline_312(struct fuse_args *args,
                           struct fuse_cmdline_opts *opts);
static int fuse_parse_cmdline(struct fuse_args *args,
                       struct fuse_cmdline_opts *opts)
{
        return fuse_parse_cmdline_312(args, out_opts);
}
#endif
#endif
```
and in helper.c remove the entire:
```
/**
 * Compatibility ABI symbol for systems that do not support version symboling
 */
#if (defined(__UCLIBC__) || defined(__APPLE__))
int fuse_parse_cmdline(struct fuse_args *args,
                       struct fuse_cmdline_opts *opts)
{
        return fuse_parse_cmdline_30(args, out_opts);
}
#endif
```
?

That fixes the bug and the overhead should be almost the same as the '#define' approach.

lib/helper.c
giuliobenetti: I haven't taken into account this situation to be honest. I've pointed another approach below.

aakefbs: I might be blind, but what is the final difference between both approaches? For initial patch, it only affects ULIBC and APPLE - is not global glibc problem? And then in either way for ulibc and apple - there is no way to have ABI compatibility, as the libc does not support it?

aakefbs: Ah, actually the header approach should work better, because it gets compiled into the calling code. Better use "static inline" please.

giuliobenetti: Yes, but with the first approach once you've built the library you can't change the FUSE_USE_VERSION anymore, while in the second approach you can. It's about redirecting fuse_parse_cmdline() to fuse_parse_cmdline_30/312() but while building an application that is linking against libfuse. I think this is what @Nikratio means.
But I'm not an expert of libfuse and I've never used FUSE_USE_VERSION to be honest.

giuliobenetti: > Ah, actually the header approach should work better, because it gets compiled into the calling code. Better use "static inline" please, to avoid creating the same symbol multiple times.

Ok, I go for it then. Thanks both for reviewing!

giuliobenetti:
I've updated and build-tested succesfully with:
- glibc
- uclibc
- musl
on Buildroot.

aakefbs:
@giuliobenetti Could you test this patch?
```
diff --git a/meson.build b/meson.build
index 3cef64fd..0695cff7 100644
--- a/meson.build
+++ b/meson.build
@@ -100,6 +100,13 @@ endif
 # unfortunately return true giving a false positive. So let's try to build
 # using __attribute__ ((symver )) and see the result.
 code = '''
+#if (defined(__UCLIBC__) || defined(__APPLE__))
+/* compilers under ulibc (and apple?) might define the symbol
+ * but their libc still does not support it - fail the test
+ * on these platforms
+ */
+#error
+#endif
 __attribute__ ((symver ("test@TEST")))
 void foo(void) {
 }
```

giuliobenetti:
> @giuliobenetti Could you test this patch?
> 
> ```
> diff --git a/meson.build b/meson.build
> index 3cef64fd..0695cff7 100644
> --- a/meson.build
> +++ b/meson.build
> @@ -100,6 +100,13 @@ endif
>  # unfortunately return true giving a false positive. So let's try to build
>  # using __attribute__ ((symver )) and see the result.
>  code = '''
> +#if (defined(__UCLIBC__) || defined(__APPLE__))
> +/* compilers under ulibc (and apple?) might define the symbol
> + * but their libc still does not support it - fail the test
> + * on these platforms
> + */
> +#error
> +#endif
>  __attribute__ ((symver ("test@TEST")))
>  void foo(void) {
>  }
> ```

Hi @aakefbs, sure I can. Is the purpose to substitute `if (defined(__UCLIBC__) || defined(__APPLE__))` with HAVE_SYMVER_ATTRIBUTE?
Becuase that doesn't fix the problem we have here :-/

aakefbs:
I fear there is another problem with your patch - anything compiled against a previous libfuse won't be  ABI compatible, as the fuse_parse_cmdline() symbol will not be defined. So I thought about another solution - your initial report was that the symbol gets defined twice with ullibc. So if the meson detection excludes ulibc that should get solved? 
Is current ulibc already supporting versioned symbols right now or is this a future feature? It also wouldn't work with your patch either?

giuliobenetti:
> I fear there is another problem with your patch - anything compiled against a previous libfuse won't be ABI compatible, as the fuse_parse_cmdline() symbol will not be defined.

Oh, because it would be 'static fuse_parse_cmdline()', you're right!

>So I thought about another solution - your initial report was that the symbol gets defined twice with ullibc. So if the meson detection excludes ulibc that should get solved?

Unfortunately not, I've modified the commit log. The problem is due to this:
```
there is a #define
    in low_level.h header that is included in helper.c; that defines twice
    the same function because:
    in fuse_lowlevel.h:
    ```
    int fuse_parse_cmdline_312(struct fuse_args *args,
    			   struct fuse_cmdline_opts *opts);
    ```
    While in helper.c:
    ```
    int fuse_parse_cmdline_312(struct fuse_args *args,
    			   struct fuse_cmdline_opts *opts)
    {
    	....
    }
    
    int fuse_parse_cmdline(struct fuse_args *args,
    		       struct fuse_cmdline_opts *opts)
    {
    	....
    }
    ```
    and:
    makes helper.c expands to:
    ```
    int fuse_parse_cmdline_312(struct fuse_args *args,
    			   struct fuse_cmdline_opts *opts)
    {
    	....
    }
    
    int fuse_parse_cmdline_312(struct fuse_args *args,
    		       struct fuse_cmdline_opts *opts)
    {
    	....
    }
    ```
    though fuse_parse_cmdline_312() defined twice.
```

> Is current ulibc already supporting versioned symbols right now or is this a future feature?

uclibc already supports it, but we can also avoid to use it.

> It also wouldn't work with your patch either?

It would work but it break the ABI, that's the problem

aakefbs:
@giuliobenetti Thanks for your help to solve this. I think I know what we need to do - the `fuse_parse_cmdline` compat symbol needs to go into a separated .c file that does not get any other includes with conflicting macros, but just has the ABI compat symbol. New compilation then get API compatibility with macros. 
I'm working right now on something else and need to finish that, I will come up with a new patch later on.

giuliobenetti:
> @giuliobenetti Thanks for your help to solve this. I think I know what we need to do - the `fuse_parse_cmdline` compat symbol needs to go into a separated .c file that does not get any other includes with conflicting macros, but just has the ABI compat symbol. New compilation then get API compatibility with macros. I'm working right now on something else and need to finish that, I will come up with a new patch later on.

Ok, thank you. I can test it later then :-)

aakefbs:
I have two patches here
https://github.com/aakefbs/libfuse/commits/ulibc

One is to fix the fuse_parse_cmdline issue and the other is to test __APPLE__ and __ULIBC__. The test fails with a new issue

```[42/63] Linking target lib/libfuse3.so.3.12.0
FAILED: lib/libfuse3.so.3.12.0 
cc  -o lib/libfuse3.so.3.12.0 lib/libfuse3.so.3.12.0.p/fuse.c.o lib/libfuse3.so.3.12.0.p/fuse_loop.c.o lib/libfuse3.so.3.12.0.p/fuse_loop_mt.c.o lib/libfuse3.so.3.12.0.p/fuse_lowlevel.c.o lib/libfuse3.so.3.12.0.p/fuse_opt.c.o lib/libfuse3.so.3.12.0.p/fuse_signals.c.o lib/libfuse3.so.3.12.0.p/buffer.c.o lib/libfuse3.so.3.12.0.p/cuse_lowlevel.c.o lib/libfuse3.so.3.12.0.p/helper.c.o lib/libfuse3.so.3.12.0.p/modules_subdir.c.o lib/libfuse3.so.3.12.0.p/mount_util.c.o lib/libfuse3.so.3.12.0.p/fuse_log.c.o lib/libfuse3.so.3.12.0.p/compat.c.o lib/libfuse3.so.3.12.0.p/mount.c.o lib/libfuse3.so.3.12.0.p/modules_iconv.c.o -Wl,--as-needed -Wl,--no-undefined -shared -fPIC -Wl,--start-group -Wl,-soname,libfuse3.so.3 -Wl,--version-script,/home/bschubert2/src/libfuse/github/lib/fuse_versionscript -pthread -ldl -lrt -Wl,--end-group
/usr/bin/ld: lib/libfuse3.so.3.12.0.p/helper.c.o: in function `fuse_main_real':
/home/bschubert2/src/libfuse/github/build-ubuntu/../lib/helper.c:380: undefined reference to `fuse_loop_mt'
```

That is probably a real issue, as there is no test in fuse.h for __APPLE__ and __ULIBC__.  But I wonder how this ever worked, I think I had only extended the macro for "FUSE_MAKE_VERSION(3, 12)".

giuliobenetti:
> I have two patches here https://github.com/aakefbs/libfuse/commits/ulibc
> 
> One is to fix the fuse_parse_cmdline issue and the other is to test **APPLE** and **ULIBC**. The test fails with a new issue
> 
> ```
> FAILED: lib/libfuse3.so.3.12.0 
> cc  -o lib/libfuse3.so.3.12.0 lib/libfuse3.so.3.12.0.p/fuse.c.o lib/libfuse3.so.3.12.0.p/fuse_loop.c.o lib/libfuse3.so.3.12.0.p/fuse_loop_mt.c.o lib/libfuse3.so.3.12.0.p/fuse_lowlevel.c.o lib/libfuse3.so.3.12.0.p/fuse_opt.c.o lib/libfuse3.so.3.12.0.p/fuse_signals.c.o lib/libfuse3.so.3.12.0.p/buffer.c.o lib/libfuse3.so.3.12.0.p/cuse_lowlevel.c.o lib/libfuse3.so.3.12.0.p/helper.c.o lib/libfuse3.so.3.12.0.p/modules_subdir.c.o lib/libfuse3.so.3.12.0.p/mount_util.c.o lib/libfuse3.so.3.12.0.p/fuse_log.c.o lib/libfuse3.so.3.12.0.p/compat.c.o lib/libfuse3.so.3.12.0.p/mount.c.o lib/libfuse3.so.3.12.0.p/modules_iconv.c.o -Wl,--as-needed -Wl,--no-undefined -shared -fPIC -Wl,--start-group -Wl,-soname,libfuse3.so.3 -Wl,--version-script,/home/bschubert2/src/libfuse/github/lib/fuse_versionscript -pthread -ldl -lrt -Wl,--end-group
> /usr/bin/ld: lib/libfuse3.so.3.12.0.p/helper.c.o: in function `fuse_main_real':
> /home/bschubert2/src/libfuse/github/build-ubuntu/../lib/helper.c:380: undefined reference to `fuse_loop_mt'
> collect2: error: ld returned 1 exit status
> ```
> 
> That is probably a real issue, as there is no test in fuse.h for **APPLE** and **ULIBC**. But I wonder how this ever worked, I think I had only extended the macro for "FUSE_MAKE_VERSION(3, 12)".

Hi @aakefbs,
I've tried to build and this happens on my pc too. fuse.c.o only contains:
000000000000b4e0 T fuse_loop_mt_31
000000000000b420 T fuse_loop_mt_312
000000000000b480 T fuse_loop_mt_32
But this is similar to the other bug, because using #define's we don't end up having symbol 'fuse_loop_mt()' and now you try to use that symbol in a .c file. Before fuse.h was included before fuse_lowlevel.h so I think that in this case symbol 'fuse_loop_mt()' has never been used, that's why it worked.

Once I have time I'll try to dig more :-)

aakefbs:
Now my question is how this was ever successfully compiling on ulibc? 
And next question is then, do we actually need any of the ifdefs for ULIBC? According to your others tests, any version supports versioned symbols - so we can just remove `__ULIBC__`  ifdefs? And with that, do we actually need `__APPLE__`? I think there is a meson rule that detects darwin and then points to a fuse darwin project?

giuliobenetti:
> Now my question is how this was ever successfully compiling on ulibc?

Honestly I don't know since I don't have a test case :-/ Anyway Buildroot's Autobuilders should catch future build failures and we can find a solution for it. 

> And next question is then, do we actually need any of the ifdefs for ULIBC? According to your others tests, any version supports versioned symbols - so we can just remove `__ULIBC__` ifdefs?

Now yes, because even the oldest toolchain used in Buildroot supports symver. But this is not true for eventual external toolchain that can be used. But again, if something fails then they usually ask for it in Mailing List or they open a bug in BR bugzilla, so I can catch it and in case fix the bug.

> And with that, do we actually need `__APPLE__`? I think there is a meson rule that detects darwin and then points to a fuse darwin project?

I'm not sure about this. And this IMHO should be a separate patch. Maybe we should rely on common HAVE_SYMVER_ATTRIBUTE where we ignore uclibc but we keep __APPLE__ and then let's use it in the code instead of using __UCLIBC__(that should disappear) and __APPLE__. What about that?

aakefbs:
I wouldn't like to mix HAVE_SYMVER_ATTRIBUTE with `__APPLE__`, as HAVE_SYMVER_ATTRIBUTE is also needed for older linux distributions.
I'm going to create two patches, one to remove `__ULIBC__` and another to remove `__APPLE__` as well and will then ask here: https://github.com/osxfuse/osxfuse

giuliobenetti:
> I wouldn't like to mix HAVE_SYMVER_ATTRIBUTE with `__APPLE__`, as HAVE_SYMVER_ATTRIBUTE is also needed for older linux distributions. 

Oops, yes they are 2 different things symver and __APPLE__/__UCLIBC__

> I'm going to create two patches, one to remove `__ULIBC__` and another to remove `__APPLE__` as well and will then ask here: https://github.com/osxfuse/osxfuse

Yes, good idea!

Nikratio:
I don't think we need `__APPLE__`. libfuse does not support OS-X, and OSXFUSE does not share code with FUSE. This was probably cargo-culted at some point.

When it comes to µlibc, I suspect that the answer may simply be that libfuse does not support it. Without symver support, we would have to increase the soversion number with many releases. This is would also affect libc users and is a major pain and is IMO not worth the cost. If we decide to not do this while still claiming to support ulibc, then applications using the dynamic library could randomly crash because of ABI incompatibility, which I do not consider acceptable either.

giuliobenetti:
> I don't think we need `__APPLE__`. libfuse does not support OS-X, and OSXFUSE does not share code with FUSE. This was probably cargo-culted at some point.
> 
> When it comes to µlibc, I suspect that the answer may simply be that libfuse does not support it. Without symver support, we would have to increase the soversion number with many releases. This is would also affect libc users and is a major pain and is IMO not worth the cost. If we decide to not do this while still claiming to support ulibc, then applications using the dynamic library could randomly crash because of ABI incompatibility, which I do not consider acceptable either.

In Buildroot we only have 2 packages that depend on libfuse3:
1. fuse-overlayfs
2. sshfs
and recently default libc has been switched to glibc. So drop libfuse for uclibc should be ok. If someone wants to use it will rely on glibc.

lib/helper.c
```diff
@@ -251,18 +251,6 @@ int fuse_parse_cmdline_30(struct fuse_args *args,
 	return rc;
 }
 
-/**
- * Compatibility ABI symbol for systems that do not support version symboling
- */
-#if (defined(__UCLIBC__) || defined(__APPLE__))
-int fuse_parse_cmdline(struct fuse_args *args,
```
Nikratio, line 258: This means that when building libfuse with UCLIBC, the shared library will no longer have a `fuse_parse_cmdline` symbol. This means that programs that have been compiled against an older version of libfuse will fail when running with the newer shared object.

Am I missing something?

Nikratio (changes requested):

lib/helper.c
aakefbs: Hi Nikolaus, my tired eyes might miss something myself, but this symbol is only defined for __ULIBC__ and __APPLE__ only? At least I had introduced that symbol it for these...

Anyway, given that ulibc (µClibc?) supports version symboling according to @giuliobenetti, and given that Appli is not supported anyway, all these workarounds are not needed. 

Here is the patch I wanted to create weeks ago, https://github.com/aakefbs/libfuse/commit/0cc08aba3fc4e2df644357cffcc1e5fce179ff9c 
I will review tomorrow myself, as it got rather late here. And maybe @giuliobenetti could give it a test first?

And sorry for the late patch, I had been totally occupied for the last weeks with other work (part of it for another fuse topic).

aakefbs: Hmm, actually I'm not sure anymore
http://www.etalabs.net/compare_libcs.html

```
ABI and versioning comparison	musl	uClibc	dietlibc	glibc

Symbol versioning	no	no	no	yes
```

Maybe µClibc did not get updated to -ng, but there are other libc which also do not support it

giuliobenetti: Hi @aakefbs,

sorry for absence too, I am very busy at the moment.
uclibc/-ng supports symver from a version on. That's why I would deal with HAVE_SYMVER_ATTRIBUTE instead of __UCLIBC__. The problem we're dealing with here is if the libc supports or not Symbol versioning.

If we have false positive in testing HAVE_SYMVER_ATTRIBUTE in meson then let's try to improve it.
But if I'm not wrong we also need to keep both:
fuse_session_loop_mt()
fuse_session_loop_mt_312()
because if a binary has been linked with fuse_session_loop_mt_312() with no symver, then we will end up having only
fuse_session_loop_mt() and the runtime linking could fail. Am I correct?

What do you all think about it?

This is not an easy task taking into account the ABI compatibility!

giuliobenetti: Ah, anyway I've started the building of 44 sets of mixed toolchains with buildroot/utils/test-pkg and let's see what comes out.
There is a mix of architectures, gcc versions and libc types and versions.

aakefbs: Thank you!

HAVE_SYMVER_ATTRIBUTE is just to check if the compiler supports "__attribute__((__symver__" or the old asm way should be used, see lib/fuse_misc.h. I don't think my patch to remove __UCLIBC__ and __APPLE__ would make a difference here.

giuliobenetti: @aakefbs no because you assume we always have symver available. That's why I propose to use HAVE_SYMVER_ATTRIBUTE only. But the problem of fuse_session_loop_mt()/_312() I think remain.
@Nikratio can you give us feedback on this and this https://github.com/libfuse/libfuse/pull/698#discussion_r1015980820?

Thank you!

Nikratio: Yes, we need to keep the various compatibility versions of functions around, otherwise programs compiled against old libfuse will fail when loading a new libfuse at runtime.

I think the following course of action makes most sense:

- Refuse to compile libfuse as a shared library if libc does not support symbol versioning (static linking is still possible)
- Rename functions so that the newest implementation does not have a suffix (that way it will be available when compiling a static library)
- Use symbol versioning to map older versions of a symbol to older implementations (functions with a suffix)

Would that work? Or is there some requirement that the "main" symbol (without prefix) is not defined at all? In that case I'd say we should use a preprocessor macro that is evaluated only when compiling statically.

Nikratio: > Hi Nikolaus, my tired eyes might miss something myself, but this symbol is only defined for **ULIBC** and **APPLE** only? At least I had introduced that symbol it for these...

Yes. But that still means that if a program has been compiled against an old fuse version (under ulibc), and is now runtime-linked against a newer shared library (still compiled with ulibc), then it will suddenly fail.

That said, I do not know if there are any such programs. Has it ever been possible to compile libfuse dynamically against ulibc?

giuliobenetti: > Yes, we need to keep the various compatibility versions of functions around, otherwise programs compiled against old libfuse will fail when loading a new libfuse at runtime.
> 
> I think the following course of action makes most sense:
> 
>     * Refuse to compile libfuse as a shared library if libc does not support symbol versioning (static linking is still possible)
> 
>     * Rename functions so that the newest implementation does not have a suffix (that way it will be available when compiling a static library)
> 
>     * Use symbol versioning to map older versions of a symbol to older implementations (functions with a suffix)
> 
> 
> Would that work? Or is there some requirement that the "main" symbol (without prefix) is not defined at all? In that case I'd say we should use a preprocessor macro that is evaluated only when compiling statically.

This is a solution for me.

giuliobenetti: > > Hi Nikolaus, my tired eyes might miss something myself, but this symbol is only defined for **ULIBC** and **APPLE** only? At least I had introduced that symbol it for these...
> 
> Yes. But that still means that if a program has been compiled against an old fuse version (under ulibc), and is now runtime-linked against a newer shared library (still compiled with ulibc), then it will suddenly fail.
> 
> That said, I do not know if there are any such programs. Has it ever been possible to compile libfuse dynamically against ulibc?

Yes, uclibc dynmically  links.

aakefbs: @Nikratio I don't think symbol versioning is an absolute requirement for ABI compatibility. On architechtures that do not support versioned symbols, the function is always redirected at compilation time via a macro to a versioned function name. When the function is made incompatible, a new function with a higher version in the name is created and the old function then just calls into the new one with everything to provide compatibility. I don't see how this could break something and why static compilation would be needed for libc's that do not support versioned symbols.

My assumption till yesterday night just was that the APPLE and UCLIBC macros are not required at all, as APPLE is not supported anyway and UCLIBC seemed to support versioned symbols. Now given the libc table I had found, it is really just gnu-libc that support version symbol, I *think* we should do the other way around - replace the UCLIBC and APPLE macros with a more generic way like HAVE_VERSIONED_SYMBOL and then support more libcs. Personally I actually like musl...

The only interesting part for me is, that I had created a patch last time to artifically test ulibc compilation and that had failed for me. I will check more closely now what needs to be fixed.

giuliobenetti: @aakefbs Are you sure that the #define redirects the function name?
I mean:
```
fuse_session_loop_mt_312()
#define fuse_session_loop_mt() fuse_session_loop_mt_312()
```
should end up with only fuse_session_loop_mt_312 symbol defined and not fuse_session_loop_mt.
So some binary out there will require fuse_session_loop_mt() while other fuse_session_loop_mt_312().
#define only substitute fuse_session_loop_mt() with fuse_session_loop_mt_312() all around, including the binaries linked to
to libfuse since that #define is specified in the header.
Can you try to hint:
```
nm libfuse.x.x.so
```
and grep for both functions? If build for the case where fuse_session_loop_mt_312() is used, *without symver*, I'm not sure you will find fuse_session_loop_mt symbol, but you should find fuse_session_loop_mt_312 symbol instead.

Right?

giuliobenetti: And instead some binary will require fuse_session_loop_mt_312() explicitly because it's defined in .c file.

aakefbs: The ugly part is that other libc also define __GLIBC__ and similar macros *sigh*. Looks like we need to make it mostly a meson option, users need to define. Like here

https://lists.fedorahosted.org/archives/list/lvm2-commits@lists.fedorahosted.org/thread/OGP5IMLH7H6E4KCTB2WTLYGWPSGU6DOY/

UCLIBC can be tested for with a macro, but musl cannot, afaik. At least there needs to be a manual override.

aakefbs: @giuliobenetti It is not about if the symbol is defined or not (it gets defined via the glibc version mechanism), it is about if the caller will ever use the function_name() or if the caller will use function_name_version(). With `#define function_name function_name_version` in an included header file, function_name() is always redirected. This already required for API compatibility (i.e. recompilation with FUSE_USE_VERSION XYZ, when XYZ is not the most recent version).

The entire discussion here started, as the macro redirecting fuse_parse_cmd_line was conflicting with a compat ABI symbol - fuse_parse_cmdline was not versioned so far.

giuliobenetti: @aakefbs Yes, it was what I meant about the define, regarding the symbol you're right because of the FUSE_USE_VERSION XYZ, I only wanted to point that we need all function versions.
It looks a bit like a loop this case :-/

Nikratio:
I'm somewhat confused by the current status of this patch. As-is, we are replacing a preprocessor macro with a static function for all build environments (rather than just Apple and uclibc).

Which issue is this fixing specifically? I thought the problem was specific to uclibc?

Is it possible that the same issue could be fixed by only removing the special #ifdef, and leaving the preprocessor macros as-is?

aakefbs:
I have day off today and tomorrow, gives me a chance to eventually work on this. To my understanding ulibc and others do not have full symbol versioning support, so it all needs to be solved with function name symbols. Patch follows in the evening when kids are in bed :)

giuliobenetti:
> I have day off today and tomorrow, gives me a chance to eventually work on this. To my understanding ulibc and others do not have full symbol versioning support, so it all needs to be solved with function name symbols. 

Exactly.
Sorry both if I’ve disappeared but I’ve been very busy and I still am. If you @aakefbs or @Nikratio can fix it I can bump the package version to the latest in Buildroot and it would great!

> Patch follows in the evening when kids are in bed :)

Nikratio:
The trouble is that I'm not sure what is being fixed anymore. Can you provide some recipe for reproducing the problem?

aakefbs:
Problem with the current branch is that does not compile with ulibc. Stay tuned, almost done.

aakefbs:
Here is the new pull request: https://github.com/libfuse/libfuse/pull/719

@giuliobenetti  would it possible for you test that branch?

giuliobenetti:
> Here is the new pull request: #719
> 
> @giuliobenetti would it possible for you test that branch?

Sure, I'm launching the build test on 44 arch/toolchain combinations and let you know

Nikratio:
Okay, so this pull request doesn't actually fix the problem? In that case I'd like to close it.

giuliobenetti:
Exactly #719 fixes the problem so I close this PR.
Thank you for solving!

Issue #695, opened 2022-08-31
Deadlock with rmdir and rename


spectral54:
The partial locking support in `queue_element_wakeup` can cause a deadlock under the following situation:

- something triggers a two-path queue entry (such as for a `rename` operation) to be the first entry and to partially lock
- an `rmdir` for a common grandparent directory of the two paths in the queue arrives

In this situation, the 'rmdir' attempts to acquire a write lock on the directory it wants to remove. This does not succeed, since the reader lock is already held by the partially-locked rename, so the directory is instead marked as "writer waiting" by adding `TREELOCK_WAIT_OFFSET` (`INT_MIN`) to the `treelock`. Every time afterward that the queue attempts to process sees:
- first queue element (`rename`):
    - one of the paths is already locked, so it's ignored
    - the other path tries to reader lock the directory that `rmdir` wants to remove, but fails because there's a writer waiting
    - (no forward progress on this queue element)
- second queue element (`rmdir`):
    - attempt to lock the directory, it fails due to it already being locked
    - (no forward progress on this queue element due to `treelock` already indicating that there's a writer waiting)
 
I've been able to reproduce this reliably using a modified version of the `hello.c` example
[deadlock.patch.txt](https://github.com/libfuse/libfuse/files/9457305/deadlock.patch.txt), and using the command line (in zsh, but should work in bash too):

```
# Build it and run it
$ mkdir -p /tmp/hello
$ gcc -Wall hello.c `pkg-config fuse3 --cflags --libs` -o hello
$ ./hello /tmp/hello

# Reproduce the issue
$ getfattr -nuser.sleep_5s /tmp/hello/rmthis/foo/file.txt &
$ sleep 3
$ getfattr -nuser.sleep_5s /tmp/hello/rmthis/foo/file.txt &
$ mv /tmp/hello/rmthis/{foo,bar}/file.txt &
$ sleep 3
$ rmdir /tmp/hello/rmthis
```

What this is doing is:
1. create a situation where /rmthis, /rmthis/foo, and /rmthis/foo/file.txt all have _two_ reader locks on them due to the getxattrs [note: on my machine only one `getfattr` call is necessary, but if nothing issues spurious getxattr calls for things like the security xattrs, the `sleep 3` and extra `getfattr` should set up things to execute in the proper order, but that's untested]
[...]

PR #699, opened 2022-09-14
Remove partial locking of paths

spectral54:
As described in https://github.com/libfuse/libfuse/issues/695, partial locking of paths can cause a deadlock. Partial locking was added to prevent starvation, but it's unclear what specific cases of starvation were of concern. As far as I was able to determine, since we support reader locks, this means that to starve the queue element, we'd need a constant stream of queued requests that lock the path for write. Write locks are used when the element is being removed, so this stream of requests that starve the `rename` or `lock` operations seems unlikely.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -1091,6 +1091,27 @@ static void queue_element_unlock(struct fuse *f, struct lock_queue_element *qe)
 	}
 }
 
+static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
+			 fuse_ino_t nodeid2, const char *name2,
+			 char **path1, char **path2,
+			 struct node **wnode1, struct node **wnode2)
+{
+	int err;
+
+	/* FIXME: locking two paths needs deadlock checking */
+	err = try_get_path(f, nodeid1, name1, path1, wnode1, true);
+	if (!err) {
+		err = try_get_path(f, nodeid2, name2, path2, wnode2, true);
+		if (err) {
+			struct node *wn1 = wnode1 ? *wnode1 : NULL;
+
+			unlock_path(f, nodeid1, wn1, NULL);
+			free(*path1);
+		}
+	}
+	return err;
+}
+
 static void queue_element_wakeup(struct fuse *f, struct lock_queue_element *qe)
 {
 	int err;
@@ -1306,27 +1327,6 @@ static int check_dir_loop(struct fuse *f,
 	return 0;
 }
 #endif
-
-static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
-			 fuse_ino_t nodeid2, const char *name2,
-			 char **path1, char **path2,
-			 struct node **wnode1, struct node **wnode2)
-{
-	int err;
-
-	/* FIXME: locking two paths needs deadlock checking */
-	err = try_get_path(f, nodeid1, name1, path1, wnode1, true);
-	if (!err) {
-		err = try_get_path(f, nodeid2, name2, path2, wnode2, true);
-		if (err) {
-			struct node *wn1 = wnode1 ? *wnode1 : NULL;
-
-			unlock_path(f, nodeid1, wn1, NULL);
-			free(*path1);
-		}
-	}
-	return err;
-}
 
 static int get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
 		     fuse_ino_t nodeid2, const char *name2,
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -80,8 +80,6 @@ struct lock_queue_element {
 	char **path2;
 	struct node **wnode2;
 	int err;
-	bool first_locked : 1;
-	bool second_locked : 1;
 	bool done : 1;
 };
 
@@ -1075,22 +1073,6 @@ static int try_get_path(struct fuse *f, fuse_ino_t nodeid, const char *name,
 	return err;
 }
 
-static void queue_element_unlock(struct fuse *f, struct lock_queue_element *qe)
-{
-	struct node *wnode;
-
-	if (qe->first_locked) {
-		wnode = qe->wnode1 ? *qe->wnode1 : NULL;
-		unlock_path(f, qe->nodeid1, wnode, NULL);
-		qe->first_locked = false;
-	}
-	if (qe->second_locked) {
-		wnode = qe->wnode2 ? *qe->wnode2 : NULL;
-		unlock_path(f, qe->nodeid2, wnode, NULL);
-		qe->second_locked = false;
-	}
-}
-
 static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
 			 fuse_ino_t nodeid2, const char *name2,
 			 char **path1, char **path2,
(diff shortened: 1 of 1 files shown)


lib/fuse.c
```diff
@@ -1075,26 +1073,30 @@ static int try_get_path(struct fuse *f, fuse_ino_t nodeid, const char *name,
...
-	if (qe->first_locked) {
-		wnode = qe->wnode1 ? *qe->wnode1 : NULL;
-		unlock_path(f, qe->nodeid1, wnode, NULL);
-		qe->first_locked = false;
-	}
-	if (qe->second_locked) {
-		wnode = qe->wnode2 ? *qe->wnode2 : NULL;
-		unlock_path(f, qe->nodeid2, wnode, NULL);
-		qe->second_locked = false;
+	/* FIXME: locking two paths needs deadlock checking */
```
Nikratio, line 1083: This sounds like something that we may want to address before merging?

Nikratio (changes requested):
Let's also be a bit more specific in the commit message: with this change, we are explicitly accepting that write requests through the high level API may never complete as long as there is a constant stream of read requests. However, we think that this is not as bad as having a deadlock for some operations. Let's also include the example to reproduce the deadlock (better to have it in the commit message than just on Github).

Nikratio:
Thanks for working on this!

lib/fuse.c
Nikratio: Would it be a solution to e.g. always lock in alphabetical order?

aakefbs:
I need to find the time to go through this code. Wouldn't it be a solution here to detect a lock conflict, bail out and then to take an exclusive lock? Like

```
bool exclusive = false;

again:
if (exclusive) 
    pthread_rwlock_wrlock(&f->rwlock);
else
   pthread_rwlock_rdlock(&f->rwlock);
...
<all existing code>

if (err == -EAGAIN) {
    if (!exclusive)
        exclusive = true;
        reverse_everything_that_was_done_for_existing_locking()
        goto again;
    else {
        fuse_log(FUSE_LOG_ERR, "Bug, EAGAIN, although exclusive");
        // hard to handle - try again or return an error ?
    }
}

pthread_rwlock_unlock(&f->rwlock);
```

So basically it tries to do as much as possible in parallel, but if a conflict is detected it switches to an exclusive lock. The hard part is probably that right now everything is mutex protected, while I'm suggesting an rw-lock.

That being said, I'm not familiar at all with that code and the hand coded locking (node->treelock) is interesting.

aakefbs:
> This would not solve the current issue, as the current issue is entirely in the 

It is waiting in a pthread-cond-wait. I meant to avoid doing that and to instead take a global exclusive write lock.

spectral54:
> > This would not solve the current issue, as the current issue is entirely in the
> 
> It is waiting in a pthread-cond-wait. I meant to avoid doing that and to instead take a global exclusive write lock.

I'm still not understanding the proposal. I believe that the pthread locking is essentially irrelevant to the issue. The pthread mutex is guarding ~everything about the high level data structures, but is _not_ actually doing the "locking" of the nodes themselves - that's managed via the `node->treelock` variable (which is guarded by the mutex, but is not the mutex itself).

aakefbs:
I mean to introduce a new f->rw_lock. Typically the code would take a read-lock. Once it detects that there is a lock conflict - i.e. wants to put itself onto a wait queue it would not do that, but bail out, take a write lock and then try again. If two sides switch to a write lock, it will get serialized. Disadvantage would be a global lower performance when write lock is taken. As I said, I find the manually code locking interesting. My guess is that this old code, when atomics and read-write locks didn't exist yet in user space.

spectral54:
> I mean to introduce a new f->rw_lock. Typically the code would take a read-lock. Once it detects that there is a lock conflict - i.e. wants to put itself onto a wait queue it would not do that, but bail out, take a write lock and then try again. If two sides switch to a write lock, it will get serialized. Disadvantage would be a global lower performance when write lock is taken. As I said, I find the manually code locking interesting. My guess is that this old code, when atomics and read-write locks didn't exist yet in user space.

I still don't understand how the proposal helps avoid the deadlock in this case. If partial locking remains, I don't understand how changing the locking of the `fuse` struct (and the data it contains/manages) will change the behavior here at all, except by possibly meaning that we could get rid of `lockq` entirely (which ends up removing the partial locking as a side effect).

If the proposal is to hold _any_ filesystem-wide lock (whether in shared or exclusive mode) for the entirety of the operation (calling into the filesystem), it would likely have very negative effects on mutli-threaded performance either via the course of normal interactions with the filesystem, or by enabling DoS opportunities. It also means that re-entrant filesystems will be much more prone to deadlock.

To document the current locking behavior:
1. acquire `f->lock` (exclusively)
2. manipulate `node->treelock`
    1. If that fails, add element to `f->lockq`, create a condvar to wake up when it's potentially lockable, wait on the condvar, then try manipulating `node->treelock` again. Loop until success.
3. drop `f->lock`
4. call into filesystem  <-- `f->lock` is NOT held here. 
5. acquire `f->lock`
6. manipulate `node->treelock` (should never fail)
7. drop `f->lock` and possibly wake up waiters in lockq

I could envision a possible switch from instead of using `int node->treelock` we use actual pthread mutexes, but I don't see any significant benefit to doing such a transition: we already have a lock that protects these variables from multi-threaded accesses, adding pthread overhead for what is achievable using just a raw `int` seems likely to be detrimental.

spectral54:
I've discovered an issue with this change as-is, I'm working on understanding/resolving these issues now.

Nikratio:
@spectral54 There is nothing else you need to do, in principle I can merge this as-is. I just did not have the time to review this properly.

@aakefbs What's your take on this in it's current form?

aakefbs:
Can you give me a few days more to review please? Latest by Friday I should find the time, maybe already tomorrow.

spectral54:
Anything needed from me?

Nikratio:
@aakefbs Did you have a chance to review this?

lib/fuse.c
```diff
@@ -1075,26 +1073,30 @@ static int try_get_path(struct fuse *f, fuse_ino_t nodeid, const char *name,
 	return err;
 }
 
-static void queue_element_unlock(struct fuse *f, struct lock_queue_element *qe)
+static int try_get_path2(struct fuse *f, fuse_ino_t nodeid1, const char *name1,
```
Nikratio, line 1076: Could you move definition of this function to where it originally was (further down) to simplify the diff?

aakefbs:
I'm really sorry, didn't find the time yet, but I definitely promise, I will do till Sunday.

lib/fuse.c
spectral54: Only by forward declaring it, and forward declaring (and taking the permanent complexity hit, however slight) just to simplify the diff during review feels weird to me.

I split out the movement of this function to an earlier commit, though I know that github's UI doesn't make that all that obvious. The second commit in the stack should be just the actual changes, if you want to review the two commits separately. I can also send a separate PR instead of a multi-commit PR if that would help.

aakefbs: For me it is ok, I see it as two different patches. I just really didn't find the time to review, as I wrote before, will definitely do this week.

aakefbs:
Just a heads up, I'm still on it.

spectral54:
In case it's helpful: we've had this deployed for months now in at least 3 different filesystems on our developer workstations and there have been no reports of issues from the patch or any further reports of deadlocks.

PR #710, opened 2022-11-14
libfuse custom communication interface

TSonono:
libfuse can be used without having a mountpoint. This can be useful in different cases, for example where you might want to use libfuse in an IPC manner with unix domain sockets. Also for potential server communication. So for instance, instead of using `fuse_session_mount` / `fuse_session_unmount`, one would call `fuse_custum_session_fd` with a fd created by the consumer of the API.

I understand if such functionality is out of scope for this library, but I'll post this PR to see what you think

### DIFF
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -1955,6 +1955,25 @@ struct fuse_session *fuse_session_new(struct fuse_args *args,
 				      size_t op_size, void *userdata);
 
 /**
+ * Set a file descriptor for the session.
+ *
+ * This function can be used if you want to have a custom communication
+ * interface instead of using a mountpoint. In practice, this means that instead
+ * of calling fuse_session_mount() and fuse_session_unmount(), one could call
+ * fuse_custom_session_fd() where fuse_session_mount() would have otherwise been
+ * called.
+ *
+ * This function does not open or close any file descriptors, meaning it is the
+ * responsibility of the caller to provide a valid file descriptor, such as a
+ * accepted socket. It is also the responsibility to close the socket.
+ *
+ * @param se session object
+ * @param fd file descriptor for the session
+ *
+ **/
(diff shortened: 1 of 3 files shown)


Nikratio:
It's not such an invasive change, so I'm happy to consider this. However, could you please:

- get rid of the whitespace changes
- add more documentation (at which point does the new function need to be called? Do we still need to termination functions? Is the fd open at the end of the main loop? Is this compatible with the high level API?)

And lastly, is there anything out there that uses this / wants to use this or is this purely speculative?

TSonono:
Thanks for the swift reply.

In the passthrough_ll example, `fuse_custum_session_fd` would be called instead of `fuse_session_mount` and there would be no need to call `fuse_session_umount`. The fd would have to be closed by the consumer of the API, as done at the in passthrough_ll. The termination function should still be called (except for `fuse_session_unmount`).

I have only tried it with the low level API, but I don't see why it wouldn't work with the high level API. I think we just would have to make it possible `fuse_main_real` to not call `fuse_mount` and `fuse_umount` and instead call `fuse_custum_session_fd` where `fuse_mount` is called. This could be conditional based on the arguments passed by the API consumer. I can try it out to verify.

I'm considering the options for a functionality that I'm trying to implement in a product working on. It requires a process to forward FUSE requests to the host OS. With this change, there is a possibility I would be using this library for that. In reality, what would have been more convenient would be the possibility to override the `fuse_send_msg` function. Then, I could just link this library to my application without needing any fd communication. But I felt that that would a very invasive change that probably does not make sense for this library. And using fd:s is most likely sufficient for my use case.

I'll remove the whitespace changes.

TSonono:
I think to make it work with the high level API, we would have to put a new field in the `fuse_cmdline_opts`, and I'm not sure that makes too much sense, since it makes most sense to have created this socket in the program itself.

I added some more documentation to the function.

PR #713, opened 2022-11-30
Make it work even if max_idle_threads is set to 0

zhsgao:
It may happen that none of the worker threads are running if max_idle_threads is set to 0 although few people will do this. Adding a limit of keeping at least one worker thread will make our code more rigorous.

Signed-off-by: Zhansong Gao <zhsgao@hotmail.com>

### DIFF
--- a/lib/fuse_loop_mt.c
+++ b/lib/fuse_loop_mt.c
@@ -179,7 +179,7 @@ static void *fuse_do_work(void *data)
 		 * is indeed a good reason to destruct threads it should be done
 		 * delayed, a moving average might be useful for that.
 		 */
-		if (mt->max_idle != -1 && mt->numavail > mt->max_idle) {
+		if (mt->max_idle != -1 && mt->numavail > mt->max_idle && mt->numworker > 1) {
 			if (mt->exit) {
 				pthread_mutex_unlock(&mt->lock);
 				return NULL;


aakefbs (approved):

aakefbs:
Looks good to me.

jpandre:
`if (mt->max_idle != -1 && mt->numavail > mt->max_idle && mt->numworker > 1)` 

or ?

`if (mt->max_idle != -1 && mt->numavail > mt->max_idle && mt->numworker >= 1)`

aakefbs:
@jpandre You don't want to exit your last thread? 

@zhsgao I would suggest to update ceph to the latest API, to avoid stopping threads at all  - start/stop is expensive.

zhsgao:
> @jpandre You don't want to exit your last thread?
> 
> @zhsgao I would suggest to update ceph to the latest API, to avoid stopping threads at all - start/stop is expensive.

Yes, I agree. The API version used by Ceph is 3.5 now and the latest version of the library can be used.
I have tested ceph-fuse with libfuse 3.12, it is ok after I set max_idle_threads to the correct value.

aakefbs:
What I actually mean is that with the 3.12 API you shouldn't use max_idle_threads at all anymore, but you should use `max_threads`.

zhsgao:
> What I actually mean is that with the 3.12 API you shouldn't use max_idle_threads at all anymore, but you should use `max_threads`.

Ceph uses the low-level API now. I change the value of 'FUSE_USE_VERSION' from 35 to 312 but I get a compile error, 'fuse_loop_config' is declared but not defined.

aakefbs:
Yes, please read the changelog - this is why I had said fuse API and not version. You will need to change your code a bit.

zhsgao:
> Yes, please read the changelog - this is why I had said fuse API and not version. You will need to change your code a bit.

Thanks very much. I know how to use the latest API now. I will try to use it later.

PR #714, opened 2022-12-19
Greater control of custom IO

TSonono:
Follow-up to  https://github.com/libfuse/libfuse/pull/710

The IO for FUSE requests and responses can be further customized by allowing to write custom functions for reading/writing the responses. This includes overriding the splice IO. Also extended the hello_ll example (with #ifdefs) to provide an example of utilizing the custom IO (and a test for it). When no custom IO is provided, libfuse will behave the same as it did before any changes in this PR were made.

As it turns out, the changes in https://github.com/libfuse/libfuse/pull/710 are not enough to provide the user necessary control of the IO handling of the request and response data. Different types of file descriptor require different mechanisms of IO interaction. For example, some file descriptor types have boundaries (SOCK_DGRAM, EOF, etc...), while other types of might be unbounded (SOCK_STREAMS, ...). For unbounded communication, one call to `read` does not necessarily return all the data expected, especially in cases where the request are large. This requires further calls to `read` in a loop. You also would have to read the header of the FUSE request first, and then read the remaining packet data to ensure that you get all of the data in the request.

The `fuse_custom_session_fd()` (renamed `fuse_session_custom_io()` in this PR) function has not been part or a libfuse release yet (only part of the master branch), so my hope is that it's not considered a breaking change to modify that function.

### DIFF
--- a/example/hello_ll.c
+++ b/example/hello_ll.c
@@ -20,6 +20,15 @@
 
 #define FUSE_USE_VERSION 34
 
+#ifdef IO_UDS_STREAMS
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <fuse_kernel.h>
+#endif
+
 #include <fuse_lowlevel.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -161,19 +170,165 @@ static const struct fuse_lowlevel_ops hello_ll_oper = {
 	.read		= hello_ll_read,
 };
 
+#ifdef IO_UDS_STREAMS
+static int create_socket(const char *socket_path) {
+	struct sockaddr_un addr;
+
+	if (strnlen(socket_path, sizeof(addr.sun_path)) >=
+		sizeof(addr.sun_path)) {
+		printf("Socket path may not be longer than %lu characters\n",
+			 sizeof(addr.sun_path) - 1);
+		return -1;
+	}
+
+	if (remove(socket_path) == -1 && errno != ENOENT) {
+		printf("Could not delete previous socket file entry at %s. Error: "
+			 "%s\n", socket_path, strerror(errno));
+		return -1;
+	}
+
+	memset(&addr, 0, sizeof(struct sockaddr_un));
+	strcpy(addr.sun_path, socket_path);
+
+	int sfd = socket(AF_UNIX, SOCK_STREAM, 0);
+	if (sfd == -1) {
+		printf("Could not create socket. Error: %s\n", strerror(errno));
+		return -1;
+	}
+
+	addr.sun_family = AF_UNIX;
+	if (bind(sfd, (struct sockaddr *) &addr,
+		   sizeof(struct sockaddr_un)) == -1) {
+		printf("Could not bind socket. Error: %s\n", strerror(errno));
+		return -1;
+	}
+
+	if (listen(sfd, 1) == -1)
+		return -1;
+
+	printf("Awaiting connection on socket at %s...\n", socket_path);
+	int cfd = accept(sfd, NULL, NULL);
+	if (cfd == -1) {
+		printf("Could not accept connection. Error: %s\n",
+			 strerror(errno));
+		return -1;
+	} else {
+		printf("Accepted connection!\n");
+	}
+	return cfd;
+}
+
+static ssize_t stream_writev(int fd, struct iovec *iov, int count,
+                             void *userdata) {
+	(void)userdata;
+
+	ssize_t written = 0;
+	int cur = 0;
+	for (;;) {
+		written = writev(fd, iov+cur, count-cur);
+		if (written < 0)
+			return written;
+
+		while (cur < count && written >= iov[cur].iov_len)
+			written -= iov[cur++].iov_len;
+		if (cur == count)
+			break;
+
+		iov[cur].iov_base = (char *)iov[cur].iov_base + written;
+		iov[cur].iov_len -= written;
+	}
+	return written;
+}
+
+
+static ssize_t readall(int fd, void *buf, size_t len) {
+	size_t count = 0;
+
+	while (count < len) {
+		int i = read(fd, (char *)buf + count, len - count);
+		if (!i)
+			break;
+
+		if (i < 0)
+			return i;
+
+		count += i;
+	}
+	return count;
+}
+
+static ssize_t stream_read(int fd, void *buf, size_t buf_len, void *userdata) {
+    (void)userdata;
+
+	int res = readall(fd, buf, sizeof(struct fuse_in_header));
+	if (res == -1)
+    	return res;
+
+
+    uint32_t packet_len = ((struct fuse_in_header *)buf)->len;
+    if (packet_len > buf_len)
+    	return -1;
+
+    int prev_res = res;
+
+    res = readall(fd, (char *)buf + sizeof(struct fuse_in_header),
+                  packet_len - sizeof(struct fuse_in_header));
+
+    return  (res == -1) ? res : (res + prev_res);
+}
+
+static ssize_t stream_splice_send(int fdin, __off64_t *offin, int fdout,
+					    __off64_t *offout, size_t len,
+                                  unsigned int flags, void *userdata) {
+	(void)userdata;
+
+	size_t count = 0;
+	while (count < len) {
+		int i = splice(fdin, offin, fdout, offout, len - count, flags);
+		if (i < 1)
+			return i;
+
+		count += i;
+	}
+	return count;
+}
+
+static void fuse_cmdline_help_uds(void)
+{
+	printf("    -h   --help            print help\n"
+	       "    -V   --version         print version\n"
+	       "    -d   -o debug          enable debug output (implies -f)\n");
+}
+#endif
+
 int main(int argc, char *argv[])
 {
 	struct fuse_args args = FUSE_ARGS_INIT(argc, argv);
 	struct fuse_session *se;
 	struct fuse_cmdline_opts opts;
+#ifdef IO_UDS_STREAMS
+	const struct fuse_custom_io io = {
+		.writev = stream_writev,
+		.read = stream_read,
+		.splice_receive = NULL,
+		.splice_send = stream_splice_send,
+	};
+	int cfd = -1;
+#else
 	struct fuse_loop_config config;
+#endif
 	int ret = -1;
 
 	if (fuse_parse_cmdline(&args, &opts) != 0)
 		return 1;
 	if (opts.show_help) {
+#ifdef IO_UDS_STREAMS
+		printf("usage: %s [options]\n\n", argv[0]);
+		fuse_cmdline_help_uds();
+#else
 		printf("usage: %s [options] <mountpoint>\n\n", argv[0]);
 		fuse_cmdline_help();
+#endif
 		fuse_lowlevel_help();
 		ret = 0;
 		goto err_out1;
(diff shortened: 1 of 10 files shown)


Nikratio:
Thanks for following up on this.

These changes are much more invasive than what was done in #710, and make the code more complex. I think to justify something like this, we need a stronger case than you "considering the options for a functionality that I'm trying to implement in a product I'm working on" (quoting from #710).

I think the better way forward is to revert #710 and postpone this until a clear and open-sourced use-case is present.

Thoughts?

TSonono:
Hi again, and thanks for the reply.

The changes introduced in this PR are indeed more invasive than the ones in https://github.com/libfuse/libfuse/pull/710. I have tried to ensure that the changes do not impact the current flow of the code in any meaningful way. I do agree that the previous PR should be reverted, since it is not complete. Sorry for submitting it before stress/corner testing the implementation.

I work at Intel, and we currently have a use case that requires the proposed changes in the patch. Let me give some background to how I am using libfuse (with this patch). At Intel I work with their full system simulator, [Simics](https://www.intel.com/content/www/us/en/developer/articles/tool/simics-simulator.html). We are adding support for [Virtio File System](https://virtio-fs.gitlab.io/) to Simics. VirtioFS is a device which has drivers in the Linux kernel, and allows for sharing a directory from a host system, to a simulated guest system, using FUSE. Essentially, in our case, the guest operating system would be sending FUSE requests to Simics using the VirtioFS device. The VirtioFS device would then forward these requests to a libfuse-enabled daemon over UNIX domain sockets. The response is then sent back to the guest.

We have created a VirtioFS device and a libfuse-enabled daemon and have demonstrated that a guest system can boot with a root file system that resides in a VirtioFS device. We believe that this patch would also allow other open-source projects to achieve similar goals with libfuse. Our other alternative would be to maintain a patched version whose source code is shipped with the Simics product, but we would rather use libfuse ‘as is’.

Simics itself is not open source,  but it is released publicly [Simiccs](https://www.intel.com/content/www/us/en/developer/articles/tool/simics-simulator.html), which is free to install for anyone. Many devices that are implemented in Simics, have their source code shipped as part of Simics. This would include the VirtioFS device, as well as the libfuse-enabled daemon.

Hopefully this use case can justify this patch being a part of the upstream libfuse repository.

Best regards,
Tofik

Nikratio:
Thank you for the additional background, Tofik. I have a few questions to better understand this, but could you already rebase this pull request so that it's fully functional (i.e., merge it with the now rolled-back #710)?

Nikratio:
A few questions to make sure I understand correctly:

- I believe VirtioFS uses the FUSE communication protocol, a virtio driver in the guest system (not related to the FUSE driver) and a virtiofs daemon in the host system (written in Rust, not related to libfuse). Is that correct?

- Do I understand correctly that Simics is a replacement for e.g. qemu (to run the guest system), and wants to run a FUSE daemon that communicates on a pre-opened Unix domain socket instead of opening /dev/fuse?

- If so, wouldn't it make more sense to call a virtofs daemon rather than a libfuse daemon?

- Isn't it already possible to boot a guest system from a VirtioFS root device using qEMU/KVM? Why doesn't the same work with Simics?

TSonono:
Thanks for your questions, I'll do my best to answer them.

>I believe VirtioFS uses the FUSE communication protocol, a virtio driver in the guest system (not related to the FUSE driver) and a virtiofs daemon in the host system (written in Rust, not related to libfuse). Is that correct?
>but could you already rebase this pull request so that it's fully functional (i.e., merge it with the now rolled-back https://github.com/libfuse/libfuse/pull/710)?

Done


Yes, in the QEMU case, the virtiofs daemon running in the host is written in Rust and has its own implementation of the FUSE protocol.

>Do I understand correctly that Simics is a replacement for e.g. qemu (to run the guest system), and wants to run a FUSE daemon that communicates on a pre-opened Unix domain socket instead of opening /dev/fuse?

Yes, that is exactly right

>If so, wouldn't it make more sense to call a virtofs daemon rather than a libfuse daemon?

Yes, technically that makes more sense

> Isn't it already possible to boot a guest system from a VirtioFS root device using qEMU/KVM? Why doesn't the same work with Simics?

It is indeed already possible in QEMU, using the aforementioned virtiofs daemon written in Rust. The are multiple reasons that it dosn't work with Simics. The virtiofs device in QEMU is written for QEMU and uses the QEMU API. It cannot be used in Simics. One could then ask if one could create a device in Simics that could use the aforementioned virtiofs daemon written Rust. The reason this approach was not chosen, is that there are several layers of complexity to tackle to make a virtiofs device compatible with that virtiofs daemon. It uses shared memory to communicate the FUSE requests from the virtiofs device in QEMU to the daemon. It also uses a protocol, [vhost](https://qemu.readthedocs.io/en/latest/interop/vhost-user.html), to setup the daemon and the shared memory properties. This can of course result in higher performance, and considering the use cases of the virtiofs device for qemu/kvm, this might make sense. However, if you just want to have a daemon that can handle FUSE requests, without the aforementionend complexities, being able to simply pass the requests using unix domain sockets makes things much easier. Furthermore, with this libfuse approach, you could write a daemon that implements your own file system, where as the Rust virtiofs daemon only implements a passthrough system for a selected directory on the host.

As a note, this feature also opens up for more than the exact use case of using virtiofs to boot from a "remote" root file system / sharing files to a guest. It can also be used can be used to create a file system and expose it using different io interfaces, in addition to unix domain sockets. I think this is a unique feature that perhaps someone else might be interested in the future.

example/hello_ll.c
```diff
@@ -20,6 +20,15 @@
 
 #define FUSE_USE_VERSION 34
 
+#ifdef IO_UDS_STREAMS
```
Nikratio, line 23: Can we leave hello_ll.c as-is (it's meant to be the simplest possible example) and instead add a new example that demonstrates how to use a custom fd? I know you wrapped it in an ifdef, but that doesn't help the reader all that much...

include/fuse_lowlevel.h
```diff
@@ -1954,6 +1968,36 @@ struct fuse_session *fuse_session_new(struct fuse_args *args,
...
+/**
+ * Set a file descriptor for the session.
+ *
+ * This function can be used if you want to have a custom communication
+ * interface instead of using a mountpoint. In practice, this means that instead
+ * of calling fuse_session_mount() and fuse_session_unmount(), one could call
+ * fuse_session_custom_io() where fuse_session_mount() would have otherwise been
+ * called.
+ *
+ * In `io`, implementations for read and writev MUST be provided. Otherwise -1
```
Nikratio, line 1980: Why?

lib/fuse_lowlevel.c
```diff
@@ -186,8 +186,13 @@ static int fuse_send_msg(struct fuse_session *se, struct fuse_chan *ch,
 		}
 	}
 
-	ssize_t res = writev(ch ? ch->fd : se->fd,
-			     iov, count);
+	ssize_t res;
+	if (se->io != NULL)
```
Nikratio, line 190: Can we instead test for `se->io != NULL && se->io->writev !+ NULL` to rule out a potential bug and to increase flexibility?

lib/fuse_lowlevel.c
```diff
@@ -3009,6 +3039,36 @@ struct fuse_session *fuse_session_new(struct fuse_args *args,
...
+	if (fd < 0) {
+		fuse_log(FUSE_LOG_ERR, "Invalid file descriptor value %d passed to "
+			"fuse_session_custom_io()\n", fd);
+		return -EBADF;
+	}
+	if (io == NULL) {
+		fuse_log(FUSE_LOG_ERR, "No custom IO passed to "
+			"fuse_session_custom_io()\n");
+		return -EINVAL;
+	} else if (io->read == NULL || io->writev == NULL) {
```
Nikratio, line 3054: I'm worried about this invariant being safe. We're checking once when calling the function, but later changes to the code may not be aware of it, or someone may even modify the structure after the function has been called. So if we can find a way to avoid this, that'd be great.

Nikratio (changes requested):
Thanks for the extra explanations. The code also looks well written and has a test case, so I'm okay merging it. If you could consider a few minor changes per my other comments?

example/hello_ll.c
TSonono: That's fair. Is it ok if I make a new example that copies a lot of the code from hello_ll.c? Since I think extracting common code between hello_ll.c and this new example I will create to a separate file also makes the hello_ll.c example less clear

include/fuse_lowlevel.h
TSonono: If the user provides their own file descriptor, we can't guarantee that the default behavior of the io operations (the `read`, `writev`, `splice`) made in libfuse will function properly. Therefore, I think it's safest if we enforce the user to implement these io operations if they are the ones providing file descriptors

lib/fuse_lowlevel.c
TSonono: At the end of this function, the passed `fuse_custom_io` structure is actually copied (`*se->io = *io;`), so changes to the passed structure after this function is called, should have no impact.

Maybe `memcpy(se->io, io, sizeof(io))` would have been clearer instead of `*se->io = *io;`?

lib/fuse_lowlevel.c
TSonono: I considered that option as well, and if you'd prefer it, I wouldn't mind changing it. I'll give you the reason I omitted and let you decide the way forward

Given the condition in `fuse_session_custom_io`, which requires `writev` to be implemented, there is no way of `se->io->writev` being `NULL` if `se->io != NULL`. This is unless the user starts the session loop even though `fuse_session_custom_io` did not return 0, which would be a bug. So essentially I omitted the `se->io->writev != NULL` to indicate to the next person reading the code that there is no way of `se->io->writev` being `NULL` if  `se->io != NULL`.

example/hello_ll.c
Nikratio: Yes, having code duplication between the two examples is fine.

include/fuse_lowlevel.h
Nikratio: Hmm. ok.

lib/fuse_lowlevel.c
Nikratio: Ok, then let's just add a comment here that says that fuse_session_custom_io guarantees that if io != NULL, we also have io->writev != NULL.

include/fuse_lowlevel.h
aakefbs: I guess the explanation should be added as comment somewhere or at least in the commit message?

lib/fuse_lowlevel.c
```diff
@@ -2772,6 +2788,9 @@ int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
 	if (se->conn.proto_minor < 14 || !(se->conn.want & FUSE_CAP_SPLICE_READ))
 		goto fallback;
 
+	if (se->io != NULL && se->io->splice_receive == NULL)
+		goto fallback;
```
aakefbs, line 2792: Can you already ensure this in fuse_lowlevel.c do_init()? So that it does not need to check here all the time?

include/fuse_lowlevel.h
TSonono: Sure, I'llcan add it as a comment in the relevant if-block

lib/fuse_lowlevel.c
TSonono: The same semantics is used for all the io operation calls (line 820, 2814 and 2907), should I comment it at all places?

lib/fuse_lowlevel.c
TSonono: You're right. I'll remove these checks for both read/write and handle it by setting the `conn->capable` field accordingly in `do_init()`

lib/fuse_lowlevel.c
Nikratio: All places would be best, I think.

TSonono:
Thanks for accepting this patch. I assume libfuse does not have a release schedule. May I inquire for when one could expect a new release of libfuse?

Issue #721, opened 2023-01-10
Support for external modules always causes Segmentation fault


mrvn:
I have a libfuse module named "used" that keeps track of files being used via unionfs to generate a initramfs with only files accessed during a test run. The module worked fine with fuse2 but after adapting it to fuse3 it just segfaults on load.

The code is based on the example https://github.com/libfuse/libfuse/blob/master/lib/modules/subdir.c and for debug purposes I start the factory function like this:
```
static struct fuse_fs *used_new(struct fuse_args *args,
				  struct fuse_fs *next[]) {
  struct fuse_fs *fs;
  struct used *d;
  fuse_log(FUSE_LOG_ERR, "fuse-used: used_new()\n");
  ....
}

FUSE_REGISTER_MODULE(used, used_new);
```

I compile the module with debug infos:
```
% file /usr/lib/x86_64-linux-gnu/libfusemod_used.so
/usr/lib/x86_64-linux-gnu/libfusemod_used.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=53078773b3db0a877796a3c6725ee247adf8b797, with debug_info, not stripped
```
but when I run it I get:

```
% gdb unionfs
GNU gdb (Ubuntu 12.1-0ubuntu1~22.04) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from unionfs...
(No debugging symbols found in unionfs)
(gdb) r -d -o allow_other -o default_permissions -o dev -o kernel_cache -o noforget -o noinitgroups -o cow -o modu
[...]

PR #722, opened 2023-01-12
Dereference fuse_module_factory_t symbol before assignment

mrvn:
When loading an external module the `fuse_load_so_module()` function assigns the address of the modules factory **symbol** to the `fuse_module_factory_t` function pointer. The pointer then points into the .data section of the module where the address of the factory function is stored. Executing the factory then leads to a segfault as the .data section is not executable.

This patch casts the symbol returned by `dlsym()` to the type it actually is and dereferences it before assigning it to the local function pointer to get the actual address of the factory function.

Fixes: #721

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -249,7 +249,7 @@ static int fuse_load_so_module(const char *module)
 	int ret = -1;
 	char *tmp;
 	struct fusemod_so *so;
-	fuse_module_factory_t factory;
+	fuse_module_factory_t *factory;
 
 	tmp = malloc(strlen(module) + 64);
 	if (!tmp) {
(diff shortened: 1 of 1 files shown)


bsbernd:
Looks right and is basically a revert of 8ec7fd92eb5f41f3baffba91a397296afcd926ab - can you mention this in the commit message? And would be good if @Nikratio could point of detailed reasons for that commit.
Actually, you are also introducing a dereference of the symbol again, why that?  Not just the pointer?

factory = (fuse_module_factory_t*)dlsym(so->handle, tmp);

mrvn:
Because `factory` is a `fuse_module_factory_t` and not a `fuse_module_factory_t*`. The code makes a copy of the function pointer.

Looking at 8ec7fd9 and the comment. It actually would be better to use `fuse_module_factory_t*` because `dlsym` can indeed return NULL and that would the get dereferenced unconditionally.

bsbernd:
Looks good to me. In general missing is a test...

mrvn:
I'm afraid I'm not familiar with meson and ninja to whip that up in a minute. You need a fuse filesystem and a module for this. Looks like the test directory contains 2 test filesystems so one would only have to add a module by e.g. copying and renaming subdir.c.

Nikratio:
Sorry, I can't help much here. If completely forgotten the context of the prior change.

I'll this pull request on faith since you both seem to be an agreement. I agree that a test would be nice, but this shouldn't block us from fixing the crasah.

Issue #724, opened 2023-01-25
HAVE_LIBC_VERSIONED_SYMBOLS not set in installed headers


amesser:
When libfuse detects symver support during build, it sets define HAVE_LIBC_VERSIONED_SYMBOLS for the build. However, when installing headers in the system, these headers do not know about that. Thus applications build against libfuse later, will not have this define set and may show unexpected behavior.

It would be better to use more project specific define name which is also placed in some "config.h" header file to be incoporated in development headers.

bsbernd:
Hmm right, I didn't notice that config.h didn't get included into header files (and is also not distributed).

Nikratio:
@bsbernd Would you have time to come up with a fix for that in the next 2 weeks? If not, then I'll make an intermediate release where HAVE_LIBC_VERSIONED_SYMBOLS is always set. That way, at least the more common case works reliably.

bsbernd:
I will try today, latest over the weekend. Right now working on fuse anyway (you will see soon) - next week I will probably need to shift most time to another project.

Nikratio:
Thanks a lot!

bsbernd:
@amesser if you should have chance, could you please test this patch? https://github.com/libfuse/libfuse/pull/726

amesser:
I'm sorry but i have not really something to test this with. I found the issue during code review while I was digging into a Devuan package bug. We observed a symbol issue with libvirts lxc module. It turned out, that libvirt built with libfuse 3.13 would show the issue when used with binary build of libfuse 3.12.

https://bugs.devuan.org/cgi/bugreport.cgi?bug=738

PR #726, opened 2023-01-27
Install a the configure_file (config.h) and use in headers

bsbernd:
This addresses: https://github.com/libfuse/libfuse/issues/724

HAVE_LIBC_VERSIONED_SYMBOLS configures the library if to use versioned symbols and is set at meson configuration time. External filesystems (the main target, actually)
include fuse headers and the preprocessor
then acts on HAVE_LIBC_VERSIONED_SYMBOLS. Problem was now that 'config.h' was not distributed with libfuse and so HAVE_LIBC_VERSIONED_SYMBOLS was never defined with external tools and the preprocessor did the wrong decision.

This commit also increases the the minimal meson version, as this depends on meson feature only available in 0.50

<quote 'meson' >
WARNING: Project specifies a minimum meson_
version '>= 0.42' but uses features which were added
 in newer versions:
 * 0.50.0: {'install arg in configure_file'} </quote>

Additionally the config file has been renamed to "fuse_config.h" to avoid clashes - 'config.h' is not very specific.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,6 @@
 sudo: required
-dist: bionic
+dist: jammy
 
-before_install:
-    - curl https://spot.fedorapeople.org/spotfoss-ppa.key | sudo apt-key add -
-    - wget https://launchpad.net/~spotfoss/+archive/ubuntu/ppa/+files/valgrind_3.16.1-2_amd64.deb
-    - sudo apt install ./valgrind_3.16.1-2_amd64.deb
 language:
     - c
     - c++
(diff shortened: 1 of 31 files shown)


bsbernd:
Oh well, I'm afraid the test environment is too old,  as we need to install fuse_config,h we need to set "configure_file: true", but according to the meson message that is only available with meson-0.50.

bsbernd:
Option 2 was simple - test updated to ubuntu 22.04 and it passes. Question is how many users there are on old distros and if they could just update meson. Actually updating meson through doesn't seem hard:
https://askubuntu.com/questions/1212424/update-meson-on-ubuntu-18-04-4-lts-force-update

Issue #729, opened 2023-02-07
3.13.1: PACKAGE_VERSION redefined


jengelh:
Forwarded from an openSUSE report:

FUSE 3.13.1 is breaking compilation in other packages, e.g. open-vm-tools (last-known-good 3.13.0):

```
[...]

PR #731, opened 2023-02-07
Split config.h into private and public config

bsbernd:
This addresses https://github.com/libfuse/libfuse/issues/729

commit db35a37def14b72181f3630efeea0e0433103c41 introduced a public config.h (rename to fuse_config.h to avoid conflicts) that was installed with the package and included by libfuse users through fuse_common.h. Probablem is that this file does not have unique defines so that they are unique to libfuse - on including the file conflicts with libfuse users came up.

In principle all defines could be prefixed, but then most of them are internal for libfuse compilation only. So this splits out publically required defines to a new file 'libfuse_config.h' and changes back to include of "fuse_config.h" only when HAVE_LIBFUSE_PRIVATE_CONFIG_H is defined.

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -948,15 +948,15 @@ struct fuse *fuse_new_30(struct fuse_args *args, const struct fuse_operations *o
 			 size_t op_size, void *private_data);
 #define fuse_new(args, op, size, data) fuse_new_30(args, op, size, data)
 #else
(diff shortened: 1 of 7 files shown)


Nikratio:
Thank you Bernd! Third time's the charm :-).

bsbernd:
@Nikratio I think no test had been running? I actually also wanted to review my own changes - I will do tomorrow. So please don't release a new version yet.

PR #735, opened 2023-02-19
Add a github actions file

bsbernd:
converted from travis with _sligh_ editing by
https://akx.github.io/travis-to-github-actions/

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -1,0 +1,26 @@
+name: Converted Workflow
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  build:
+    runs-on: '${{ matrix.os }}'
+    strategy:
+      matrix:
+        os:
+          # 18.04 does not work because of ‘no_sanitize’ attribute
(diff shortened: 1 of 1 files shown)


bsbernd:
@Nikratio No idea if this works, maybe we can give it a try?

bsbernd:
I just got a mail from git telling me that the automatic build failed - so this basically works, but has issues.

Nikratio:
I don't see any output anywhere (I would have expected something under the "Checks" tab). But maybe it need to be merged first, let's try.

bsbernd:
@Nikratio I got emails from github that my patch had failed, because of the missing cpp -DHAVE_PRIVATE....

Not sure if you get full logs:
https://github.com/bsbernd/libfuse/actions/runs/4213988753/jobs/7314133328

bsbernd:
I think we should try out a test patch, but I'm rather busy today and traveling tomorrow.

Nikratio:
Thanks for your help on this! At least following a merge, the test scripts are being run now. They currently fail because what looks like a toolchain incompatibility with me (valgrind not liking the clang debug information). I'll look into this.

Issue #734, opened 2023-02-18
test/test_examples.py::test_passthrough_hp[False] fails because it uses private defs


pierre-labastie:
The recent change splitting defs between private and public has broken the example/passthrough_hp.cc program, because it uses(at least) HAVE_UTIMENSAT.  I guess -DHAVE_LIBFUSE_PRIVATE_CONFIG_H should be added to the cpp project_arguments (in meson.build) too.
I can make a PR if you tell me to (but the fix seems simple to me. I may not have the whole picture though).

bsbernd:
Hmm, I guess it got broken because in the previous patch I had removed manual includes of config.h - assumption was that we could use it globally. After splitting example/ builds still worked, but I hadn't run the tests (somehow always fail for me when run locally) and I had hoped that github/projects tests would catch issues  - with the additional problem that there are Travis issues and it didn't run - issue not caught...
Summary: There might be more than passthrough_hp.cc

I like your idea of -DHAVE_LIBFUSE_PRIVATE_CONFIG_H in example/meson.build - would add to all files there. If you do, would you mind to also add a comment, explaining that examples rely on HAVE_ configure tests?

bsbernd:
One more
>grep -r HAVE test/
test/test_syscalls.c:#ifdef HAVE_COPY_FILE_RANGE

bsbernd:
With github actions activated it fails because of the issue here.

https://github.com/bsbernd/libfuse/actions/runs/4213834063/jobs/7313849593

pierre-labastie:
Thanks for the fix. Actually, I hadn't thought of using meson.build in sub-directories ;)  I was just thinking of replacing `-DHAVE_CONFIG_H` with `-DHAVE_LIBFUSE_PRIVATE_CONFIG_H` in top meson.build. The problem is that this replacement has been done for C, but not for CPP (oversight?). I think this propagate to subdirs, because when I made this replacement, the test passed.

Note that what you have done in https://github.com/bsbernd/libfuse/commit/ad3d685a122b767d9ca5b4f3c599693744ffe8f5 is probably redundant, since all C programs were getting the right define already.

bsbernd:
Ah right, I shouldn't work too long in the night (on something else). Do you want to submit a patch?

pierre-labastie:
Will do.

PR #737, opened 2023-02-19
Change define for C++ too

pierre-labastie:
Commit d7560cc has split defines into private and public, and passed -DHAVE_LIBFUSE-PRIVATE_CONFIG_H to all C programs. But the arguments of C++ programs have not been changed. This leads to a test failure as reported in issue #734. Pass -DHAVE_LIBFUSE-PRIVATE_CONFIG_H to C++ programs too.

Fixes: #734


pierre-labastie:
Sorry, made a pull request against my own repository at first. This should be against the right one now

bsbernd (approved):
Thanks for the patch!

PR #740, opened 2023-03-02
Use a btree for inode lookup in passthrough_ll

bsbernd:


trapexit:
Arguably, if the change in data structure is useful it should really be added to the high level API. That said some performance numbers would be good before changing something that has been around for many years.

bsbernd:
@trapexit That has nothing to do with high level - it is pure problem of passthrough_ll itself. 

@Nikratio See the patch - from my point of view it simplifies the code by abstracting manual list handling to a common and rather well tested btree lib.
If you want to change passthrough_hp - basically the entire manual argument parsing should be replaced in favor of fuse_parse_cmdline - a big and cumbersome change.

bsbernd:
I also do not see big differences between passthrough_ll and passthrough_hp - the latter uses C++ and an unordered map (hash table) instead of C and a plain list. Otherwise passthrough_ll and passthrough_hp are rather similar. Well, the other big difference is argument parsing.

trapexit:
> @trapexit That has nothing to do with high level - it is pure problem of passthrough_ll itself.

I know it doesn't. My point was that plausibly this tree could be more efficient than what is already there in the HLAPI. And that API is used by a number of projects so a performance gain there is worth more than in a demo app.

bsbernd:
Ah ok, misunderstood you.

Nikratio:
The most important difference between passthrough_ll and passthrough_fh is the intention. As the file comments says:

```
*  In contrast to passthrough.c and passthrough_fh.c,
 * this implementation uses the low-level API. Its performance should
 * be the least bad among the three, but many operations are not
 * implemented. In particular, it is not possible to remove files (or
 * directories) because the code necessary to defer actual removal
 * until the file is not opened anymore would make the example much
 * more complicated.
 *
 * When writeback caching is enabled (-o writeback mount option), it
 * is only possible to write to files for which the mounting user has
 * read permissions. This is because the writeback cache requires the
 * kernel to be able to issue read requests for all files (which the
 * passthrough filesystem cannot satisfy if it can't read the file in
 * the underlying filesystem).
```

Contrast this with:

```
 * This is a "high-performance" version of passthrough_ll.c. While
 * passthrough_ll.c is designed to be as simple as possible, this
 * example intended to be as efficient and correct as possible.
 *
 * passthrough_hp.cc mirrors a specified "source" directory under a
 * specified the mountpoint with as much fidelity and performance as
 * possible.
```

If this is not yet fully reflected in the implementation, then I'd like us see move towards this, not away from it.

That said, I agree that the changes to passthrough_ll.c are not obviously making things more or less simple (it depends on how you think about using preprocessor macros). 

However:

- I am not a big friend of adding custom/copied b-tree implementation to FUSE 
- Especially when this implementation is used only by one of the examples
- Even less so when the example that uses is is supposed to be simple & self-contained
- passthrough_ll is not intended to be used as a benchmark or to run tests (cf file comments above)

Is the problem just that passthrough_hp doesn't accept `-o <stuff>` options?

bsbernd:
Here is my `/sbin/mount.fuse.passthrough_ll`script for xfstests, which results in 

```
bernd@squeeze1 ~>cat /tmp/mount.fuse.log 
non1 /mnt/test -o rw,source=/home/test/test,allow_other,default_permissions
non2 /mnt/scratch -o rw,source=/home/test/scratch,allow_other,default_permissions
...
```

The `for-next`branch of xfstests has the patches to test fuse and documents passthrough_ll. As I wrote before, I had tried passthrough_hp- it does not work - does not accept the options above.

I had especially moved the tree.h into the include dir, as my assumption is that a tree might be helpful for lib and other examples as well.

bsbernd:
Also, xfstests is by no means a benchmark...

Nikratio:
Sorry, I'm still not sure I fully understand what you are saying. When you say "it does not work", do you mean it has other problems that would make it unfeasible to use with xfstests even if it were accepting `-o <stuff>` options? (From your example it looks as if you're not using any other options).

bsbernd:
Well, as I wrote before test generic/531 takes endless to complete - on my debug system 72h, with non debug kernel still several hours. On my colleagues laptop in summary more than 24h (most time in generic/531).
Mount options are specified in xfstest's `local.config`, but I'm not sure if the way passthrough_hp wants options is compatible with xfstests - the README.fuse example gives options in fstab style.

To be honest, I didn't expect this patch to be a big problem - let's just close the pull request. Less time consuming for me (and other group members and also for you ;) ) to move forward with a private branch and/or maybe forked off passthrough_ll. I will leave it to someone else to fix passthrough_hp or to write a new tool.

Nikratio:
I have a personal interest in getting passthrough_hp to be as compliant and fast as possible, so I might well be the person who makes the necessary changes once I understand what they are :-).

If I understand your commit message correctly, then generic/531 takes ages to complete with passthrough_ll. So currently the only known problem with passthrough_hp is the option parsing?

Are https://github.com/tytso/xfstests/blob/master/README.fuse the instructions I should follow to get passthrough_hp to work, or is there something more recent (since you talked about having a related private branch)?

bsbernd:
Ok thanks, I really need to concentrate on other work....

Yes, the only issue I know of is option parsing.

I think this is the official repo: https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git
There is nothing more to follow than README.fuse. Personally I run it in a VM, with a backup - you need to run it as root and previous fuse patch versions were running `rm -rf /`. Does not happen anymore right now, but I still don't feel comfortable to give it access to my $HOME

Nikratio:
I think I have an idea of how things should work, but I'm struggling to test it. I followed the instructions in the README and README.fuse file, but:

```
$ cat local.config
export TEST_DEV=non1
export TEST_DIR=/mnt/test
export SCRATCH_DEV=non2
export FSTYP=fuse
export FUSE_SUBTYP=.passthrough
export MOUNT_OPTIONS="-osource=/home/test/scratch,allow_other,default_permissions"
export TEST_FS_MOUNT_OPTS="-osource=/home/test/test,allow_other,default_permissions"

$  ./check
mkfs not found
```

Any chance you could give me a complete recipe for running the tests with FUSE?

bsbernd:
Mine looks like this:

```
(etch64)fusetests@squeeze1:~/src/xfstests-dev$ cat local.config
export TEST_DEV=non1
export TEST_DIR=/mnt/test
export SCRATCH_DEV=non2
export FSTYP=fuse
export FUSE_SUBTYP=.passthrough_ll
export MOUNT_OPTIONS="-osource=/home/test/scratch,allow_other,default_permissions"
export TEST_FS_MOUNT_OPTS="-osource=/home/test/test,allow_other,default_permissions"
```

And then run
`sudo ./check -fuse -b`

*NOTE*: This runs as root and _had_ the habit (earlier fuse patch versions  for xfstests) to run `rm -fr /` - be careful and better run in a VM or container without important data.

Nikratio:
Closing in favor of #747 (I think you said that is a reasonable path forward?)

PR #741, opened 2023-03-02
Fix a warning in iconv_convpath

bsbernd:
gcc-12 brings up a warning

[13/63] Compiling C object lib/libfuse3.so.3.12.0.p/modules_iconv.c.o ../lib/modules/iconv.c: In function ‘iconv_convpath’: ../lib/modules/iconv.c:85:38: warning: pointer ‘newpath’ may be used after ‘realloc’ [-Wuse-after-free]
   85 |                         p = tmp + (p - newpath);
      |                                   ~~~^~~~~~~~~~
../lib/modules/iconv.c:80:31: note: call to ‘realloc’ here
   80 |                         tmp = realloc(newpath, newpathlen + 1);
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

This also refactors the code a bit, in order to make it better readable. There was also no problem with the old code, the order just confused the compiler.

### DIFF
--- a/lib/modules/iconv.c
+++ b/lib/modules/iconv.c
@@ -44,11 +44,12 @@ static int iconv_convpath(struct iconv *ic, const char *path, char **newpathp,
 {
 	size_t pathlen;
 	size_t newpathlen;
-	char *newpath;
-	size_t plen;
-	char *p;
+	char *newpath, *outbuf_ptr;
+	size_t outbytesleft;
 	size_t res;
 	int err;
+	char **inbuf_ptr = (char **restrict)&path;
+	size_t inbuf_left;
 
 	if (path == NULL) {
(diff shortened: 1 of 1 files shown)


bsbernd:
@Nikratio I didn't run any test on that. Not sure if we want to merge immediately. Maybe wait a bit, if I find the time to create a test.

bsbernd:
@nm004  This addresses the same issue as your patch here https://github.com/libfuse/libfuse/pull/686

Nikratio:
Is this PR still relevant?

bsbernd:
I still think that this code is better readable, but I don't have the time for a test - let's close for the moment, I keep my branch and once I find some I will add test.

Issue #10, opened 2016-01-16
so version has not been upgraded to 2.9.5


Lesseps:
Previous releases of fuse (libfuse) had package and so versions identical:

```
fuse-2.9.x -> libfuse.so.2.9.x
```

In present version 2.9.5, should be **libfuse.so.2.9.5**, not libfuse.so.2.9.4 . This has not been done.

It is defined in _lib/Makefile.am_, and should be

```
libfuse_la_LDFLAGS = -pthread @libfuse_libs@ -version-number 2:9:5 \
```

not

```
libfuse_la_LDFLAGS = -pthread @libfuse_libs@ -version-number 2:9:4 \
```

For users, fix just changing _lib/Makefile.in_ with

```
sed -i  's/2:9:4/2:9:5/' lib/Makefile.in
```
For the package, tested the same with _lib/Makefile.am_:

```
sed -i  's/2:9:4/2:9:5/' lib/Makefile.am
```
but needed to _autoreconf_.

Please, tell me if you need any other info.

Thanks.

Nikratio:
Thanks for the extensive report! I'll make sure to do this correctly in the future!

Nikratio:
(I don't think I need to update this release, because there were absolutely no changes to the API)

Nikratio:
Opening this again to prevent further duplicates.

Nikratio:
Fixed in 2.9.6.

PR #747, opened 2023-03-17
Xfstests

Nikratio:
@bsbernd 

Could you take a look at this? Would this work for your use-case instead of #740 ?

I am not perfectly happy with the --fsname option. It would be much nicer to just use the source directory as the filesystem name, but it seems that the `check` script makes some weird assumptions about this. You can reproduce the same problem with `passthrough_ll`  and the following `local.config`:

```
export TEST_DEV=/mnt/src/test
export TEST_DIR=/mnt/test
export SCRATCH_DEV=/mnt/src/scratch
export FSTYP=fuse
export FUSE_SUBTYP=.passthrough_ll
export MOUNT_OPTIONS="-osource=/mnt/src/scratch,allow_other,default_permissions"
export TEST_FS_MOUNT_OPTS="-osource=/mnt/src/test,allow_other,default_permiss
```

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -157,6 +157,7 @@ struct Fs {
     bool nocache;
     size_t num_threads;
     bool clone_fd;
+    std::string fuse_mount_options;
 };
 static Fs fs{};
 
@@ -1169,8 +1170,36 @@ static cxxopts::ParseResult parse_wrapper(cxxopts::Options& parser, int& argc, c
 }
 
 
+static void string_split(std::string s, std::vector<std::string>& out, std::string delimiter) {
+    size_t pos_start = 0, pos_end, delim_len = delimiter.length();
+    std::string token;
+
+    while ((pos_end = s.find(delimiter, pos_start)) != std::string::npos) {
+        token = s.substr(pos_start, pos_end - pos_start);
+        pos_start = pos_end + delim_len;
+        out.push_back(token);
+    }
+
+    out.push_back(s.substr(pos_start));
+}
+
+
+static std::string string_join(const std::vector<std::string>& elems, char delim)
+{
+    std::ostringstream out;
+    for (auto ii = elems.begin(); ii != elems.end(); ++ii) {
+        out << (*ii);
+        if (ii + 1 != elems.end()) {
+            out << delim;
+        }
+    }
+    return out.str();
+}
+
+
 static cxxopts::ParseResult parse_options(int argc, char **argv) {
     cxxopts::Options opt_parser(argv[0]);
+    std::vector<std::string> mount_options;
     opt_parser.add_options()
         ("debug", "Enable filesystem debug messages")
         ("debug-fuse", "Enable libfuse debug messages")
@@ -1179,6 +1208,8 @@ static cxxopts::ParseResult parse_options(int argc, char **argv) {
         ("nocache", "Disable all caching")
         ("nosplice", "Do not use splice(2) to transfer data")
         ("single", "Run single-threaded")
+        ("o", "Mount options (see mount.fuse(5) - only use if you know what "
+              "you are doing)", cxxopts::value(mount_options))
         ("num-threads", "Number of libfuse worker threads",
                         cxxopts::value<int>()->default_value(SFS_DEFAULT_THREADS))
         ("clone-fd", "use separate fuse device fd for each thread",
@@ -1220,6 +1251,30 @@ static cxxopts::ParseResult parse_options(int argc, char **argv) {
     fs.source = std::string {resolved_path};
     free(resolved_path);
 
+    std::vector<std::string> flattened_mount_opts;
+    for (auto opt : mount_options) {
+        string_split(opt, flattened_mount_opts, ",");
+    }
+
+    bool found_fsname = false;
+    for (auto opt : flattened_mount_opts) {
+        if (opt.find("fsname=") == 0) {
+            found_fsname = true;
+            continue;
+        }
+
+        /* Filter out some obviously incorrect options. */
+        if (opt == "fd") {
+            std::cout << argv[0] << ": Unsupported mount option: " << opt << "\n";
+            print_usage(argv[0]);
+            exit(2);
+        }
+    }
+    if (!found_fsname) {
+        flattened_mount_opts.push_back("fsname=" + fs.source);
+    }
+    flattened_mount_opts.push_back("default_permissions");
+    fs.fuse_mount_options = string_join(flattened_mount_opts, ',');
     return options;
 }
 
@@ -1244,7 +1299,6 @@ int main(int argc, char *argv[]) {
 
     // Parse command line options
     auto options {parse_options(argc, argv)};
-
 
     // We need an fd for every dentry in our the filesystem that the
     // kernel knows about. This is way more than most processes need,
@@ -1272,7 +1326,7 @@ int main(int argc, char *argv[]) {
     fuse_args args = FUSE_ARGS_INIT(0, nullptr);
     if (fuse_opt_add_arg(&args, argv[0]) ||
         fuse_opt_add_arg(&args, "-o") ||
-        fuse_opt_add_arg(&args, "default_permissions,fsname=hpps") ||
+        fuse_opt_add_arg(&args, fs.fuse_mount_options.c_str()) ||
         (fs.debug_fuse && fuse_opt_add_arg(&args, "-odebug")))
         errx(3, "ERROR: Out of memory");
 
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -1288,8 +1288,6 @@ int main(int argc, char *argv[]) {
     // Don't apply umask, use modes exactly as specified
     umask(0);
 
-    fuse_daemonize(fs.foreground);
-
     // Mount and run main loop
     loop_config = fuse_loop_cfg_create();
 
@@ -1298,6 +1296,9 @@ int main(int argc, char *argv[]) {
 
     if (fuse_session_mount(se, argv[2]) != 0)
         goto err_out3;
+
+    fuse_daemonize(fs.foreground);
+
     if (options.count("single"))
         ret = fuse_session_loop(se);
     else
--- a/xfstests/mount.fuse.passthrough
+++ b/xfstests/mount.fuse.passthrough
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+ulimit -n 1048576
+
+# It would be easier if we could just set SCRATCH_DEV and TEST_DEV to the source directory
+# path in local.options. Unfortunately, setting these variables to a path seems get
+# xfstests all worked up (even though it should treat these as opaque values), and it
+# refuses to even start running any tests).
+dev="$1"
+shift
+if [ "$dev" = "${SCRATCH_DEV}" ]; then
+  source="${SCRATCH_SOURCE}"
+else
+  source="${TEST_SOURCE}"
+fi
+
+exec "$PASSTHROUGH_PATH" -o fsname=$dev,allow_other "${source}" "$@"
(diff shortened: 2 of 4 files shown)


example/passthrough_hp.cc
```diff
@@ -1211,6 +1214,13 @@ static cxxopts::ParseResult parse_options(int argc, char **argv) {
     if (fs.debug || fs.debug_fuse)
         fs.foreground = true;
 
+    auto mnt_opts = options["o"].as<std::string>();
```
bsbernd, line 1217: I guess there should be added a comment here that this is specially for xfstests.

example/passthrough_hp.cc
```diff
@@ -1211,6 +1214,13 @@ static cxxopts::ParseResult parse_options(int argc, char **argv) {
...
+    if (mnt_opts != "" && mnt_opts != "rw") {
+        std::cout << argv[0] << ": Unsupported mount option(s): " << mnt_opts << "\n";
```
bsbernd, line 1219: Hmm, I admit that I don't like that specific testing for "rw" too much. What is if we want to test with options like "sync,nolazytime" - all these flags handled by the vfs?

bsbernd (commented):
Thanks a lot Nicolas, I guess it works (will test later today), but I'm concerned about "rw".

example/passthrough_hp.cc
bsbernd: I guess we can extend it later and add a function to iterate over all parameters and to check if these are vfs parameters.

Nikratio:
Yeah, I'm not happy with the handling of mount options, but I wanted to first check if this is a suitable alternative otherwise.

Instead of allowlisting acceptable mount options, we could instead denylist everything that does not make sense to set (starting with what mount.fuse(5) describes as "It is set by libfuse"). But I am not sure if that truly captures everything. For example, I believe that `relatime` and `noatime` need to be handled by the filesystem daemon (the kernel module only updates timestamps when writeback caching is enabled). 

As far as I can tell, we have to pick between accepting some mount options then don't actually take effect, or rejecting some mount options that the kernel would happily implement for us...

Thoughts?

Nikratio:
(added an alternate implementation that - for now - accepts any mount option).

bsbernd:
See <linux>/fs/inode.c

```
bool atime_needs_update(const struct path *path, struct inode *inode)
{
	struct vfsmount *mnt = path->mnt;
	struct timespec64 now;

	if (inode->i_flags & S_NOATIME)
		return false;
		
	...
	
	if (!relatime_need_update(mnt, inode, now))
		return false;

```

So `noatime` and `relatime` are entirely handled by the vfs. 

Isn't what `fuse_parse_cmdline` and `fuse_opt_parse` already doing what we need? We want it to handle all local options and forward the remaining to libfuse. Except that passthrough_hp isn't using it and has its own option parser?

bsbernd:
Somehow it does not start the tests,  so far I don't figure out why. 

With passthrough_ll
```
(etch64)fusetests@squeeze1:~$ src/xfstests-dev/run-fuse-check.sh
Running cd src/xfstests-dev
Running sudo ./check -fuse -b
FSTYP         -- fuse
PLATFORM      -- Linux/x86_64 squeeze1 6.2.0-rc8 #10 SMP Wed Mar  8 16:51:00 UTC 2023
MKFS_OPTIONS  -- non2
MOUNT_OPTIONS -- -osource=/home/test/scratch,allow_other,default_permissions non2 /mnt/scratch

generic/001 78s ... ^C(etch64)fusetests@squeeze1:~$ ^C
...
```
With passthrough_hp

```
(etch64)fusetests@squeeze1:~$ src/xfstests-dev/run-fuse-check.sh
Running cd src/xfstests-dev
Running sudo ./check -fuse -b
FSTYP         -- fuse
PLATFORM      -- Linux/x86_64 squeeze1 6.2.0-rc8 #10 SMP Wed Mar  8 16:51:00 UTC 2023
MKFS_OPTIONS  -- non2
MOUNT_OPTIONS -- non2 /mnt/scratch

```

bsbernd:
And we definitely should have the paths in local.config instead of in the mount command.

bsbernd:
Do you mind if we hold on a bit on this? I really need to send out my kernel patches first. Right now missing is a description and design file - in the middle to create this...

Nikratio:
> Somehow it does not start the tests, so far I don't figure out why.
> 
> With passthrough_ll
> 
> ```
> (etch64)fusetests@squeeze1:~$ src/xfstests-dev/run-fuse-check.sh
> Running cd src/xfstests-dev
> Running sudo ./check -fuse -b
> FSTYP         -- fuse
> PLATFORM      -- Linux/x86_64 squeeze1 6.2.0-rc8 #10 SMP Wed Mar  8 16:51:00 UTC 2023
> MKFS_OPTIONS  -- non2
> MOUNT_OPTIONS -- -osource=/home/test/scratch,allow_other,default_permissions non2 /mnt/scratch
> 
> generic/001 78s ... ^C(etch64)fusetests@squeeze1:~$ ^C
> ...
> ```
> 
> With passthrough_hp
> 
> ```
> (etch64)fusetests@squeeze1:~$ src/xfstests-dev/run-fuse-check.sh
> Running cd src/xfstests-dev
> Running sudo ./check -fuse -b
> FSTYP         -- fuse
> PLATFORM      -- Linux/x86_64 squeeze1 6.2.0-rc8 #10 SMP Wed Mar  8 16:51:00 UTC 2023
> MKFS_OPTIONS  -- non2
> MOUNT_OPTIONS -- non2 /mnt/scratch
> ```

I've just re-checked, and it works for me:

```
root@debian-vm:/home/nikratio/xfstests-dev# ./check -fuse
Found mount option: fsname=non1
Found mount option: allow_other
Found mount option: rw
Final mount options: fsname=non1,allow_other,rw,default_permissions
FSTYP         -- fuse
PLATFORM      -- Linux/x86_64 debian-vm 5.10.0-21-amd64 #1 SMP Debian 5.10.162-1 (2023-01-21)
MKFS_OPTIONS  -- non2
MOUNT_OPTIONS -- non2 /mnt/scratch

generic/001 1s ...  2s
generic/002 0s ...  0s
generic/003       - output mismatch (see /home/nikratio/xfstests-dev/results//generic/003.out.bad)
    --- tests/generic/003.out	2023-03-11 15:54:54.822981256 +0000
    +++ /home/nikratio/xfstests-dev/results//generic/003.out.bad	2023-03-18 15:22:49.445480032 +0000
    @@ -1,2 +1,64 @@
     QA output created by 003
    +Found mount option: fsname=non2
    +Found mount option: allow_other
    +Found mount option: rw
    +Found mount option: relatime
    +Final mount options: fsname=non2,allow_other,rw,relatime,default_permissions
    +fuse: unknown option(s): '-o relatime'
[...]
```

I've encountered a number of similar quiet refusals to work before.  Causes were:

- fsname not set as expected
- allow_other not set
- TEST_DEV set to something that xfstests seems to dislike

bsbernd:
Ok it works for me as well (your example is using different dirs and one didn't exist), but generic/003 should not fail, but succeed with 

```
generic/003       [not run] relatime not supported by the current kernel
```
(which probably isn't correct either, but probably a fusermount issue), but at least passthrough_ll does not fail.

Look how it passes mount options

```
+ PASSTHROUGH_PATH=/home/bernd/src/libfuse/github/build-debian/example/passthrough_hp
+ SCRATCH_SOURCE=/mnt/src/scratch
+ TEST_SOURCE=/mnt/src/test
+ echo non2 /mnt/scratch -o rw,strictatime
+ tee -a /tmp/mount.fuse.log
non2 /mnt/scratch -o rw,strictatime
+ ulimit -n 1048576
+ dev=non2
+ shift
+ '[' non2 = non2 ']'
+ source=/mnt/src/scratch
+ exec /home/bernd/src/libfuse/github/build-debian/example/passthrough_hp -o fsname=non2,allow_other /mnt/src/scratch /mnt/scratch -o rw,strictatime
Found mount option: fsname=non2
Found mount option: allow_other
Found mount option: rw
Found mount option: strictatime
Final mount options: fsname=non2,allow_other,rw,strictatime,default_permissions
fuse: unknown option(s): '-o strictatime'
```
From the output I actually start to wonder if passthrough_hp now passes it more correctly than passthrough_ll - I need to see what passthrough_ll does with that `strictatime`. 

One way for the other, we need to update lib/mount.c to support missing options/flags.

bsbernd:
Hmm, after adding the mount options
```
+ERROR: access time has not been updated after accessing file1 first time
+non2 /mnt/scratch -o rw,nodiratime
--
+ERROR: access time has not been updated after accessing file2
+non2 /mnt/scratch -o rw,noatime
--
+ERROR: access time has not been updated after accessing file3 third time
+non2 /mnt/scratch -o ro,strictatime
```

Looks like it found a real issue.

bsbernd:
https://github.com/libfuse/libfuse/pull/749

Nikratio:
Regarding the handling of atime:

At least without any mount options specified, atime is not updated by the kernel. For example,

```
$ examples/hello_ll ~/tmp/mnt -d -f &
$ cat ~/tmp/mnt/hello
```

does not print any SETATTR requests - i.e., atime remains unchanged unless the file system daemon updates it on its own.

(Cross-check: `touch ~/tmp/mnt/hello` results in a SETATTR request as expected)

I couldn't find a straightforward test for the handling of `noatime` or `relatime`, but it seems to be that if a filesystem is taking care of updating atime, but does not check for the mount options, then they may fail to take effect no matter what the kernel does with them.

bsbernd:
Yeah, I would have expected the kernel to send a setattr, at least with "strictatime" - I guess I need to check kernel side what exactly happens. Btw, I had forgotten fusermount for the time options and also wrongly used the key "lazy_a_time" instead of "lazytime".

Nikratio:
@bsbernd Any update on this?

bsbernd (commented):
Rename mount.fuse.passthrough to `mount.fuse.passthrough_hp`?

xfstests/mount.fuse.passthrough
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Change the following:
+PASSTHROUGH_PATH=/home/nikratio/libfuse/build/example/passthrough_hp
+SCRATCH_SOURCE=/mnt/src/scratch
+TEST_SOURCE=/mnt/src/test
+
```
bsbernd, line 7: This should really go into local.config

bsbernd:
I could get it working with your patch, just having the destination fixed in /sbin/mount.fuse.passthrough is confusing (and caused the initial issue for me that the dest directory did not exist).

Nikratio:
What about the question which mount options to refuse, and which ones to pass-through (and allowlisting vs denylisting)?

Nikratio:
> Rename mount.fuse.passthrough to `mount.fuse.passthrough_hp`?

I'd rather keep that as-is. The other passthrough's are really just meant for people to study the code (not to be used for anything), and the filenames are already long enough as-is.

bsbernd:
Well, it might make sense to test passthrough.c  with xfstests - high level has especially path resolution issues and xfstests _might_ be helpful to test high level... We might even consider to use a short version of xftests for all pull requests - and then low and high level.

bsbernd:
> What about the question which mount options to refuse, and which ones to pass-through (and allowlisting vs denylisting)?

Can't we leave it as it is now, just pass through everything given with -o?

Nikratio:
> > What about the question which mount options to refuse, and which ones to pass-through (and allowlisting vs denylisting)?
> 
> Can't we leave it as it is now, just pass through everything given with -o?

We can, it's probably still an improvement over the status quo.

I was hoping we could at least firmly establish the status of the *time options, i.e. should they be handled in the kernel (and if so, why isn't this working) or should they be handled by the FS (in which case I'd at least refuse to accept these mount options).

I thought that there was a test failing because of this (e.g. lazyatime being accepted but not having any effect).

Issue #586, opened 2021-02-24
auto_unmount doesn't work without allow_other


mahkoh:
In https://github.com/libfuse/libfuse/blob/77d662459a0fcdf358d515477d33795837e859d5/util/fusermount.c#L1219 `open` is executed as root which does not have access to the mount point if allow_other was not used and the real user id is not 0. Since allow_other usually cannot be specified by unprivileged users, auto_unmount has no effect for unprivileged users.

@kevin-vigor Should we maybe drop privileges before calling `open`?

Nikratio:
Thanks for taking the time to report this issue!

Unfortunately, this project does not currently have any active, regular contributors. As the maintainer, I try to review pull requests and make regular releases, but unfortunately I have no capacity to do significant development beyond that.
Issue reports that do not come with a pull request or clearly have high impact on a large number of users are therefore likely to languish.

I understand that this is frustrating for users, but I hope you can also understand that any development work that I do on this project has to compete with spending time with my family, doing work that I get paid for, doing something recreational without a computer, or working on features/bugs that affect me personally. Most bugs and ideas - unfortunately including this one - loose out in this competition.

In other words, unless you plan to work on this yourself or can recruit someone who will, it's unlikely that anyone is going to do anything about it anytime soon.

This is just to calibrate expectations. I am grateful that you took the time and effort to report this! I'll leave this issue open to document the problem and who knows, maybe someone will pick it up after all :-).

matthiasgoergens:
@mahkoh @Nikratio I gave addressing this issue a try in https://github.com/libfuse/libfuse/pull/754.  Please have a look.  Thanks!

Feverfew:
@matthiasgoergens I tried fixing this [some time ago](https://invent.kde.org/system/kio-fuse/-/issues/180) by dropping privileges too, but with the key difference being that I didn't do this in a forked child process. Do you know why that forking is necessary here?

matthiasgoergens:
I forked, because the kind of privilege dropping that is required to get past a missing allow_other is not reversible.

The reversible privilege dropping that fusermount already does isn't enough.

PR #754, opened 2023-03-28
Fix `auto_unmount` to work without `allow_other`

matthiasgoergens:
In
https://github.com/libfuse/libfuse/blob/77d662459a0fcdf358d515477d33795837e859d5/util/fusermount.c#L1219 `open` is executed as root which does not have access to the mount point if `allow_other` was not used and the real user id is not 0. Since `allow_other` usually cannot be specified by unprivileged users, `auto_unmount` has no effect for unprivileged users.

In this commit, we work around this limitation:

We first try to open the mountpoint as root, and if we get `EACCES`, we retry as the user who started fusermount, and see if we get `ENOTCONN`.

In my testing, I found that `setfsuid` and `setfsgid` don't work to get around the lack of `allow_other`.  (Sorry, I don't know enough about the Linux kernel to tell whether that's significant.)  As a workaround, I decided to use `setresuid` and `setresgid` in a forked child process, and communicate via its exit status.

Please give feedback on correctness, style and suggest tests.

Fixes https://github.com/libfuse/libfuse/issues/586

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -1190,6 +1190,40 @@ static int send_fd(int sock_fd, int fd)
 	return 0;
 }
 
+static int mnt_is_ENOTCONN_for_owner(const char *mnt)
+{
+	int pid = fork();
+	if(pid == -1) {
+		perror("fuse: mnt_is_ENOTCONN_for_owner can't fork");
+		_exit(EXIT_FAILURE);
+	} else if(pid == 0) {
+		uid_t uid = getuid();
+		gid_t gid = getgid();
+		if(setresgid(gid, gid, gid) == -1) {
+			perror("fuse: can't set resgid");
+			_exit(EXIT_FAILURE);
+		}
+		if(setresuid(uid, uid, uid) == -1) {
+			perror("fuse: can't set resuid");
+			_exit(EXIT_FAILURE);
+		}
+
+		int fd = open(mnt, O_RDONLY);
+		if(fd == -1 && errno == ENOTCONN)
+			_exit(EXIT_SUCCESS);
+		else
+			_exit(EXIT_FAILURE);
+	} else {
+		int status;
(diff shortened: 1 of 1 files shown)


util/fusermount.c
```diff
@@ -1190,6 +1190,40 @@ static int send_fd(int sock_fd, int fd)
 	return 0;
 }
 
+static int mnt_is_ENOTCONN_for_owner(const char *mnt)
```
bsbernd, line 1193: Would you mind to add the explanation in the commit message as function comment? Like

/*
 * fusermount typically has the s-bit set - initial open was as root and got EACCESS
 * as 'allow_other' was not specified.
 * Try again with uid of the calling process process
 */

util/fusermount.c
```diff
@@ -1225,10 +1259,21 @@ static int should_auto_unmount(const char *mnt, const char *type)
...
 	if (fd != -1) {
 		close(fd);
 	} else {
-		result = errno == ENOTCONN;
+		switch(errno) {
+		case ENOTCONN:
+			result = 1;
+			break;
+		case EACCES:
+			result = mnt_is_ENOTCONN_for_owner(mnt);
```
bsbernd, line 1271: I find the function name a bit confusing and only understood it, once I read the code. Maybe
something `?recheck_ENOTCONN_as_owner`

matthiasgoergens: Done!

util/fusermount.c
matthiasgoergens: Done.

matthiasgoergens:
@bsbernd Thanks for the quick review.  I pushed a new version that (hopefully!) addresses your remarks.

bsbernd (approved):
Thanks for the quick updates! Looks good to me.

Issue #148, opened 2017-03-15
FUSE mount ignores suid/dev mount options when auto_unmount is specified


jjobi:
FUSE mount is ignoring suid/dev mount options when auto_unmount option is specified.

Nikratio:
Could you provide more information on how/when this fails? It seems to work with the examples:

```
$ ./passthrough_fh -o nodev ~/tmp/mnt
$ echo foo > ~/tmp/mnt/dev/null 
-bash: /home/nikratio/tmp/mnt/dev/null: Permission denied

$ fusermount -u ~/tmp/mnt
$ ./passthrough_fh -o nodev,auto_unmount ~/tmp/mnt

$ echo foo > ~/tmp/mnt/dev/null 
-bash: /home/nikratio/tmp/mnt/dev/null: Permission denied
```

jjobi:
fuse_parse_cmdline() and fuse_level_low() when invoked with "-o suid,dev,auto_unmount" in the arguments, the file systems gets mounted with "nosuid,nodev".  You can see the output of mount command to verify.  "suid,dev" options get dropped somewhere in between.  If "auto_unmount" is not specified, then the file system gets mounted with the provided mount options.  I tested this with fuse.2.9.7.

Nikratio:
I see. When you run as root, fuse normally does not call `fusermount` but uses the mount system call directly. When you specify `auto_unmount`, it goes through `fusermount` instead. However, `fusermount` is a setuid binary that is normally called by regular users, so it cannot in general accept `suid` or `dev` options.

It may be possible for `fusermount3` to check if it was actually called as root (rather than running as root because of being setuid) and, in that case, accept the `dev` and `suid` options. Patches welcome :-).

Nikratio:
Now that this is at least documented, turning it into an enhancement request...

jjobi:
I could get this working with the following change

fuse-3.0.0/util/fusermount.c  2016-11-29 08:26:29.000000000 -0800
fuse-3.0.0/util/fusermount.c.new        2017-03-21 19:18:33.489382420 -0700
***************
*** 715,721 ****
                    char **mnt_optsp)
  {
        int res;
!       int flags = MS_NOSUID | MS_NODEV;
        char *optbuf;
        char *mnt_opts = NULL;
        const char *s;
--- 715,721 ----
                    char **mnt_optsp)
  {
        int res;
!     int flags = getuid() ? (MS_NOSUID | MS_NODEV) : 0;
[...]

PR #759, opened 2023-03-29
fusermount: provide suid,dev for privileged users

matthiasgoergens:
Fixes https://github.com/libfuse/libfuse/issues/148

This is a simple attempt to provide access to options like suid and dev for privileged users, without opening security holes in a suid binary.

I'm not sure this is a good idea, it's mostly for discussion.

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -736,7 +736,11 @@ static int do_mount(const char *mnt, const char **typep, mode_t rootmode,
 		    char **mnt_optsp)
 {
 	int res;
+	#ifdef NO_PRIVS
+	int flags = 0;
+	#else
 	int flags = MS_NOSUID | MS_NODEV;
+	#endif
 	char *optbuf;
 	char *mnt_opts = NULL;
 	const char *s;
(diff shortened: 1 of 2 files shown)


Nikratio:
I believe that when a privileged user starts a FUSE filesystem, fusermount3 is not used in the first place. Instead, the filesystem does the system call directly: https://github.com/libfuse/libfuse/blob/master/lib/mount.c#L467

Is this not working for you?

Nikratio:
Ah, looking at the referenced issue, what you are trying to do here is support `suid,dev` for privileged users *in combination with auto_unmount*, right?

matthiasgoergens:
Yes, exactly.

This code isn't doing that, yet, because at a minimum it would still need some extra logic in the rest of the library.

I just wanted to prototype the mechanism to deal with the new options while minimising the chance to mess up the binary that has suid-bit-set, since that part seemed to be a point of contention in the original discussion.

Here we let the OS handle that logic, instead of coding it up ourselves.

matthiasgoergens:
Closing this prototype in favour of https://github.com/libfuse/libfuse/pull/762

Issue #634, opened 2021-12-13
fusermount -u cannot find entry with a space in mountpoint


dgerber:
```
/tmp/test-space$ ls
'mnt space'   srcdir
/tmp/test-space$ findmnt |grep space
/tmp/test-space$ bindfs --no-allow-other srcdir/ mnt\ space/
/tmp/test-space$ findmnt |grep space
└─/tmp/test-space/mnt space /tmp/test-space/srcdir   fuse       rw,nosuid,nodev,relatime,user_id=1000,group_id=1000,default_permissions
/tmp/test-space$ grep space /etc/mtab
/tmp/test-space/srcdir /tmp/test-space/mnt\040space fuse rw,nosuid,nodev,relatime,user_id=1000,group_id=1000,default_permissions 0 0
/tmp/test-space$ fusermount3 -u mnt\ space/
fusermount3: entry for /tmp/test-space/mnt space not found in /etc/mtab
/tmp/test-space$ findmnt |grep space
└─/tmp/test-space/mnt space /tmp/test-space/srcdir   fuse       rw,nosuid,nodev,relatime,user_id=1000,group_id=1000,default_permissions
/tmp/test-space$ umount mnt\ space/
/tmp/test-space$ findmnt |grep space
/tmp/test-space$
```

The issue does not happen when running with uid 0, presumably because `may_unmount` is not called.

dgerber:
Precision: this is in Alpine Linux. The issue is with `getmntent` in musl -- have reported it there.

Nikratio:
Thanks for taking the time to report this issue!

Unfortunately, this project does not currently have any active, regular contributors. As the maintainer, I try to review pull requests and make regular releases, but unfortunately I have no capacity to do significant development beyond that.
Issue reports that do not come with a pull request or clearly have high impact on a large number of users are therefore likely to languish.

I understand that this is frustrating for users, but I hope you can also understand that any development work that I do on this project has to compete with spending time with my family, doing work that I get paid for, doing something recreational without a computer, or working on features/bugs that affect me personally. Most bugs and ideas - unfortunately including this one - loose out in this competition.

In other words, unless you plan to work on this yourself or can recruit someone who will, it's unlikely that anyone is going to do anything about it anytime soon.

This is just to calibrate expectations. I am grateful that you took the time and effort to report this! I'll leave this issue open to document the problem and who knows, maybe someone will pick it up after all :-).

matthiasgoergens:
[...]

PR #761, opened 2023-04-02
Workaround musl bug when mountdir has whitespace

matthiasgoergens:
Fixes https://github.com/libfuse/libfuse/issues/634 and https://github.com/mpartel/bindfs/issues/106

It'll be a while before my upstream fix makes it into musl.


matthiasgoergens:
Thanks for the quick review!

PR #762, opened 2023-04-04
Fuse mount: make auto_unmount compatible with suid/dev mount options

matthiasgoergens:
> When you run as root, fuse normally does not call fusermount but uses
> the mount system call directly. When you specify auto_unmount, it goes
> through fusermount instead. However, fusermount is a setuid binary that
> is normally called by regular users, so it cannot in general accept suid
> or dev options.

In this patch, we split up how fuse mounts as root when `auto_unmount` is specified.

First, we mount using system calls directly, then we reach out to fusermount to set up auto_unmount only (with no actual mounting done in fusermount).

Fixes https://github.com/libfuse/libfuse/issues/148

### DIFF
--- a/example/passthrough_ll.c
+++ b/example/passthrough_ll.c
@@ -1241,6 +1241,10 @@ int main(int argc, char *argv[])
 
 	} else {
 		lo.source = strdup("/");
+		if(!lo.source) {
+			fuse_log(FUSE_LOG_ERR, "fuse: memory allocation failed\n");
+			exit(1);
+		}
 	}
 	if (!lo.timeout_set) {
 		switch (lo.cache) {
--- a/example/passthrough_ll.c
+++ b/example/passthrough_ll.c
@@ -89,7 +89,7 @@ struct lo_data {
 	int writeback;
 	int flock;
 	int xattr;
-	const char *source;
+	char *source;
 	double timeout;
 	int cache;
 	int timeout_set;
@@ -1240,7 +1240,7 @@ int main(int argc, char *argv[])
 		}
 
 	} else {
-		lo.source = "/";
+		lo.source = strdup("/");
 	}
 	if (!lo.timeout_set) {
 		switch (lo.cache) {
@@ -1302,5 +1302,6 @@ err_out1:
 	if (lo.root.fd >= 0)
 		close(lo.root.fd);
 
+	free(lo.source);
 	return ret ? 1 : 0;
 }
--- a/lib/mount.c
+++ b/lib/mount.c
@@ -322,6 +322,65 @@ void fuse_kern_unmount(const char *mountpoint, int fd)
 	waitpid(pid, NULL, 0);
 }
 
+static int fuse_mount_fusermount_setup_auto_unmount_only(const char *mountpoint, int quiet)
+{
+	int fds[2], pid;
+	int res;
+
+	if (!mountpoint) {
+		fuse_log(FUSE_LOG_ERR, "fuse: missing mountpoint parameter\n");
+		return -1;
+	}
+
+	res = socketpair(PF_UNIX, SOCK_STREAM, 0, fds);
+	if(res == -1) {
+		perror("fuse: socketpair() failed");
+		return -1;
+	}
+
+	pid = fork();
+	if(pid == -1) {
+		perror("fuse: fork() failed");
+		close(fds[0]);
+		close(fds[1]);
+		return -1;
+	}
+
+	if(pid == 0) {
+		char env[10];
+		const char *argv[32];
+		int a = 0;
+
+		if (quiet) {
+			int fd = open("/dev/null", O_RDONLY);
+			if (fd != -1) {
+				dup2(fd, 1);
+				dup2(fd, 2);
+			}
+		}
+
+		argv[a++] = FUSERMOUNT_PROG;
+		argv[a++] = "--setup-auto-unmount-only";
+		argv[a++] = "--";
+		argv[a++] = mountpoint;
+		argv[a++] = NULL;
+
+		close(fds[1]);
+		fcntl(fds[0], F_SETFD, 0);
+		snprintf(env, sizeof(env), "%i", fds[0]);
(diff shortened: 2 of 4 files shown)


lib/mount.c
```diff
@@ -321,6 +321,65 @@ void fuse_kern_unmount(const char *mountpoint, int fd)
 	waitpid(pid, NULL, 0);
 }
 
+static int fuse_mount_fusermount_setup_auto_unmount_only(const char *mountpoint, int quiet)
```
matthiasgoergens, line 324: @Nikratio This is essentially a variant of `fuse_mount_fusermount`.

My first version crammed both variants into the same function.  That had less code duplication, but felt a bit too complicated.  Especially since the return value here communicates both errors xor a file desciptor.

lib/mount.c
```diff
@@ -589,7 +642,13 @@ int fuse_kern_mount(const char *mountpoint, struct mount_opts *mo)
 		goto out;
 
 	res = fuse_mount_sys(mountpoint, mo, mnt_opts);
-	if (res == -2) {
+	if (res >= 0 && mo->auto_unmount) {
+		if(0 > fuse_mount_fusermount_setup_auto_unmount_only(mountpoint, 0)) {
+			// Something went wrong, let's umount like in fuse_mount_sys.
+			umount2(mountpoint, 2); /* lazy umount */
```
matthiasgoergens, line 648: I'm not sure why we don't use `MNT_DETACH` (instead of 2) in `fuse_mount_sys`.  But I just copied the usage here.

util/fusermount.c
```diff
@@ -1347,6 +1347,7 @@ static void usage(void)
 	       " -V		    print version\n"
 	       " -o opt[,opt...]    mount options\n"
 	       " -u		    unmount\n"
+	       " -U 		    setup auto-unmount only\n"
```
matthiasgoergens, line 1350: I also thought of re-using the combination of `--unmount` with `-o auto_unmount`.

That would have been neater, but I decided not accidentally breaking existing users of `fusermount3` was more important.

util/fusermount.c
```diff
@@ -1347,6 +1347,7 @@ static void usage(void)
 	       " -V		    print version\n"
 	       " -o opt[,opt...]    mount options\n"
```
matthiasgoergens, line 1348: Btw, any reason why `-o` doesn't have a long version?  I like using long versions in bash scripts, as they are more self-documenting.

util/fusermount.c
```diff
@@ -1374,9 +1375,11 @@ int main(int argc, char *argv[])
 	int cfd;
 	const char *opts = "";
 	const char *type = NULL;
+	int setup_auto_unmount_only = 0;
 
 	static const struct option long_opts[] = {
 		{"unmount", no_argument, NULL, 'u'},
+		{"setup-auto-unmount-only", no_argument, NULL, 'U'},
```
matthiasgoergens, line 1382: I'm more than open to other suggestions for how the option (both long and short versions) should be named.

matthiasgoergens:
Btw, I'm a bit worried about race conditions for auto_unmount in general.  After all, `auto_unmount` identifies the mountpoint by path only (both in my new usage, and in established usage).  Couldn't something happen in between check and unmount?

bsbernd:
Yeah, I had scanned through your patches and the code and also seen the race.

Nikratio:
Thanks for the patch. I am generally not a big fan of `auto_unmount`, and I don't think I would add this feature today (but it existing we're unfortunately stuck with it).

I've just pushed a commit that explains the drawbacks in more detail: #763

Nikratio (changes requested):

lib/mount.c
matthiasgoergens: Will do.  I can also make a new PR to use MNT_DETACH in the other place.

util/fusermount.c
matthiasgoergens: Will do!

matthiasgoergens: https://github.com/libfuse/libfuse/pull/764

lib/mount.c
matthiasgoergens: Hmm, I leave it as `2` because I am not sure the BSDs have `MNT_DETACH`.

Nikratio: I think this was just introduced in mount_compat.h, so that should be fine.

Nikratio:
Could you also add a unit tests that checks if this works?

matthiasgoergens:
Yes, working on it.

I discovered a few minor issues when looking at the tests. You've already seen the PRs.

matthiasgoergens:
@Nikratio I added a test.

Nikratio:
Looks like the test discovered some memory leaks.

matthiasgoergens:
Yes, I'll check that.  Perhaps it's one I just introduced, even.

example/passthrough_ll.c
```diff
@@ -1240,7 +1240,11 @@ int main(int argc, char *argv[])
 		}
 
 	} else {
-		lo.source = "/";
+		lo.source = strdup("/");
```
matthiasgoergens, line 1243: The true-branch of this if gives us a malloc-ed string.  So for symmetry, we make this one a malloced on, too.  So we can `free` without a care further down.

matthiasgoergens:
@Nikratio Does this one need more changes?

test/test_examples.py
```diff
@@ -372,6 +372,30 @@ def test_notify_inval_entry(tmpdir, only_expire, notify, output_checker):
     else:
         umount(mount_process, mnt_dir)
 
+def test_dev_auto_unmount(short_tmpdir, output_checker):
```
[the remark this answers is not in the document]
matthiasgoergens: You are right.

I kept it in to serve as a dual purpose: making sure that we didn't accidentally enable 'dev' for non-root. But perhaps that's too confusing, and I'll just skip instead like you suggest.

Nikratio (approved):

PR #763, opened 2023-04-04
Document risks of auto_unmount

Nikratio:


matthiasgoergens:
Thanks!

Do you know whether the Linux kernel has enough tools to support something like auto_unmount safely and without the possibilities of races?  Ie some combination of syscalls and setup?

Nikratio:
I can't think of a way to implement this reliably with the means that I'm aware of. However, I think it should be possible (and not too hard) to implement this fully in kernel space, i.e. pass the mount option to the FUSE kernel module and let it be responsible for unmounting the filesystem when the FUSE fd no longer has a userspace reader.

matthiasgoergens:
@Nikratio Thanks for the explanation!

A related question: do you know how the numbers in `/sys/fs/fuse/connections/` are assigned?  Or specifically, how can my userspace daemon find out its number?  (I'm wondering if we can use those mechanisms to cobble together a race-free userspace auto_unmount without any modification in the kernel?)

bsbernd:
In inode.c `fc->dev = sb->s_dev`, sb->s_dev is is assigned outside of fuse (fs/super.c `set_bdev_super`), then `fuse_ctl_add_conn()` in control.c: `sprintf(name, "%u", fc->dev);`

That number is also shown as minor in `/proc/self/mountinfo` (3rd column). I don't know how one can make this race free. Isn't what you basically want a  syscall that would take a unique per mount/sb number?

matthiasgoergens:
@bsbernd Thanks for the quick and useful reply!

> Isn't what you basically want a syscall that would take a unique per mount/sb number?

That would be one way to try and implement this.  Alternatively, when you mount your filesystem, you might also be handed some kind of handle (perhaps a file-descriptor to the relevant directory in `/proc/self/mountinfo`?)

I'm still just speculating, and don't know too much.  Perhaps the design that @Nikratio sketched of just telling the kernel to unmount when your FUSE fd no longer has a user-space reader would be the most practical.

Please pardon another question:

https://github.com/torvalds/linux/blob/76f598ba7d8e2bfb4855b5298caedd5af0c374a8/include/uapi/linux/fuse.h#L571 has a handful more opcodes than we support in libfuse in https://github.com/libfuse/libfuse/blob/4cf25c27f89754e0e0310cb430dee0ddd9493c70/include/fuse_kernel.h#L397

Are there other (open source) libraries that implement the fuse protocol that support those?  Or who is using them?

bsbernd:
Any kind of n new umount approach is not only related to fuse, but is vfs. I'm not sure if the way @Nikratio  suggested would be feasible either - basically a vfs plugin shall remove itself - I don't think it has the authority.
If at all you want to ask about this on linux-fsdevel.

Other fuse kernel users I know off

https://gitlab.com/virtio-fs/virtiofsd
https://github.com/cloud-hypervisor/fuse-backend-rs
gluster

matthiasgoergens:
Thanks!  I've actually looked into the Rust `fuse-backend-rs` first and that let me here.  I'm working on https://github.com/cloud-hypervisor/fuse-backend-rs/issues/76 for that one.

I'll ask on linux-fsdevel.  Thanks for your patience!

Nikratio:
There's also Gofuse, I believe.

I'm curious why you're interested in auto-unmount. So you have a use case where it's desirable for the filesystem to look empty when it crashes?

matthiasgoergens:
Pretty much.

In our use case we had a Fuse Daemon that displayed some dynamic data on demand. We mounted it over a directory that contained a static approximation of some of that data as plain old files (and directories).

So that if/when the fuse Daemon crashes, user programs still work on a best effort basis.

(To give more details: the fuse daemon gave access to a git repository as if every branch, tag and commit had been checked out to their own directories simultaneously. It also periodically fetched new commits from a server.

The underlying normal filesystem just kept master checked out and up to date.

The repository we were interested in had a lot of very big files, so we benefitted from giving identical git blobs (ie files) the same inode, even if they were in different commits or directories.)

PR #766, opened 2023-04-07
Fix meson deprecation warning also in FreeBSD

matthiasgoergens:
See ed9be128370c93b93895bb1bb61cc1e606277613

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -8,5 +8,5 @@ task:
   script:
     - mkdir build
     - cd build
-    - meson ..
+    - meson setup ..
     - ninja


bsbernd:
Out of interest, what exactly is using this file

matthiasgoergens:
One of the two files is used by https://github.com/libfuse/libfuse/blob/master/.github/workflows/pr-ci.yml

bsbernd:
I meant cirrus.yml - is that left over from travis?

matthiasgoergens:
Cirrus seems to be its own thing, different from Travis. As far as I can tell.

Nikratio (approved):

Issue #769, opened 2023-04-07
`example/notify_inval_entry --only-expire` fails on modern kernels


matthiasgoergens:
On Linux 5.15.49 `example/notify_inval_entry --only-expire` works just fine, but on 6.2.9 it fails as follows:

```
~/p/f/l/build (master-ful)$ ./example/notify_inval_entry --only-expire -s -d -f example/mountpoint
[snip]
notify_inval_entry: ../example/notify_inval_entry.c:261: update_fs_loop: Assertion `fuse_lowlevel_notify_expire_entry (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY) == 0' failed.
Aborted (core dumped)
```

I strongly suspect https://github.com/torvalds/linux/commit/4f8d37020e1fd0bf6ee9381ba918135ef3712efd is the responsible kernel commit.

(The two kernel versions in question are what I'm getting from running Docker on a MacOSX laptop for the old kernel, and my Archlinux desktop for the new kernel.)

I noticed the problem, because `test/test_examples.py::test_notify_inval_entry[True-expire_entries] ` fails on the newer kernel.

I'm working on understanding what's happening and writing a fix.

matthiasgoergens:
`test_notify1` in `test_ctests.py` might also be affected, but I'm not so sure.

Nikratio:
I believe this has since been fixed in the above commits.

PR #771, opened 2023-04-08
Fix notify_inval_entry --only-expire

matthiasgoergens:
On Linux 5.15.49 `example/notify_inval_entry --only-expire` works just fine, but on 6.2.9 it fails as follows:

```
$ ./example/notify_inval_entry --only-expire -s -d -f example/mountpoint
[.. snip ..]
notify_inval_entry: ../example/notify_inval_entry.c:261: update_fs_loop: Assertion `fuse_lowlevel_notify_expire_entry (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY) == 0' failed.
Aborted (core dumped)
```

I strongly suspect https://github.com/torvalds/linux/commit/4f8d37020e1fd0bf6ee9381ba918135ef3712efd is the responsible kernel commit.

(The two kernel versions in question are what I'm getting from running Docker on a MacOSX laptop for the old kernel, and my Archlinux desktop for the new kernel.)

I noticed the problem, because `test/test_examples.py::test_notify_inval_entry[True-expire_entries] ` fails on the newer kernel.

Fixes https://github.com/libfuse/libfuse/issues/769

### DIFF
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -229,13 +229,10 @@ static void tfs_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
 }
 
 static void tfs_init(void *userdata, struct fuse_conn_info *conn) {
-    (void) userdata;
     if(options.only_expire && !(conn->capable & FUSE_CAP_EXPIRE_ONLY)) {
         fprintf(stderr, "FUSE_CAP_EXPIRE_ONLY not supported by kernel\n");
-        exit(1);
+        fuse_session_exit(*(struct fuse_session **) userdata);
     }
-    if(options.only_expire)
-        conn->want |= FUSE_CAP_EXPIRE_ONLY;
 }
 
 static const struct fuse_lowlevel_ops tfs_oper = {
@@ -327,7 +324,7 @@ int main(int argc, char *argv[]) {
     update_fs();
 
     se = fuse_session_new(&args, &tfs_oper,
-                          sizeof(tfs_oper), NULL);
+                          sizeof(tfs_oper), &se);
     if (se == NULL)
         goto err_out1;
 
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -93,6 +93,7 @@
 static char file_name[MAX_STR_LEN];
 static fuse_ino_t file_ino = 2;
 static int lookup_cnt = 0;
+static int old_lookup_cnt = 0;
 
 /* Command line parsing */
 struct options {
@@ -266,12 +267,16 @@ static void* update_fs_loop(void *data) {
         update_fs();
         if (!options.no_notify && lookup_cnt) {
             if(options.only_expire) {
-                assert(fuse_lowlevel_notify_expire_entry
-                   (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY) == 0);
-                // For some reason, the kernel doesn't send a forget request,
-                // so we have to do it ourselves.
-                // (At least that's the behaviour on 6.2.9)
-                lookup_cnt = 0;
+                int ret = fuse_lowlevel_notify_expire_entry
+                   (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY);
+               // forget is not being called
+               // the dentry is running just into a timeout
+               if (old_lookup_cnt == lookup_cnt) {
+                 assert(ret == -ENOENT);
+               } else {
+                 assert(ret == 0);
+                 old_lookup_cnt = lookup_cnt;
+               }
             } else {
                 assert(fuse_lowlevel_notify_inval_entry
                       (se, FUSE_ROOT_ID, old_name, strlen(old_name)) == 0);
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -228,7 +228,18 @@ static void tfs_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
     }
 }
 
+static void tfs_init(void *userdata, struct fuse_conn_info *conn) {
+    (void) userdata;
+    if(options.only_expire && !(conn->capable & FUSE_CAP_EXPIRE_ONLY)) {
+        fprintf(stderr, "FUSE_CAP_EXPIRE_ONLY not supported by kernel\n");
+        exit(1);
+    }
+    if(options.only_expire)
+        conn->want |= FUSE_CAP_EXPIRE_ONLY;
+}
+
 static const struct fuse_lowlevel_ops tfs_oper = {
+    .init	= tfs_init,
     .lookup	= tfs_lookup,
     .getattr	= tfs_getattr,
     .readdir	= tfs_readdir,
@@ -260,6 +271,10 @@ static void* update_fs_loop(void *data) {
             if(options.only_expire) {
                 assert(fuse_lowlevel_notify_expire_entry
                    (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY) == 0);
+                // For some reason, the kernel doesn't send a forget request,
+                // so we have to do it ourselves.
+                // (At least that's the behaviour on 6.2.9)
+                lookup_cnt = 0;
             } else {
                 assert(fuse_lowlevel_notify_inval_entry
                       (se, FUSE_ROOT_ID, old_name, strlen(old_name)) == 0);
--- a/test/test_examples.py
+++ b/test/test_examples.py
@@ -347,6 +347,8 @@ def test_notify_inval_entry(tmpdir, only_expire, notify, output_checker):
         cmdline.append('--no-notify')
     if only_expire == "expire_entries":
         cmdline.append('--only-expire')
+        if fuse_proto < (7,38):
+            pytest.skip('only-expire not supported by running kernel')
     mount_process = subprocess.Popen(cmdline, stdout=output_checker.fd,
                                      stderr=output_checker.fd)
     try:


example/notify_inval_entry.c
```diff
@@ -260,6 +260,10 @@ static void* update_fs_loop(void *data) {
             if(options.only_expire) {
                 assert(fuse_lowlevel_notify_expire_entry
                    (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY) == 0);
```
matthiasgoergens, line 262: I'm not quite sure whether manually fiddling with `lookup_cnt` is the right thing to do.  Because as far as I can tell on older kernels this the flag `FUSE_LL_EXPIRE_ONLY` is silently ignored, and thus we get behaviour identical to the `else` branch and do get a forget request.

I'm not sure whether _not_ getting the forget request from the kernel is intentional?

Without this patch to reset `lookup_cnt`, we call `fuse_lowlevel_notify_expire_entry` even when the kernel has already forgotten about us, and thus we get an `-ENOENT` from the kernel instead of a `0`.

So an alternative fix for this example would be to change the assert to allow 0 or `-ENOENT`, and only in the latter case set `lookup_cnt` to `0`.

Keep in mind that this alternative fix would have us send one redundant `fuse_lowlevel_notify_expire_entry`.  (First the one that's actually useful, then the one that fails with `-ENOENT`.)

example/notify_inval_entry.c
```diff
@@ -260,6 +271,10 @@ static void* update_fs_loop(void *data) {
...
+                // For some reason, the kernel doesn't send a forget request,
+                // so we have to do it ourselves.
+                // (At least that's the behaviour on 6.2.9)
+                lookup_cnt = 0;
```
bsbernd, line 277: Kernel will send a forget request once it gives up the corresponding inode. So now the invalidate request sends FUSE_EXPIRE_ONLY, which basically suppresses d_invalidate() on the kernel side.

`fuse_reverse_inval_entry()`
```

	fuse_dir_changed(parent);
	if (!(flags & FUSE_EXPIRE_ONLY))
		d_invalidate(entry);
	fuse_invalidate_entry_cache(entry);

	if (child_nodeid != 0 && d_really_is_positive(entry)) {
```
The caller (`fuse_notify_inval_entry`) does not set `child_nodeid`, so we can ignore that code path - the request just invalidates the fuse cache. On _next_ file access (stat/open/whatever) vfs will trigger a revalidate and if that fails dentry will be removed. Once the last inode reference is dropped a forget will be eventually send.

Regarding the test - without file access in between, lookup is not dropped - easiest way to get that would be something like `stat(old_name, ...)`.
Also, lookup_cnt is not lock protected - if anything would run in parallel - test will give unexpected results.

example/notify_inval_entry.c
```diff
@@ -228,7 +228,18 @@ static void tfs_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
...
 
+static void tfs_init(void *userdata, struct fuse_conn_info *conn) {
+    (void) userdata;
+    if(options.only_expire && !(conn->capable & FUSE_CAP_EXPIRE_ONLY)) {
+        fprintf(stderr, "FUSE_CAP_EXPIRE_ONLY not supported by kernel\n");
+        exit(1);
+    }
+    if(options.only_expire)
+        conn->want |= FUSE_CAP_EXPIRE_ONLY;
+}
```
bsbernd, line 239: What is this change about? This is not used in fuse_lowlevel.c `do_init`. See below my comment about kernel side - notifying kernel side that daemon side supports the extended struct would make sense, but then I really do not see why this could not be done in general in `do_init`

example/notify_inval_entry.c
bsbernd: And while we are at it, looking at kernel code, I'm worried about a missing test if daemon side supports flags at all - it seems to rely on that the buffer was zeroed in previous lib versions. Now that is implementation dependent - what guarantees zeroing of a padded value?

bsbernd (changes requested):

example/notify_inval_entry.c
matthiasgoergens: Sorry, I just adapted this from another example, and didn't get an error.

Happy to fix it, if you have a suggestion.

example/notify_inval_entry.c
matthiasgoergens: > Also, lookup_cnt is not lock protected - if anything would run in parallel - test will give unexpected results.

Yes, that's already a problem with the existing code, I think?  Might be worth a separate PR.

(I'm still digesting the rest of what you explained.)

matthiasgoergens: > And while we are at it, looking at kernel code, I'm worried about a missing test if daemon side supports flags at all - it seems to rely on that the buffer was zeroed in previous lib versions. Now that is implementation dependent - what guarantees zeroing of a padded value?

Depending on exactly how you allocate your struct, in most cases using uninitialized variables (or buffers) in C is much worse than just getting the occasional unexpected value: it's undefined behaviour, and your program might do anything; including formatting your hard drive or sending embarrassing emails.

Luckily, that also means that tools like Valgrind are on the lookout for such omissions.  It seems like the commit 24b35c3d97ffdbf0a1f8e8b4e94ed892343603a6 that introduces `fuse_lowlevel_notify_inval_entry` already has `outarg.padding = 0;`.

You can see the evolution of that function with `git log -L :fuse_lowlevel_notify_inval_entry:lib/fuse_lowlevel.c`.

If you run `git log --pickaxe-regex -S 'padding *= *0' --stat --patch`, you can see a few more instances where valgrind lead to more careful coding, like c2a47b08afdfb75bbdd7dec62336bd0a04346628 or df0d539d04247dbbfcb10c364638df7edaef697e.

example/notify_inval_entry.c
matthiasgoergens: I removed it, as the kernel doesn't actually support this.  Thanks!

example/notify_inval_entry.c
matthiasgoergens: > [...] Once the last inode reference is dropped a forget will be eventually send.

Ok, it seems to properly exercise this functionality, we should probably have more than one inode?

example/notify_inval_entry.c
matthiasgoergens: I still want to exit the filesystem, when the kernel doesn't support what the user asked for.  So I adapted the mechanism from `printcap.c`.

matthiasgoergens:
Hmm, the failure in https://github.com/libfuse/libfuse/actions/runs/4654417338/jobs/8236136337?pr=771 looks suspicious:

```
==================================== ERRORS ====================================
_________________ ERROR at teardown of test_write_cache[True] __________________
Traceback (most recent call last):
  File "/tmp/libfuse-build-yWuLEt/build-address/test/conftest.py", line 87, in output_checker
    checker._check()
  File "/tmp/libfuse-build-yWuLEt/build-address/test/conftest.py", line 80, in _check
    raise AssertionError('Suspicious output to stderr (matched "%s")'
AssertionError: Suspicious output to stderr (matched "ERROR")
----------------------------- Captured stdout call -----------------------------
ERROR: Expected 8192 bytes, got 4096
!ERROR: Expected 8192 bytes, got 4096
!test_write_cache: ../../../home/runner/work/libfuse/libfuse/test/test_write_cache.c:271: int main(int, char **): Assertion `got_write == 1' failed.
=================================== FAILURES ===================================
____________________________ test_write_cache[True] ____________________________
Traceback (most recent call last):
  File "/tmp/libfuse-build-yWuLEt/build-address/test/test_ctests.py", line 41, in test_write_cache
    subprocess.check_call(cmdline, stdout=output_checker.fd, stderr=output_checker.fd)
  File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/tmp/libfuse-build-yWuLEt/build-address/test/../test/test_write_cache', '/tmp/pytest-of-root/pytest-1/test_write_cache_True_0', '-owriteback_cache']' died with <Signals.SIGABRT: 6>.
----------------------------- Captured stdout call -----------------------------
ERROR: Expected 8192 bytes, got 4096
!ERROR: Expected 8192 bytes, got 4096
!test_write_cache: ../../../home/runner/work/libfuse/libfuse/test/test_write_cache.c:271: int main(int, char **): Assertion `got_write == 1' failed.
```

I don't think I touched anything related to `test_write_cache`.

Nikratio:
@HereThereBeDragons Could you take a look at this? I think the test case from your commit c0a344e37978448 may not be working as intended

example/notify_inval_entry.c
```diff
@@ -327,7 +324,7 @@ int main(int argc, char *argv[]) {
     update_fs();
 
     se = fuse_session_new(&args, &tfs_oper,
-                          sizeof(tfs_oper), NULL);
+                          sizeof(tfs_oper), &se);
```
bsbernd, line 327: Dunno, yeah it works, but it makes the code complex? Use container_of() in tfs_init() to retrieve 'struct fuse_session' from 'struct fuse_conn_info'?

bsbernd (changes requested):

example/notify_inval_entry.c
matthiasgoergens: There are copy-and-pasted `container_of` definitions in `fuse.c` and `fuse_lowlevel.c`:

```c
#define container_of(ptr, type, member) ({                              \
			const typeof( ((type *)0)->member ) *__mptr = (ptr); \
			(type *)( (char *)__mptr - offsetof(type,member) );})
```

I could copy-and-paste them into this file as well, and use them.  But this seems rather more complicated to me than using straight-forward pointers?  Especially since the `userdata` parameter seems to be designed for exactly the kind of 'passing-of-extra-info' that we are doing here?

In any case, I'm happy to use copy and use that macro, if you like it.  (Or we re-organise and stick it into a common internal header, even.)

If I understand that macro right, it would only work in `tfs_init` for as long as the `struct fuse_conn_info` we are getting is really a part of a `struct fuse_session`?  We'd need to document that this is now a requirement and something that the API would never be allowed to change?

bsbernd: I guess the macro could go into fuse_common.h. I only yesterday realized libfuse already has it and had placed another copy into my uring patches. And yes, it only works as long as fuse_conn_info is part of fuse_session. From my point of view stopping correctly in the ->init() method is a common issue - I know of at least two other file system exiting there (one opensource is unionfs-fuse - I had placed an exit there  myself) and typically one stores local information into `userdata`, see for example passthrough_ll.c. 
So at least for me using `container_of` seems to be the better example. And then documenting that one can do that would be helpful to others as well. @Nikratio  what do you think?

example/notify_inval_entry.c
bsbernd: > > And while we are at it, looking at kernel code, I'm worried about a missing test if daemon side supports flags at all - it seems to rely on that the buffer was zeroed in previous lib versions. Now that is implementation dependent - what guarantees zeroing of a padded value?
> 
> Depending on exactly how you allocate your struct, in most cases using uninitialized variables (or buffers) in C is much worse than just getting the occasional unexpected value: it's undefined behaviour, and your program might do anything; including formatting your hard drive or sending embarrassing emails.
> 
> Luckily, that also means that tools like Valgrind are on the lookout for such omissions. It seems like the commit [24b35c3](https://github.com/libfuse/libfuse/commit/24b35c3d97ffdbf0a1f8e8b4e94ed892343603a6) that introduces `fuse_lowlevel_notify_inval_entry` already has `outarg.padding = 0;`. That commit's from 2009.
> 
> You can see the evolution of that function with `git log -L :fuse_lowlevel_notify_inval_entry:lib/fuse_lowlevel.c`. (EDIT: Or rather `git log -L :fuse_lowlevel_notify_expire_entry:lib/fuse_lowlevel.c` because of a rename.)
> 
> If you run `git log --pickaxe-regex -S 'padding *= *0' --stat --patch`, you can see a few more instances where valgrind lead to more careful coding, like [c2a47b0](https://github.com/libfuse/libfuse/commit/c2a47b08afdfb75bbdd7dec62336bd0a04346628) or [df0d539](https://github.com/libfuse/libfuse/commit/df0d539d04247dbbfcb10c364638df7edaef697e).

That is libfuse, what about all other (including possibly unknown) fuse users? The kernel code assumes the buffer is zeroed - I don't think that requirement is documented anywhere, so implementation depending... Reminder to myself, I need to be more active with fuse kernel patch review ;) This already the 2nd instance of such an issue. 
https://patchwork.kernel.org/project/linux-fsdevel/patch/165002363635.1457422.5930635235733982079.stgit@localhost/#24825548

bsbernd:
@HereThereBeDragons Why so complicated and not simply a stat(old_name), which will trigger a revalidate and set lookup_cnt to 0?
And isn't the entire test not reflecting the purpose of  `EXPIRE_ONLY`? From my understanding the point is to keep the inode - i.e. something changes inode attributes and `stat/getattr` is supposed to see these changes immediately. The test only checks if FUSE_LL_EXPIRE_ONLY works, but does not check the result?

matthiasgoergens:
> And isn't the entire test not reflecting the purpose of `EXPIRE_ONLY`? From my understanding the point is to keep the inode - i.e. something changes inode attributes and `stat/getattr` is supposed to see these changes immediately. The test only checks if FUSE_LL_EXPIRE_ONLY works, but does not check the result?

I might be able to answer that one.  c0a344e3797844896d04efc4b565a2627067b67f says

> Test for fuse_lowlevel_notify_expire_entry.
> This test is too simple to check for all functionalities of notify_expire as it always successfully passes when libfuse supports the function (even if kernel does not support it -  it just takes it as notify_inval)

So that test was really just a bare minimum effort to exercise the functionality, but not check whether it actually works properly.

Well, and my change here was just the bare minimum required to make the test pass on newer kernels.

As you suggest, we should probably make a test that actually checks that things work.

I'll have a look at @HereThereBeDragons's suggestion.

matthiasgoergens:
@HereThereBeDragons I changed the PR to your suggestion for how to fix the example.

example/notify_inval_entry.c
Nikratio: A `fuse session` instance is required not only to exit the session, but also for the various `fuse_notify_*` functions. So I think the problem of getting this instance is not limited to just the `init` callback.

I believe using `container_of` is common in kernel-code, but I'm not a friend of it at all since it leaks implementation details.

In this particular example, I think we can simply make `fuse_session` a global variable and be done with it.

Where global variables are not desirable, I think the `userdata` mechanism should be used for this. I would expect that in most cases this is a struct where a new session member can easily be added and carried around (this seems to apply to passthrough_ll.c as well).

Nikratio:
> @Nikratio Yes the test is broken. But I would suggest a different solution to this fix here.
> 
> I would suggest is to change how the interaction is with `lookup_cnt` inside the loop:
> 
> As we know how `lookup_cnt` changes related to `invalidate` vs `expire`:
> 
>     * `invalidate` puts `lookup_cnt` to `0`
> 
>     * `expire` does not reset `lookup_cnt` and as such the `if` inside the loop is called always because `lookup_cnt` is not 0.
> 
>     * We know a system can run libfuse with `expire` but if the kernel does not support it it behaves like `invalidate` and as such resets the `lookup_cnt`

This sounds fundamentally broken to me, not just for the purposes of writing a unit test, but also for using it in production.

How can you correctly use invalidate() if you can't tell whether it's modifying the lookup count?

Furthermore, how do you prevent race conditions with simultaneous lookup count increases? The reason that FUSE processes have to handle the lookup count at all is that it is not possible to signal (in a race free way) when the kernel is no longer aware of an inode.

I hope I am missing something here?

HereThereBeDragons:
> 



> > @Nikratio Yes the test is broken. But I would suggest a different solution to this fix here.
> > I would suggest is to change how the interaction is with `lookup_cnt` inside the loop:
> > As we know how `lookup_cnt` changes related to `invalidate` vs `expire`:
> > ```
> > * `invalidate` puts `lookup_cnt` to `0`
> > 
> > * `expire` does not reset `lookup_cnt` and as such the `if` inside the loop is called always because `lookup_cnt` is not 0.
> > 
> > * We know a system can run libfuse with `expire` but if the kernel does not support it it behaves like `invalidate` and as such resets the `lookup_cnt`
> > ```
> 
> This sounds fundamentally broken to me, not just for the purposes of writing a unit test, but also for using it in production.
> 
> How can you correctly use invalidate() if you can't tell whether it's modifying the lookup count?
> 
> Furthermore, how do you prevent race conditions with simultaneous lookup count increases? The reason that FUSE processes have to handle the lookup count at all is that it is not possible to signal (in a race free way) when the kernel is no longer aware of an inode.
> 
> I hope I am missing something here?

i think i expected some pre-knowledge about how this example works:
when ever `invalidate` is called, `forget` gets executed which decreases the `lookup_cnt` by 1. as in this example there is only ever 1 file, the `lookup_cnt` is either 1 or 0 (if not running `expire` with a kernel that supports `expire`).

Nikratio:
> > 
> 
> > > @Nikratio Yes the test is broken. But I would suggest a different solution to this fix here.
> > > I would suggest is to change how the interaction is with `lookup_cnt` inside the loop:
> > > As we know how `lookup_cnt` changes related to `invalidate` vs `expire`:
> > > ```
> > > * `invalidate` puts `lookup_cnt` to `0`
> > > 
> > > * `expire` does not reset `lookup_cnt` and as such the `if` inside the loop is called always because `lookup_cnt` is not 0.
> > > 
> > > * We know a system can run libfuse with `expire` but if the kernel does not support it it behaves like `invalidate` and as such resets the `lookup_cnt`
> > > ```
> > 
> > 
> > This sounds fundamentally broken to me, not just for the purposes of writing a unit test, but also for using it in production.
> > How can you correctly use invalidate() if you can't tell whether it's modifying the lookup count?
> > Furthermore, how do you prevent race conditions with simultaneous lookup count increases? The reason that FUSE processes have to handle the lookup count at all is that it is not possible to signal (in a race free way) when the kernel is no longer aware of an inode.
> > I hope I am missing something here?
> 
> i think i expected some pre-knowledge about how this example works: when ever `invalidate` is called, `forget` gets executed which decreases the `lookup_cnt` by 1.

Ah, ok. That makes more sense. Thanks for clarifying.

> as in this example there is only ever 1 file, the `lookup_cnt` is either 1 or 0 (if not running `expire` with a kernel that supports `expire`).

That is probably the case *most* of the time, but not guaranteed. The kernel may issue forgot followed by lookup(), but FUSE userspace may process lookup() before forget(), resulting in an intermittent lookup count of 2.

bsbernd:
I was suggesting to add stat(old_file, ) for exactly the reason to have a simple and a sane state, but then I don't like the entire test at all. The non-atomic and non-locked lookup_cnt makes things even worse. One should also consider that anything running in parallel (like cron job doing update_db) will trigger parallel lookups...

Nikratio:
As a start, could we make `fuse_lowlevel_notify_expire_entry` fail if it's not supported by the kernel (instead of silently falling back to `fuse_lowlevel_notify_inval_entry`?

bsbernd:
I just tested how it behaves

```
bschubert2@imesrv1 build-rhel8>example/notify_inval_entry  -f --update-interval=1 --timeout=5 --only-expire /tmp/test
inc lookup_cnt: 1
inc lookup_cnt: 2
inc lookup_cnt: 3
inc lookup_cnt: 4
inc lookup_cnt: 5
inc lookup_cnt: 6
```

--> every of these lookup_cnt increases represents an `ls -l /tmp/test`

Then on `[root@imesrv1 ~]# echo 3 >/proc/sys/vm/drop_caches`:

```
`dec (6) lookup_cnt: 0`

So I was actually wrong, `stat(old_name, )` does not trigger a lookup_cnt decrease - vfs negative dentry cache...
@matthiasgoergens  @HereThereBeDragons  last version of your patches is indeed the best simple solution, but random system events that trigger a kernel inode release might trigger weird results.

Nikratio:
> > As a start, could we make `fuse_lowlevel_notify_expire_entry` fail if it's not supported by the kernel (instead of silently falling back to `fuse_lowlevel_notify_inval_entry`?
> 
> You mean for the test? That could be done. I would suggest adding an extra flag that adds an `assert` if `expire` does an `invalidate`
> 
> Or you mean in general? I guess it depends on what kind of behavior people are expecting. Were there other cases before where such exension where introduced? How do they behave? In case we want the failure I guess we would then need to add an extra check when setting the expire flag if `se->conn.capable` includes `FUSE_CAP_EXPIRE_ONLY`?

I mean in general. I do not see any advantages of an implicit fallback but the clear drawback that the application cannot rely on specific semantics. Where this is desired, the application can always fallback deliberately.

Nikratio:
@HereThereBeDragons Any update on this?

matthiasgoergens:
Sorry, I was busy with other stuff. No update from me at this point.

HereThereBeDragons:
> @HereThereBeDragons Any update on this?

PR #789  is the fix for it. i am waiting for your reply if the base function (`fuse_lowlevel_notify_entry`) that always falls back to `invalidate` should be available to people or not.

Nikratio:
Thanks! I'll close this PR then.

PR #773, opened 2023-04-11
Stop suppressing memory leak detection

matthiasgoergens:
In order to fix the memory leaks, I tried to just disable leak suppression, and see what breaks.  But from https://github.com/libfuse/libfuse/pull/762 it looks like our suppression didn't actually work; and we didn't actually see any memory leaks in the tests anyway.

### DIFF
--- a/test/lsan_suppress.txt
+++ b/test/lsan_suppress.txt
@@ -1,11 +1,1 @@
 # Suppression file for address sanitizer.
-
-# There are some leaks in command line option parsing. They should be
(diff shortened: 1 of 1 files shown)


Nikratio:
I assure you that they did work in the past :-). But it's great that we no longer need them.

Nikratio:
There seems to be some odd merge conflict here, could you take a look?

matthiasgoergens:
Oh, the merge-conflict was because I fixed a typo in this file in another PR.

Issue #746, opened 2023-03-16
Race condition in high-level API's may leave behind  .fuse_hiddenXXX files


devPeri:
# Issue Description
When an open file is closed and immediately deleted and the entry is cached, FUSE_RELEASE and FUSE_UNLINK are sent back to back. 

Handling FUSE_RELEASE, `fuse_do_release()` calls `fuse_fs_release()` outside the lock.

If `fuse_fs_release()` is slow to complete, the FUSE_UNLINK will arrive and start processing in another thread.
`fuse_lib_unlink()` will see that the file is open and call `fuse_fs_rename()` via `hide_node()` to rename the file to `.fuse_hiddenXXX`.

`fuse_fs_rename()` is also called outside the lock and while it is processing, the `fuse_fs_release()` has completed and `fuse_do_release()` acquires the lock and continues processing.

As `fuse_fs_rename()` is still processing, `rename_node()` has not been called yet and thus `node->is_hidden` is still `0`.

With `fuse_do_release()` now processing with the lock, `node->open_count` is decremented to `0` but because `node->is_hidden` is still `0`  the `unlink_hidden` flag is never set to `1` and thus the file is considered as closed but not deleted.

`fuse_do_release()` has now finished completely, and `fuse_lib_unlink()` is still waiting for `fuse_fs_rename()` to complete.
Once `fuse_fs_rename()` completes, the processing resumes and `rename_node()` is called to hide the node etc.

**But by this time the file has already been closed in the user-space and the file has been renamed to `.fuse_hiddenXXX`.
As the file is already closed, `fuse_do_release()` will not be called to clean-up the `.fuse_hiddenXXX` file.**

If entry caching is not used, between FUSE_RELEASE and FUSE_UNLINK there is a whole bunch of FUSE_GETATTR happening that allow FUSE_RELEASE to complete processing before FUSE_UNLINK starts.

## Code flaw
```
static void fuse_do_release(struct fuse *f, fuse_ino_t ino, const char *path,
                            struct fuse_file_info *fi)
{
        struct node *node;
        int unlink_hidden = 0;

        fuse_fs_release(f->fs, path, fi);        <--- called here without any entry lock protection

        pthread_mutex_lock(&f->lock);
        ...
```

# Potential fix
**Call fuse_fs_release() after the lock is acquired.**
I don't know if this is the best solution or if there is reason why it should not be called with the lock held.

# Sample debug logs
[...]

PR #779, opened 2023-04-12
Fix race condition

matthiasgoergens:
Fixes https://github.com/libfuse/libfuse/issues/746

This is an experimental commit, to see whether the fix suggested in the linked issue works.


bsbernd:
@matthiasgoergens I think this can closed?

matthiasgoergens:
I think so, yes.

PR #780, opened 2023-04-13
FUSE loop: fix a race condition when free memory

anatasluo:
When I call the fuse_main_real, I met the following error: free(): invalid pointer

The problem here is that I received a signal between fuse_set_signal_handlers and fuse_loop.

So when free takes effect, fbuf.mem is still null.

### DIFF
--- a/lib/fuse_loop.c
+++ b/lib/fuse_loop.c
@@ -34,7 +34,8 @@ int fuse_session_loop(struct fuse_session *se)
 		fuse_session_process_buf_int(se, &fbuf, NULL);
 	}
 
-	free(fbuf.mem);
+	if (fbuf.mem)
+		free(fbuf.mem);
 	if(res > 0)
 		/* No error, just the length of the most recently read
 		   request */


bsbernd (changes requested):
https://pubs.opengroup.org/onlinepubs/9699919799/functions/free.html

` If ptr is a null pointer, no action shall occur.`

So there is something else wrong - ptr is somehow an invalid (non null) pointer?

matthiasgoergens:
Could you write a test that reproduces the problem you are encountering?

anatasluo:
@matthiasgoergens After I met this error, I test with the following code:
`
diff --git a/lib/fuse_loop.c b/lib/fuse_loop.c
index b9c498b..3b1963b 100644
--- a/lib/fuse_loop.c
+++ b/lib/fuse_loop.c
@@ -34,6 +34,8 @@ int fuse_session_loop(struct fuse_session *se)
diff --git a/lib/fuse_loop.c b/lib/fuse_loop.c
index b9c498b..3b1963b 100644
--- a/lib/fuse_loop.c
+++ b/lib/fuse_loop.c
@@ -34,6 +34,8 @@ int fuse_session_loop(struct fuse_session *se)
diff --git a/lib/fuse_loop.c b/lib/fuse_loop.c
index b9c498b..3b1963b 100644
--- a/lib/fuse_loop.c
+++ b/lib/fuse_loop.c
@@ -34,6 +34,8 @@ int fuse_session_loop(struct fuse_session *se)
                fuse_session_process_buf_int(se, &fbuf, NULL);
        }
 
+       if (fbuf.mem == NULL)
+               printf("NULL pointer \n");
        free(fbuf.mem);
        if(res > 0)
                /* No error, just the length of the most recently read
diff --git a/lib/helper.c b/lib/helper.c
index 14a0df3..824b4a5 100644
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -25,6 +25,7 @@
 #include <limits.h>
 #include <errno.h>
 #include <sys/param.h>
+#include <signal.h>
 
 #define FUSE_HELPER_OPT(t, p) \
        { t, offsetof(struct fuse_cmdline_opts, p), 1 }
@@ -364,6 +365,10 @@ int fuse_main_real(int argc, char *argv[], const struct fuse_operations *op,
                goto out3;
        }
 
+       kill(getpid(), SIGTERM);
+       printf("send signal to myself \n");
+       sleep(5);
+
        if (opts.singlethread)
                res = fuse_loop(fuse);
        else {

`

But, accroding to @bsbernd , free null point is allowed in the POSIX description.

However, this patch did solve my problem, I will try to check my original code and find more infomation.

anatasluo:
@bsbernd You are right, my problem is not triggered by this free function

matthiasgoergens:
I suspect the problem might be a double-free.

matthiasgoergens:
This patch gives me the leak:

```
diff --git a/lib/fuse_loop.c b/lib/fuse_loop.c
index b9c498b..a0a8029 100644
--- a/lib/fuse_loop.c
+++ b/lib/fuse_loop.c
@@ -23,6 +23,7 @@ int fuse_session_loop(struct fuse_session *se)
                .mem = NULL,
        };
 
+       fuse_session_exit(se);
        while (!fuse_session_exited(se)) {
                res = fuse_session_receive_buf_int(se, &fbuf, NULL);
 ```

```
$ ninja && ./example/hello -s -d -f mountpoint/
[9/9] Linking target example/hello_ll
FUSE library version: 3.14.1
nullpath_ok: 0

=================================================================
==178330==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 352 byte(s) in 1 object(s) allocated from:
    #0 0x7fbb19abf411 in __interceptor_calloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x7fbb1a0efd3b in fuse_fs_new ../lib/fuse.c:4814
    #2 0x7fbb1a0f02b5 in fuse_new_31 ../lib/fuse.c:4913
    #3 0x7fbb1a10ec5e in fuse_main_real ../lib/helper.c:345
    #4 0x5625db8ab418 in main ../example/hello.c:176
    #5 0x7fbb1983c78f  (/usr/lib/libc.so.6+0x2378f)

SUMMARY: AddressSanitizer: 352 byte(s) leaked in 1 allocation(s).
```

The patch simulates getting a SIGTERM just before we process the first operation (INIT) from the kernel.

It doesn't occur in `hello_ll`, but in `hello`.

**Comment by matthiasgoergens** [comment]
```c
void fuse_session_destroy(struct fuse_session *se)
{
	struct fuse_ll_pipe *llp;

	if (se->got_init && !se->got_destroy) {
		if (se->op.destroy)
			se->op.destroy(se->userdata);
	}
	llp = pthread_getspecific(se->pipe_key);
	if (llp != NULL)
		fuse_ll_pipe_free(llp);
	pthread_key_delete(se->pipe_key);
	pthread_mutex_destroy(&se->lock);
	free(se->cuse_data);
	if (se->fd != -1)
		close(se->fd);
	if (se->io != NULL)
		free(se->io);
	destroy_mount_opts(se->mo);
	free(se);
}
```

OK, I figured out the missing `free`: `fuse_session_destroy` only calls the user supplied `op.destroy`, if `init` ran.  But if you receive a signal before the loop, you never process `init`.  Unfortunately, the high level api relies on `op.destroy` to clean up after itself.

matthiasgoergens:
I created a new PR for my leak.  But I'm not sure how relevant it is for your finding, sorry.

Nikratio:
Closing this for now since it's a no-op (free(NULL) is valid).

matthiasgoergens:
@Nikratio I think the freeing of NULL was a red herring in the investigation. It was a hypothesis that turned out to be wrong.

PR #782, opened 2023-04-14
Fix issue #746.

devPeri:
Added a secondary check in fuse_lib_unlink() after hide_node() to check again under a lock if the (now hidden) file is still open. If not then delete it.

This should synchronise fuse_lib_unlink() with fuse_lib_release(), when nullpath_ok is set.


devPeri:
#### One-liner test
Start FUSE: `./passthrough_fh -f -d /mnt/temp`
Run test: ```bash -c 'T=`mktemp -d /mnt/temp/tmp/test746-XXXX` ; cd $T ; python3 -c "import os; f=open(\"test\", \"xt\"); f.close(); os.remove(\"test\");" ; ls -al'```

#### Automated test
https://github.com/devPeri/libfuse/commit/dfb3ef57c32f2df3914aaaae0220663cf4da162d
Set the `TEST746_DELAY_DISABLE` env variable to disable the delay on demand at runtime.

lib/fuse.c
```diff
@@ -2967,6 +2967,20 @@ static void fuse_lib_unlink(fuse_req_t req, fuse_ino_t parent,
...
+				if (!is_open(f, parent, wnode->name)) {
+					char *unlinkpath;
+
+					/* get the hidden file path, to unlink it */
+					if (try_get_path(f, wnode->nodeid, NULL, &unlinkpath, NULL, false) == 0) {
+						err = fuse_fs_unlink(f->fs, unlinkpath);
+						if (!err)
+							remove_node(f, parent, wnode->name);
+						free(unlinkpath);
+					}
```
bsbernd, line 2981: Make sense,  but I'm worried about `remove_node`, what is if someone opened the 'hidden' file in the mean time?  I _think_ one can open these .fuse_hidden files?

devPeri: Unless I am mistaken, the path should be locked by the previous `get_path_wrlock()`.
An `fuse_lib_open()` of the hidden file will wait for `fuse_lib_unlink()` to finish.

bsbernd: Ah right, current unlink also calls remove_node - if there would be a race, it would have been there already.

test/test746.c
Nikratio, line 1: This seems to be used only with passthrough_fh.py. Have you considered to simply have the passthrough_fh code check for TEST746_DELAY_DISABLE and, if so, wait after running os.close() and os.rename()?

test/meson.build
```diff
@@ -14,6 +14,20 @@ td += executable('readdir_inode', 'readdir_inode.c',
                  include_directories: include_dirs,
                  install: false)
 
+td += custom_target('test746.so',
```
Nikratio, line 17: Is there a reason we can't use Meson's standard rules for building shared objects (instead of custom_target)?

test/test_examples.py
```diff
@@ -442,6 +442,102 @@ def test_cuse(output_checker):
...
+    #
+    # The delay can be disabled at call-time by setting the TEST746_DELAY_DISABLE
+    # environment variable. e.g:
+    # TEST746_DELAY_DISABLE=1 python3 -m pytest test/test_examples.py::test_746
+
+    # set the path for the library to preload
+    preloadLib = base_cmdline + \
+        [ pjoin(basename, "test", "test746.so") ]
+    preloadLib = preloadLib[0]
+    #print("preloadLib: ", preloadLib)
```
Nikratio, line 482: For the final version, please remove all the commented out parts of the code (rather than just commenting them out).

test/test_examples.py
```diff
@@ -442,6 +442,102 @@ def test_cuse(output_checker):
...
+
+    # set the FUSE binary path and command
+    fuseCMD = base_cmdline + \
+        [ pjoin(basename, 'example', 'passthrough_fh'),
+        "-f", fuseMNT]
+    #print("fuseCMD: ", str(fuseCMD))
+
+    # start FUSE with our preload library
+    os.environ["LD_PRELOAD"] = preloadLib
+    fuseProcess = subprocess.Popen(fuseCMD,
```
Nikratio, line 495: Please use_underscores rather than CamelCase in Python code.

test/test_examples.py
```diff
@@ -442,6 +442,102 @@ def test_cuse(output_checker):
...
+                                   stdout=output_checker.fd,
+                                   stderr=output_checker.fd)
+    os.environ["LD_PRELOAD"] = ""
+
+    leftoverFilesCount = -1
+    try:
+        wait_for_mount(fuseProcess, fuseMNT)
+
+        # use TemporaryDirectory so that it gets cleaned up automatically
```
Nikratio, line 504: This comment just repeats what the code already says, please remove.

test/test_examples.py
```diff
@@ -442,6 +442,102 @@ def test_cuse(output_checker):
...
+        tempDir = tempfile.TemporaryDirectory(dir=fuseMNT + "/tmp/")
+        tempDirPath = tempDir.name
+        #print("tempDir: " + tempDirPath);
+
+        ## test for the race condition
+
+        # create the test file
```
Nikratio, line 511: This comment just repeats what the code already says, please remove.

test/test_examples.py
```diff
@@ -442,6 +442,102 @@ def test_cuse(output_checker):
...
+    #print("fuseCMD: ", str(fuseCMD))
+
+    # start FUSE with our preload library
+    os.environ["LD_PRELOAD"] = preloadLib
+    fuseProcess = subprocess.Popen(fuseCMD,
+                                   stdout=output_checker.fd,
+                                   stderr=output_checker.fd)
+    os.environ["LD_PRELOAD"] = ""
+
+    leftoverFilesCount = -1
```
Nikratio, line 500: Is this needed?

test/test_examples.py
```diff
@@ -442,6 +442,102 @@ def test_cuse(output_checker):
...
+        os.close(tempFile)
+        os.unlink(tempFilePath)
+
+        # check if any ".fuse_hidden" files are still there
+        tempDirFiles = os.listdir(tempDirPath)
+        #print(tempDirFiles)
+
+        leftoverFilesCount = len(tempDirFiles)
+
+        if leftoverFilesCount == 0:
```
Nikratio, line 526: You can replace all this with `assert os.listdir(tempDirPath) == []`, no need for any print output from tests

test/meson.build
devPeri: For shared libraries to work properly overloading symbols when preloaded you need specific parameters. I am not sure if Meson's standard building rules won't interfere with the (I hardly use Meson).
I'll give it a try and see if it works.

test/test746.c
devPeri: I assume you mean modifying `examples/passthrough_fh.c`. I didn't want to do that as it's in the `examples` directory rather than in `tests`, and this extra test code will probably confuse people looking at the examples when writing their own FUSE apps.

I can modify the example code instead, if you prefer.

TEST746_DELAY_DISABLE is actually used to disable the injected delays, as a way to test the test itself giving a false-PASS without the fix.  I put it there just for internal testing of the fix.

test/test_examples.py
devPeri: This is a deliberate comment so that `TemporaryDirectory()` is not replaced in the future with `mkdtemp()`, which will need the user to handle deleting the directory and any leftover contents.

test/test_examples.py
devPeri: That's just my code commenting style.
`##` is like a section header used to pinpoint where the actual testing logic starts, and the subsequent '#' are narrating what the testing sequence is doing and why it's doing it.

test/test_examples.py
devPeri: This test started as a standalone test file before merging with `test_examples.py`. I'll remove the extra prints.

Do you want the test to fail silently? I print the files so that if it fails the report will clearly show what triggered the fail.

test/meson.build
devPeri: Got it working with Meson's shared_library(). ChatGPT to the rescue. Patch coming.

test/test_examples.py
devPeri: Changes in next patch.

test/test_examples.py
devPeri: Changes in next patch.

test/test_examples.py
devPeri: Changes in next patch.

test/test_examples.py
devPeri: Removed all the excess code. Kept using a variable for the assert so that a failure will be easier to understand.
Changes in next patch.

test/meson.build
```diff
@@ -14,6 +14,14 @@ td += executable('readdir_inode', 'readdir_inode.c',
                  include_directories: include_dirs,
                  install: false)
 
+td += shared_library('test746', 'test746.c',
```
devPeri, line 17: @Nikratio Meson adds the default project flags as well, and I can't seem to figure out how to  prevent that without creating a new project and add a more complex configuration. There is an `override_options` parameter but it seems that it needs a newer Meson version to work. Any Meson expert around?

This builds a working library for now but I can't guarantee that it will not break in the future if the project flags change.

eli-schwartz: > but it seems that it needs a newer Meson version to work

Can you describe what you mean by this? The kwarg itself was added in meson 0.40.0 which is significantly older than the minimum version required by libfuse (0.51.0).

Which default project flags are you concerned about, by the way? Stuff like `-O2 -g`? Why would this cause a problem here?

devPeri: ### Meson
I saw that there is a way to override the default arguments using `override_options`.
Trying this with the `link_flags` key it gives this error:
```
WARNING: Project targets '>= 0.51' but uses feature introduced in '0.53.0': Dictionary entry using non literal key.
../test/meson.build:17:0: ERROR: Unknown variable "link_args".
```
Somewhere else I saw that the key should be as a string (e.g. `'link_args'`), but this still gives this error:
```
    if k.lang:
       ^^^^^^
AttributeError: 'str' object has no attribute 'lang'
../test/meson.build:17:0: ERROR: Unhandled python exception
    This is a Meson bug and should be reported!
```

As I mentioned I don't know much about Meson so I might be trying to do this entirely the wrong way.

### Compile flags
I add in the specific target the flags that are needed for the library's specific use, but this target configuration cannot guarantee that things won't change and break in the future. It's just a personal preference to keep such tests as tight as possible as if something changes and it breaks then the whole test could be invalidated and keep giving false positives and be very hard to notice.

The final combined compilation flags by Meson are:   
```
[1/2] cc -Itest/libtest746.so.p -Itest -I../test -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64
-Wall -Winvalid-pch -Wextra -O2 -g -D_REENTRANT -DHAVE_LIBFUSE_PRIVATE_CONFIG_
-Wno-sign-compare -Wstrict-prototypes -Wmissing-declarations -Wwrite-strings -fno-strict-aliasing
-Wno-unused-result -DHAVE_SYMVER_ATTRIBUTE -fPIC -MD -MQ test/libtest746.so.p/test746.c.o -MF
test/libtest746.so.p/test746.c.o.d -o test/libtest746.so.p/test746.c.o -c ../test/test746.c

[2/2] cc  -o test/libtest746.so test/libtest746.so.p/test746.c.o -Wl,--as-needed -Wl,--no-undefined
-shared -fPIC -Wl,--start-group -Wl,-soname,libtest746.so -Wl,-init,test746_init -Wl,-z,initfirst
-ldl -Wl,--end-group
```

eli-schwartz: What option are you trying to override? "link_args" isn't a builtin option: https://mesonbuild.com/Builtin-options.html#compiler-options

eli-schwartz: > It's just a personal preference to keep such tests as tight as possible as if something changes and it breaks then the whole test could be invalidated and keep giving false positives and be very hard to notice.

I'm not sure I buy this logic.

You could say by the same token that you prefer to keep the project runtime "as tight as possible" because if something changes and the program breaks then nothing works anymore.

test/test746.c
Nikratio: Yep, I meant `passthrough_fh.c`. The solution of compiling a separate library and pre-loading it seems unnecessary complex to me if we have full control over the program that we're using it with.

You are right that this adds unnecessary complexity though. How about forking `examples/passthrough_fh` into a separate `tests/issue_746_helper.c` that contains the extra sleep calls while getting rid of all the other code that is not needed to reproduce the problem?

test/test_examples.py
Nikratio: I have no objections to the style, it sounds quite reasonable. However, if there is only one function in one file in the project that uses the style, then it doesn't provide much value, so I'd appreciate if we could keep things consistent (even if your style might be strictly better).

test/test_examples.py
Nikratio: Try giving `assert  os.listdir(tempDirPath) == []` a try when the test is failing. I think you will find that it gives you all the information you want (pytest does a lot of magic with the assert statement).

test/meson.build
Nikratio: This discussion nicely illustrates the complexity of adding a separate LD_PRELOAD for this test. Let's try the approach with a separate `tests/issue_746_helper.c` FUSE daemon first before we continue this discussion please.

devPeri: @eli-schwartz
It's the compiling and linker flags to ensure that the generated object code, the initialisation of the preloaded library, and the symbol overloading work properly. Initially I used `custom_target` to compile this with fixed compiler settings, but I was told in a previous review to use the default Meson library build functions.

@Nikratio
Ok I'll remove all the extra stuff and add a minimal clone of passthrough_fh as a helper fuse daemon for the test.
Ten developers would approach this in ten different ways. I would prefer the PRELOAD method as that opens up the option to expand this test and easily add stress-testing tests to all passthrough examples, with various fixed and random runtime delay/fault injections without the need to create and maintain individual helper clones.

devPeri: Changes included in next patch.

test/test_examples.py
devPeri: Changes included in next patch.

test/test_examples.py
devPeri: Changes included in next patch.

test/test746.c
devPeri: Changes included in next patch.

test/issue_746_helper.c
```diff
@@ -0,0 +1,383 @@
+/*
```
matthiasgoergens, line 1: I suggest giving this file a more descriptive name perhaps, so that people don't have to refer back to github?

test/issue_746_helper.c
```diff
@@ -0,0 +1,383 @@
...
+
+	(void) path;
+	res = pwrite(fi->fh, buf, size, offset);
+	if (res == -1)
+		res = -errno;
+
+	return res;
+}
+
+static int xmp_statfs(const char *path, struct statvfs *stbuf)
```
matthiasgoergens, line 317: Do we need to implement all this functionality to make a minimal reproduction for a test here?

bsbernd: If we keep it, it could be used for other tests?

devPeri: I don't know if it is needed for the pytest mount/unmount/cleanup logic in the CI etc.

test/issue_746_helper.c
bsbernd: Keep the initial example/passthrough name? I guess it could be used for other tests as well.

test/issue_746_helper.c
matthiasgoergens: If we want to use it for other tests, I suggest folding it back into the general passthrough example, perhaps?  Or otherwise giving it a more descriptive name?

matthiasgoergens:
Thanks for writing the test!

test/issue_746_helper.c
bsbernd: Idea was to keep it in tests, to avoid polluting example/ with test specific things.

devPeri: Yes there is an earlier lengthy discussion about this. Due to the nature of the bug either you have to keep the existing example and add a PRELOAD library, or taint the example with inline test code, or create and maintain a new dedicated FUSE binary.

matthiasgoergens: Oh, I see.  Yes keeping the examples simple is indeed a good idea.

devPeri: Due to the nature of the bug and the type of test, you need at least these:

```
        .init           = xmp_init,
        .getattr        = xmp_getattr,
        .access         = xmp_access,
        .opendir        = xmp_opendir,
        .readdir        = xmp_readdir,
        .releasedir     = xmp_releasedir,
        .mkdir          = xmp_mkdir,
        .unlink         = xmp_unlink,
        .rmdir          = xmp_rmdir,
        .rename         = xmp_rename,
        .create         = xmp_create,
        .release        = xmp_release,
```

Could remove `rmdir` if you want to leave the temporary directory behind (not sure if `TemporaryDirectory` likes that).

test/test_examples.py
```diff
@@ -442,6 +442,62 @@ def test_cuse(output_checker):
     finally:
         mount_process.terminate()
 
+def test_746(tmpdir, output_checker):
+    # ###
```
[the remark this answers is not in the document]
devPeri: Changed in next patch.

test/issue_746_helper.c
devPeri: Name suggestions?

test/test_examples.py
matthiasgoergens: Btw, I usually just react with a thumbs up (and mark as resolved when appropriate) when I implement a suggestion. I feel that cuts down on noise a bit, but I'm not sure if other people have any opinions.

Not a big deal either way. Thanks for fixing this bug!

test/test_examples.py
```diff
@@ -442,6 +442,62 @@ def test_cuse(output_checker):
...
+        "-f", fuse_mountpoint]
+
+    fuse_process = subprocess.Popen(fuse_binary_command,
+                                   stdout=output_checker.fd,
+                                   stderr=output_checker.fd)
+
+    try:
+        wait_for_mount(fuse_process, fuse_mountpoint)
+
+        temp_dir = tempfile.TemporaryDirectory(dir=fuse_mountpoint + "/tmp/")
```
Nikratio, line 481: Why the need for another temporary directory? The mountpoint is already a temporary directory, so couldn't we do everything there and eliminate the need for the filesystem to implement mkdir() et al?

test/issue_746_helper.c
```diff
@@ -0,0 +1,263 @@
...
+
+	close(fi->fh);
+
+	return 0;
+}
+
+static const struct fuse_operations xmp_oper = {
+	.init           = xmp_init,
+	.getattr	= xmp_getattr,
+	.access		= xmp_access,
```
Nikratio, line 247: Shouldn't be needed with default_Permissions, I think?

test/test_examples.py
```diff
@@ -442,6 +442,62 @@ def test_cuse(output_checker):
...
         mount_process.terminate()
 
+def test_746(tmpdir, output_checker):
+    """test case for Issue #746
+
+    If RELEASE and UNLINK opcodes are sent back to back, and fuse_fs_release()
+    and fuse_fs_rename() are slow to execute, UNLINK will run while RELEASE is
+    still executing. UNLINK will try to rename the file and, while the rename
+    is happening, the RELEASE will finish executing. As a result, RELEASE will
+    not detect in time that UNLINK has happened, and UNLINK will not detect in
```
Nikratio, line 452: To test this, do we need a passthrough filesystem at all? Can't we have something super simple (along the lines of hello.c) that tests *in the filesystem* that things are happening in the right order?

test/test_examples.py
devPeri: We need a clean empty directory to test for leftovers. Unless I'm mistaken, being a passthrough the temp dir is the "/" of the system, so we need an guaranteed empty subdir with write permissions to create & delete the temp file and then check that the dir is empty. The test uses `/tmp/[random_dir]`.
Is there any pre-existing dir in the CI system that is guaranteed to be empty and have write perms?

test/issue_746_helper.c
devPeri: Not sure. I included all  FUSE OPs that are used during the test to be compliant to the expected filesystem behaviour.

test/test_examples.py
devPeri: I guess it might be able to work. I assume it will need a bit of an elaborate test to emulate the creation, renaming, and deletion of multiple files, listdirs, etc, and be `nullpath_ok` compliant. This itself will need validation as well, and future API maintenance.

This test started with the LD_PRELOAD as a proper end-to-end black-box testing of the intermediate layers through the use of the examples that should always represent the correct and expected design of a FUSE application, and with the potential to expand to other examples and to other type of fault/latency-injection and dynamic testing.
Creating a fake FS implementation with a very limited functionality and pre-coded behaviour invalidates the whole point of black-box testing.
In this case the whole test could be scraped and might be able to be replaced with an atomic flag and an assert within `fuse.c` to check that `release()` and `unlink()` do not run simultaneously.

Anyone offering to write and validate the custom FS implementation for this test?

test/issue_746_helper.c
devPeri: It seems to still work without `access` and `releasedir`, and without `rmdir` if the temp dir is not cleaned up.

Do you want a patch with these three removed?

test/test_examples.py
Nikratio: > We need a clean empty directory to test for leftovers. Unless I'm mistaken, being a passthrough the temp dir is the "/" of the system,

It is for `examples/passthrough.c` but that doesn't need to be the case for the helper. You could e.g. just use the current working directory, and change into `tmpdir` prior to running the helper.

test/issue_746_helper.c
Nikratio: Maybe `release_unlink_race`?

test/issue_746_helper.c
Nikratio: I would prefer as little code as possible in the helper, but it's your call.

I think we can also get rid of mkdir() and opendir() if you run the test in the mountpoint (i.e., no support for sub-directories), and you could simplify `listdir` a lot by not supporting offsets. We could also hardcode nullpath_ok to get rid of some more conditions in the other callbacks.

devPeri:
@Nikratio did you get the chance to review the latest commits?

Nikratio:
No, I didn't know they were waiting for review. :-)

Will take a look

Nikratio:
Thanks for all your efforts here!

It seems that the test is failing. But this may be correct, because if I'm not mistaken this pull request now only contains the test, without the fix. Did we loose the fix somewhere along the way?

devPeri:
Yes, I reverted the patch at some point as I was testing the various test combinations.
It's now reapplied.

Nikratio:
Hmm. Looks like the patch broke another unit test?

devPeri:
I have been testing this but can't get this error.
Compiled with `CC=clang CXX=clang++ meson setup -D b_sanitize=address -D b_lundef=false -D werror=true` the test does not fail.
Versions:
```
C++ compiler for the host machine: clang++ (clang 14.0.6 "Debian clang version 14.0.6")
C++ linker for the host machine: clang++ ld.bfd 2.40
Build targets in project: 29

libfuse3 3.14.1

  User defined options
    werror    : true
    b_lundef  : false
    b_sanitize: address

Found ninja-1.11.1 at /usr/bin/ninja
```

Is there any other special build/run setting in the CI setup, that I missed?

Nikratio:
Hmm. Maybe it as a flaky test. Let me re-run them...

Nikratio:
Failing again, but this time in a different test...

devPeri:
The test itself had a race condition triggered by the Github's slow workflow system. The unlink sometimes wouldn't complete the cleanup before the assert was checking for the leftover files.
Added a delay and tested it by running the test workflow a bunch of times. That other test's "got_write" assert fails randomly as well.

Nikratio:
Thanks for all your work on this!

PR #784, opened 2023-04-18
Use `posix_spawn` instead of `fork`+`exec`

matthiasgoergens:
Client code might allocate a lot of memory before starting the mount.

Forking is slow for processes that are using a lot of memory.  `posix_spawn` fixes that.

### DIFF
--- a/lib/mount.c
+++ b/lib/mount.c
@@ -9,7 +9,9 @@
 */
 
 /* For environ */
+#ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
 
 #include "fuse_config.h"
 #include "fuse_i.h"
--- a/lib/mount.c
+++ b/lib/mount.c
@@ -322,6 +322,65 @@ void fuse_kern_unmount(const char *mountpoint, int fd)
 	waitpid(pid, NULL, 0);
 }
 
+static int setup_auto_unmount(const char *mountpoint, int quiet)
+{
+	int fds[2], pid;
+	int res;
+
+	if (!mountpoint) {
+		fuse_log(FUSE_LOG_ERR, "fuse: missing mountpoint parameter\n");
+		return -1;
+	}
+
+	res = socketpair(PF_UNIX, SOCK_STREAM, 0, fds);
+	if(res == -1) {
+		perror("fuse: socketpair() failed");
+		return -1;
+	}
+
+	pid = fork();
+	if(pid == -1) {
+		perror("fuse: fork() failed");
+		close(fds[0]);
+		close(fds[1]);
+		return -1;
+	}
+
+	if(pid == 0) {
+		char env[10];
+		const char *argv[32];
+		int a = 0;
+
+		if (quiet) {
+			int fd = open("/dev/null", O_RDONLY);
+			if (fd != -1) {
+				dup2(fd, 1);
+				dup2(fd, 2);
+			}
+		}
+
+		argv[a++] = FUSERMOUNT_PROG;
+		argv[a++] = "--auto-unmount";
+		argv[a++] = "--";
+		argv[a++] = mountpoint;
+		argv[a++] = NULL;
+
+		close(fds[1]);
+		fcntl(fds[0], F_SETFD, 0);
+		snprintf(env, sizeof(env), "%i", fds[0]);
+		setenv(FUSE_COMMFD_ENV, env, 1);
+		exec_fusermount(argv);
+		perror("fuse: failed to exec fusermount3");
+		_exit(1);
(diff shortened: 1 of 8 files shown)


Nikratio:
I'd be very curious to see hard numbers on the effects here. I wouldn't be surprised if posix_spawn is implemented entirely in glibc (no special syscall) as fork+exec, so that on Linux it's not faster or resource saving at all.

matthiasgoergens:
It's faster on Linux with glibc.

They don't use a plain fork to implement posix_spawn, but something that avoids having to set up the metadata for copy-on-write memory.

I ran some benchmarks by hacking up the hello_ll example: I malloc-ed about a dozen GiB and used memset to fill them with 1s. Then I timed how long the call to mount for an unprivileged user takes (the mount that falls back on fusermount).

With posix_spawn the duration always stays below about 1 millisecond. With fork it creeps up to about 100 milliseconds. The measurement is a bit noisy and the code was a bit hacky.

The delay for fork was about linearly related to the amount of memory allocated.

bsbernd:
I actually had this on my TODO list. Fork has an issue with registered memory, like for IB RDMA and NVME RDMA and can cause severe data corruption. We sadly had such a case with a customer (caused by libfabric silenty ignoring a flag from one version to another). We have already added some counter measures for libfuse into our code... Also see here: https://blog.nelhage.com/post/a-cursed-bug/

With recent kernels the fork/rdma issue a better, but I think it is still better to avoid fork whenever possible.

Nikratio:
That's good to know, thanks!

Do you think the performance is going to matter for libfuse? I believe the fork's are all one-off, and we don't allocate GB of memory. Or is the posix_nspawn code shorter / easier to understand?

matthiasgoergens:
We (potentially) fork for every mount and unmount.

Yes, we don't allocate much memory, but client code might. Eg they might do lots of set up before calling mount. Or at the end of the program when unmount is called, they might have allocated a lot of memory from running a long time.

I left the forks in fusermount, because that's not a library and we know for sure that we only use a small amount of memory.

The posix_spawn code seems about equally complex as the fork+exec code. On the negative side, it's a move away from the status quo which is always something that needs justifying. On the plus side, it's faster.

(Come to think of it, I think that the fork+exec code is perhaps slightly more complicated, as you need to worry about a lot of things client code might have done, like threads etc.)

util/fusermount.c
```diff
@@ -1468,6 +1468,7 @@ int main(int argc, char *argv[])
 
 	if (!auto_unmount) {
 		free(mnt);
+		free((void*) type);
```
matthiasgoergens, line 1471: This is only here to make address sanitizer happy.  I've moved it into its own pull request at https://github.com/libfuse/libfuse/pull/785 so we can remove it here once it's landed in master.

matthiasgoergens:
Hmm, for some reason my latest changes don't show up on the pull request here.  I hope they show up eventually.

@bsbernd I removed the relatively complicated environment manipulation, and replaced it with a new (internal-only) argument to fusermount.

lib/mount.c
```diff
@@ -44,14 +45,12 @@
...
 
-#ifndef HAVE_FORK
-#define fork() vfork()
-#endif
-
 #ifndef MS_DIRSYNC
 #define MS_DIRSYNC 128
 #endif
 
+extern char **environ;
```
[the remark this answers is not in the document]
matthiasgoergens: No hurries.

lib/mount.c
```diff
@@ -8,6 +8,11 @@
   See the file COPYING.LIB.
 */
 
+/* For environ */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
```
matthiasgoergens, line 13: Hmm, not sure this actually worked.

Locally my tests pass, but since I defined `_GNU_SOURCE` here, I've been getting test failures on github.

Nikratio:
@matthiasgoergens What's the status of this?

matthiasgoergens:
Sorry, I need to look into this again to refresh my memory.

Nikratio:
Any update?

Nikratio:
Will close this for now, please feel free to re-open when you have time to work on this.

matthiasgoergens:
Sorry, I didn't get around to doing this.  Thanks for taking over!

bsbernd:
I'm in the middle to update it - will squash your patches into one and apply my patch on top of it.

bsbernd:
@matthiasgoergens I opened a new pull request: https://github.com/libfuse/libfuse/pull/901
I'm closing this one here.

PR #785, opened 2023-04-18
Fix memory leak

matthiasgoergens:
This is just to [keep address sanitizer happy](https://github.com/libfuse/libfuse/actions/runs/4730520764/jobs/8394347666?pr=784).  The OS would normally clean this up anyway.

Extracted from https://github.com/libfuse/libfuse/pull/784


matthiasgoergens:
Hmm, I can reproduce https://github.com/libfuse/libfuse/actions/runs/4730849094/jobs/8395069433?pr=785 locally.  Still trying.

matthiasgoergens:
Hmm, the earlier problem went away after a force push that only introduced a different commit message.  Must be a flaky test that probably has nothig to do with my changes.. hmm.

PR #787, opened 2023-04-19
Remove unnecessary `_GNU_SOURCE` in `fuse.c`

matthiasgoergens:
We stopped using pthread_rwlock_t in 3fecccca989328ed2c0ac68860ee1ceec0673972, so we don't need `_GNU_SOURCE` anymore in `fuse.c`


matthiasgoergens:
If we need _GNU_SOURCE for other reasons, we should fix the comment.

Issue #707, opened 2022-10-27
Support fuse_lowlevel_notify_expire_entry()


HereThereBeDragons:
This new function allows dentries to be marked as only expired, instead of always be invalidated as suggested in this [conversation](https://sourceforge.net/p/fuse/mailman/fuse-devel/thread/CAJfpegs8VKu45Jsv%2B-qAWwz2A%2BYpxCtxrFJr%2BA-b1g%2BzKsf%3DOA%40mail.gmail.com/#msg37703271). 

The (final?) kernel patch provided Miklos Szeredi is [here](https://git.kernel.org/pub/scm/linux/kernel/git/mszeredi/fuse.git/commit/?h=for-next&id=53e949edb7692dce02220eba926c9d75ecbb47f7)

We tested the first unofficial patch provided by Miklos Szeredi  in the email conversation and it worked.

HereThereBeDragons:
Just as a note: the kernel patch is planned to be part of kernel version 6.2

PR #44, opened 2016-05-11
lib/fuse_lowlevel.c: fix small typo in fuse_ll_help

dsheets:
s/disable remove file locking/disable remote file locking/

PR #789, opened 2023-05-04
Make expire only function fail if no kernel support

HereThereBeDragons:
Hi @Nikratio 

I finally managed to take care of this PR concering expire_only ( Issue #707 ) 

Integrates fix from PR #771  for the example. However changes a biy by removing `tfs_init` (just commented out) (currently not a proper cherry pick of that branch.. hope that is ok?)

I renamned the old function that does not check for the kenrel support of expire and just falls back to invalidate to ` fuse_lowlevel_notify_entry`. It us now designed to be a base function used both by `expire` and `invalidate`. The new `expire` then does the checking for the kernel support.

For the example i added a `raise(SIGINT)` in case the kernel support is not given. ( i am not sure how backporting functionality works but i now there is red hat ticket open to backport the kernel. as such i thought this example better fails because the function return `-ENOSYS` then check before the major/minor fuse version.

Open to discussion if that is how you want it 
Maybe @bsbernd  and @matthiasgoergens are also interested in the discussion

### DIFF
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -78,6 +78,8 @@
 
 #define FUSE_USE_VERSION 34
 
+#include <fuse_i.h>
+#include <fuse_kernel.h>
 #include <fuse_lowlevel.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -230,15 +232,7 @@ static void tfs_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
     }
 }
 
-static void tfs_init(void *userdata, struct fuse_conn_info *conn) {
-    if(options.only_expire && !(conn->capable & FUSE_CAP_EXPIRE_ONLY)) {
-        fprintf(stderr, "FUSE_CAP_EXPIRE_ONLY not supported by kernel\n");
-        fuse_session_exit(*(struct fuse_session **) userdata);
-    }
-}
-
 static const struct fuse_lowlevel_ops tfs_oper = {
-    //.init	= tfs_init,
     .lookup	= tfs_lookup,
     .getattr	= tfs_getattr,
     .readdir	= tfs_readdir,
@@ -263,28 +257,40 @@ static void* update_fs_loop(void *data) {
     struct fuse_session *se = (struct fuse_session*) data;
     char *old_name;
 
-    while(1) {
+
+    while(!fuse_session_exited(se)) {
         old_name = strdup(file_name);
         update_fs();
-        if (!options.no_notify && lookup_cnt) {
-            if(options.only_expire) {
+
+        int curr_lookup_cnt = lookup_cnt;
+        if (!options.no_notify && curr_lookup_cnt) {
+            if(options.only_expire) { // expire entry
                 int ret = fuse_lowlevel_notify_expire_entry
                    (se, FUSE_ROOT_ID, old_name, strlen(old_name));
+
+                // no kernel support
                 if (ret == -ENOSYS) {
                     printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
-                    printf("Next call to mountpoint will kill it...\n");
-                    raise(SIGINT);
-                    return NULL;
+                    printf("Exiting...\n");
+                    se->error = ret;
+
+                    fuse_session_exit(se);
+                    while(!fuse_session_exited(se)) {
+                        sleep(1);
+                    }
+                    fuse_session_unmount(se);
+
+                    break;
                 }
-                // forget is not being called
-                // the dentry is running just into a timeout
-                if (old_lookup_cnt == lookup_cnt) {
+                // kernel cache timeout deletes dentry without calling forget()
+                // lookup_cnt is therefore not decreased
+                if (old_lookup_cnt == curr_lookup_cnt) {
                     assert(ret == -ENOENT);
-                } else {
+                } else { // expire was successful
                     assert(ret == 0);
-                    old_lookup_cnt = lookup_cnt;
+                    old_lookup_cnt = curr_lookup_cnt;
                 }
-            } else {
+            } else { // invalidate entry
                 assert(fuse_lowlevel_notify_inval_entry
                       (se, FUSE_ROOT_ID, old_name, strlen(old_name)) == 0);
             }
@@ -357,12 +363,19 @@ int main(int argc, char *argv[]) {
     }
 
     /* Block until ctrl+c or fusermount -u */
-    if (opts.singlethread)
+    if (opts.singlethread) {
         ret = fuse_session_loop(se);
-    else {
+    } else {
         config.clone_fd = opts.clone_fd;
         config.max_idle_threads = opts.max_idle_threads;
         ret = fuse_session_loop_mt(se, &config);
+    }
+
+    pthread_join(updater, NULL);
+
+    // fuse session was cancelled within update_fs_loop()
+    if (ret == -FUSE_DESTROY) {
+        goto err_out3;
     }
 
     fuse_session_unmount(se);
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -371,8 +371,6 @@ int main(int argc, char *argv[]) {
         ret = fuse_session_loop_mt(se, &config);
     }
 
-    pthread_join(updater, NULL);
-
     // fuse session was cancelled within update_fs_loop()
     if (ret == -FUSE_DESTROY) {
         goto err_out3;
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -270,20 +270,20 @@ static void* update_fs_loop(void *data) {
             if(options.only_expire) {
                 int ret = fuse_lowlevel_notify_expire_entry
                    (se, FUSE_ROOT_ID, old_name, strlen(old_name));
-               // forget is not being called
-               // the dentry is running just into a timeout
-	       if (ret == -ENOSYS) {
-		 printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
-		 printf("Stopping mountpoint...\n");
-		 raise(SIGINT);
-		 return NULL;
-	       }
-               if (old_lookup_cnt == lookup_cnt) {
-                 assert(ret == -ENOENT);
-               } else {
-                 assert(ret == 0);
-                 old_lookup_cnt = lookup_cnt;
-               }
+                if (ret == -ENOSYS) {
+                    printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
+                    printf("Next call to mountpoint will kill it...\n");
+                    raise(SIGINT);
+                    return NULL;
+                }
+                // forget is not being called
+                // the dentry is running just into a timeout
+                if (old_lookup_cnt == lookup_cnt) {
+                    assert(ret == -ENOENT);
+                } else {
+                    assert(ret == 0);
+                    old_lookup_cnt = lookup_cnt;
+                }
             } else {
                 assert(fuse_lowlevel_notify_inval_entry
                       (se, FUSE_ROOT_ID, old_name, strlen(old_name)) == 0);
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -78,7 +78,6 @@
 
 #define FUSE_USE_VERSION 34
 
-#include <fuse_kernel.h>
 #include <fuse_lowlevel.h>
 #include <stdio.h>
 #include <stdlib.h>
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -78,7 +78,6 @@
 
 #define FUSE_USE_VERSION 34
 
-#include <fuse_i.h>
 #include <fuse_kernel.h>
 #include <fuse_lowlevel.h>
 #include <stdio.h>
@@ -271,10 +270,9 @@ static void* update_fs_loop(void *data) {
                 if (ret == -ENOSYS) {
                     printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
                     printf("Exiting...\n");
-                    se->error = ret;
 
                     fuse_session_exit(se);
-                    // needed to correctly exit fuse_session_loop()
+                    // Make sure to exit now, rather than on next request from userspace
                     pthread_kill(main_thread, SIGPIPE);
 
                     break;
@@ -347,7 +345,9 @@ int main(int argc, char *argv[]) {
 
     fuse_daemonize(opts.foreground);
 
-    // needed to correctly exit fuse_session_loop()
+    // Needed to ensure that the main thread continues/restarts processing as soon
+    // as the fuse session ends (immediately after calling fuse_session_exit() ) 
+    // and not only on the next request from userspace
     main_thread = pthread_self();
 
     /* Start thread to update file contents */
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -96,7 +96,7 @@
 static char file_name[MAX_STR_LEN];
 static fuse_ino_t file_ino = 2;
 static int lookup_cnt = 0;
-static int old_lookup_cnt = 0;
+static pthread_t main_thread;
 
 /* Command line parsing */
 struct options {
@@ -262,8 +262,7 @@ static void* update_fs_loop(void *data) {
         old_name = strdup(file_name);
         update_fs();
 
-        int curr_lookup_cnt = lookup_cnt;
-        if (!options.no_notify && curr_lookup_cnt) {
+        if (!options.no_notify && lookup_cnt) {
             if(options.only_expire) { // expire entry
                 int ret = fuse_lowlevel_notify_expire_entry
                    (se, FUSE_ROOT_ID, old_name, strlen(old_name));
@@ -275,21 +274,15 @@ static void* update_fs_loop(void *data) {
                     se->error = ret;
 
                     fuse_session_exit(se);
-                    while(!fuse_session_exited(se)) {
-                        sleep(1);
-                    }
-                    fuse_session_unmount(se);
+                    // needed to correctly exit fuse_session_loop()
+                    pthread_kill(main_thread, SIGPIPE);
 
                     break;
                 }
-                // kernel cache timeout deletes dentry without calling forget()
-                // lookup_cnt is therefore not decreased
-                if (old_lookup_cnt == curr_lookup_cnt) {
-                    assert(ret == -ENOENT);
-                } else { // expire was successful
-                    assert(ret == 0);
-                    old_lookup_cnt = curr_lookup_cnt;
-                }
+                // 1) ret == 0: successful expire of an existing entry
+                // 2) ret == -ENOENT: kernel has already expired the entry /
+                //                    entry does not exist anymore in the kernel
+                assert(ret == 0 || ret == -ENOENT);
             } else { // invalidate entry
                 assert(fuse_lowlevel_notify_inval_entry
                       (se, FUSE_ROOT_ID, old_name, strlen(old_name)) == 0);
@@ -354,6 +347,9 @@ int main(int argc, char *argv[]) {
 
     fuse_daemonize(opts.foreground);
 
+    // needed to correctly exit fuse_session_loop()
+    main_thread = pthread_self();
+
     /* Start thread to update file contents */
     ret = pthread_create(&updater, NULL, update_fs_loop, (void *)se);
     if (ret != 0) {
@@ -369,11 +365,6 @@ int main(int argc, char *argv[]) {
         config.clone_fd = opts.clone_fd;
         config.max_idle_threads = opts.max_idle_threads;
         ret = fuse_session_loop_mt(se, &config);
-    }
-
-    // fuse session was cancelled within update_fs_loop()
-    if (ret == -FUSE_DESTROY) {
-        goto err_out3;
     }
 
     fuse_session_unmount(se);
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -85,6 +85,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <assert.h>
+#include <signal.h>
 #include <stddef.h>
 #include <unistd.h>
 #include <pthread.h>
@@ -93,6 +94,7 @@
 static char file_name[MAX_STR_LEN];
 static fuse_ino_t file_ino = 2;
 static int lookup_cnt = 0;
+static int old_lookup_cnt = 0;
 
 /* Command line parsing */
 struct options {
@@ -228,7 +230,15 @@ static void tfs_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
     }
 }
 
+static void tfs_init(void *userdata, struct fuse_conn_info *conn) {
+    if(options.only_expire && !(conn->capable & FUSE_CAP_EXPIRE_ONLY)) {
+        fprintf(stderr, "FUSE_CAP_EXPIRE_ONLY not supported by kernel\n");
+        fuse_session_exit(*(struct fuse_session **) userdata);
+    }
+}
+
 static const struct fuse_lowlevel_ops tfs_oper = {
+    //.init	= tfs_init,
     .lookup	= tfs_lookup,
     .getattr	= tfs_getattr,
     .readdir	= tfs_readdir,
@@ -258,8 +268,22 @@ static void* update_fs_loop(void *data) {
         update_fs();
         if (!options.no_notify && lookup_cnt) {
             if(options.only_expire) {
-                assert(fuse_lowlevel_notify_expire_entry
-                   (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY) == 0);
+                int ret = fuse_lowlevel_notify_expire_entry
+                   (se, FUSE_ROOT_ID, old_name, strlen(old_name));
+               // forget is not being called
+               // the dentry is running just into a timeout
+	       if (ret == -ENOSYS) {
+		 printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
+		 printf("Stopping mountpoint...\n");
+		 raise(SIGINT);
+		 return NULL;
+	       }
+               if (old_lookup_cnt == lookup_cnt) {
+                 assert(ret == -ENOENT);
+               } else {
+                 assert(ret == 0);
+                 old_lookup_cnt = lookup_cnt;
+               }
             } else {
                 assert(fuse_lowlevel_notify_inval_entry
                       (se, FUSE_ROOT_ID, old_name, strlen(old_name)) == 0);
@@ -312,7 +336,7 @@ int main(int argc, char *argv[]) {
     update_fs();
 
     se = fuse_session_new(&args, &tfs_oper,
-                          sizeof(tfs_oper), NULL);
+                          sizeof(tfs_oper), &se);
     if (se == NULL)
         goto err_out1;
 
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -1683,30 +1683,7 @@ int fuse_lowlevel_notify_inval_inode(struct fuse_session *se, fuse_ino_t ino,
 				     off_t off, off_t len);
 
 /**
- * Notify parent attributes and the dentry matching parent/name
- * 
- * Underlying function for fuse_lowlevel_notify_inval_entry() and
- * fuse_lowlevel_notify_expire_entry().
- * 
- * @warning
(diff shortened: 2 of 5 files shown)


example/notify_inval_entry.c
```diff
@@ -228,7 +230,15 @@ static void tfs_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
...
 
+static void tfs_init(void *userdata, struct fuse_conn_info *conn) {
+    if(options.only_expire && !(conn->capable & FUSE_CAP_EXPIRE_ONLY)) {
+        fprintf(stderr, "FUSE_CAP_EXPIRE_ONLY not supported by kernel\n");
+        fuse_session_exit(*(struct fuse_session **) userdata);
+    }
+}
+
 static const struct fuse_lowlevel_ops tfs_oper = {
+    //.init	= tfs_init,
```
Nikratio, line 241: I'm a bit confused. Why are we adding this function only to comment it out? Let's either not add it in the first place, or add it and use it.

(I think for the purposes of the example, it does not matter much if we are failing at init() time or when attempting to call the function)

example/notify_inval_entry.c
```diff
@@ -258,8 +268,22 @@ static void* update_fs_loop(void *data) {
...
         if (!options.no_notify && lookup_cnt) {
             if(options.only_expire) {
-                assert(fuse_lowlevel_notify_expire_entry
-                   (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY) == 0);
+                int ret = fuse_lowlevel_notify_expire_entry
+                   (se, FUSE_ROOT_ID, old_name, strlen(old_name));
+                if (ret == -ENOSYS) {
+                    printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
+                    printf("Next call to mountpoint will kill it...\n");
+                    raise(SIGINT);
```
Nikratio, line 276: That seems somewhat over complicated to me. Let's either print an error message and continue, or print an error and terminate the session right away (with fuse_session_exit()).

example/notify_inval_entry.c
```diff
@@ -258,8 +268,22 @@ static void* update_fs_loop(void *data) {
...
+                    return NULL;
+                }
+                // forget is not being called
+                // the dentry is running just into a timeout
+                if (old_lookup_cnt == lookup_cnt) {
+                    assert(ret == -ENOENT);
+                } else {
+                    assert(ret == 0);
+                    old_lookup_cnt = lookup_cnt;
+                }
```
Nikratio, line 286: What's the purpose of this code? It looks like we are mixing a unit test into an example, but I also do not understand what exactly is being tested here. It seems we are assuming that we know exactly what file system clients will do in-between iterations of this loop?

My feeling is that we should remove this entirely. It's not libfuses job to test if kernel functionality is implemented  correctly, especially if we can't even do so reliable. Let's just make sure that we test the userspace codepath.

lib/fuse_lowlevel.c
```diff
@@ -2333,9 +2333,21 @@ int fuse_lowlevel_notify_expire_entry(struct fuse_session *se, fuse_ino_t parent
...
+}
+
+int fuse_lowlevel_notify_expire_entry(struct fuse_session *se, fuse_ino_t parent,
+							const char *name, size_t namelen)
 {
-	return fuse_lowlevel_notify_expire_entry(se, parent, name, namelen, 0);
+	if (!se)
+		return -EINVAL;
+
+	if (!(se->conn.capable & FUSE_CAP_EXPIRE_ONLY))
```
Nikratio, line 2347: Do we need to do this check? What happens if we just pass the flag to the kernel?

If the kernel returns an error (which I really hope is the case) then we could keep this code simpler.

Nikratio (changes requested):
Thanks for working on this!

The docstrings for the involved functions read somewhat convoluted to me. Do I understand correctly that this is because we need to work around the kernel bug of simply ignoring unknown flags (instead of returning an error)?

Related to that,  fuse_lowlevel_notify_expire_entry() seems to have the same problem: it looks for known flags (in this case just FUSE_LL_EXPIRE_ONLY) and silently ignores everything else. Can we add proper checking there please?

example/notify_inval_entry.c
HereThereBeDragons: the `tfs_init()` was added by `@matthiasgoergens` during his fix of this example. So I thought I better let it in here until we decide where to check for the support. Personally I would prefer to check inside the loop to see that the function really fails

But I am open also to just adding in `tfs_init`

example/notify_inval_entry.c
HereThereBeDragons: sorry my mistake.. for whatever reason i overlooked that the `fuse_session` is given as parameter in this function

example/notify_inval_entry.c
HereThereBeDragons: we talked about it before in the other PR ( https://github.com/libfuse/libfuse/pull/771#issuecomment-1502966224 and https://github.com/libfuse/libfuse/pull/771#issuecomment-1503191654 )

this examples creates a new file with the given timestamp each second. as `lookup_cnt` is not decreased if `expire` is used, the loop is always executed. so one needs to check if there is actually a file to be expired or expired is executed on a file that does not exist.

lib/fuse_lowlevel.c
HereThereBeDragons: as far as i know the kernel reports no error for backward compatibility - that's why it just fallsback to `invalidate`

HereThereBeDragons (commented):
@Nikratio  sorry i did not get any notification that you replied to it and  i had been travelling

> Do I understand correctly that this is because we need to work around the kernel bug of simply ignoring unknown flags (instead of returning an error)?

Yes this is correct. Unknown flags are ignored and `invalidate` is the fallback

> Can we add proper checking there please?

Yes, sorry, overlooked.
I will also try to better clarify the documentation

example/notify_inval_entry.c
HereThereBeDragons: ok for whatever reason the comments were deleted...

`tfs_init` was the new function `@matthiasgoergens` provided to fix this example. i would rather prefer to check not in the init but in the function itself to see that the correct return type is given. 

but i'm open to whatever you prefer!

lib/fuse_lowlevel.c
HereThereBeDragons: the kernel does not return any error. it just falls back to `invalidate`. thats why you need a check for it

example/notify_inval_entry.c
HereThereBeDragons: we talked about this in PR #771 ( https://github.com/libfuse/libfuse/pull/771#issuecomment-1502966224 and https://github.com/libfuse/libfuse/pull/771#issuecomment-1503191654 )

a file named with the current timestamp is created every 1 sec. As `expire` does not reduce the `lookup_cnt` the if inside the loop is always executed. so one needs to destinguish if `expire` is successfully executed on a file that exists or on something that does not exist anymore.

Nikratio: Sorry, I still do not understand what's being tested here.

Are you trying to test that the *kernel* correctly processes expire requests? If so, then we should check that a client stat() after expire() results in the right callback to FUSE, and that an invalidate request results in a callback to forget(). The current code seems to do neither. However, I would argue that it's not libfuse's job to test if the kernel implementation is correct anyway, especially if we can't do it reliably (tests may fail if there are other requests to the mountpoint). So I would drop this completely (unless I'm missing something crucial?).

lib/fuse_lowlevel.c
```diff
@@ -2305,9 +2305,9 @@ int fuse_lowlevel_notify_inval_inode(struct fuse_session *se, fuse_ino_t ino,
 	return send_notify_iov(se, FUSE_NOTIFY_INVAL_INODE, iov, 2);
 }
 
-int fuse_lowlevel_notify_expire_entry(struct fuse_session *se, fuse_ino_t parent,
-				      const char *name, size_t namelen,
-				      enum fuse_expire_flags flags)
+int fuse_lowlevel_notify_entry(struct fuse_session *se, fuse_ino_t parent,
```
Nikratio, line 2308: I believe this function should check that `flags` is fully understood (i.e., it should catch when we pass 42 to it instead of FUSE_LL_INVALIDATE).

lib/fuse_lowlevel.c
```diff
@@ -2333,9 +2333,21 @@ int fuse_lowlevel_notify_expire_entry(struct fuse_session *se, fuse_ino_t parent
 }
 
 int fuse_lowlevel_notify_inval_entry(struct fuse_session *se, fuse_ino_t parent,
-				     const char *name, size_t namelen)
+						 const char *name, size_t namelen)
+{
+	return fuse_lowlevel_notify_entry(se, parent, name, namelen, FUSE_LL_INVALIDATE);
```
Nikratio, line 2338: As far as I can tell, fuse_lowlevel_notify_entry silently ignores this flag value.

example/notify_inval_entry.c
HereThereBeDragons: no we do not check if the expire request is handled correctly, only that it returns the correct return value.

because of
https://github.com/libfuse/libfuse/blob/8b77a82d3fc957a4662f9deb0d6b569428a955f1/example/notify_inval_entry.c#L269
we need to check the `expire` is executed because: 
1) "yes we accessed the file and want to expire it "(returns 0) 
or
2) "oh it's still an old one and does not exist anymore" (returns  -ENOENT)

if we do not want the `tfs_init` we also want to exit the program if no kernel support is available and `-ENOSYS` is returned

lib/fuse_lowlevel.c
HereThereBeDragons: this one is the "old" `fuse_lowlevel_notfiy_expire_entry` that just fallsback to `invalidate`.

My idea was to have  this as a (more performant) base function, only checking if `invalidate` is supported - and all other flags (existing or not) do an automatic fallback to `invalidate` if not supported.  If users want the proper check they should use the specific function named after the command they want to execute (e.g. `fuse_lowlevel_notfiy_expire_entry` to `expire` the entry)

example/notify_inval_entry.c
matthiasgoergens: If memory serves right, I added an init function to fix a memory leak or so.

example/notify_inval_entry.c
```diff
@@ -258,8 +268,22 @@ static void* update_fs_loop(void *data) {
...
-                   (se, FUSE_ROOT_ID, old_name, strlen(old_name), FUSE_LL_EXPIRE_ONLY) == 0);
+                int ret = fuse_lowlevel_notify_expire_entry
+                   (se, FUSE_ROOT_ID, old_name, strlen(old_name));
+               // forget is not being called
+               // the dentry is running just into a timeout
+	       if (ret == -ENOSYS) {
+		 printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
+		 printf("Stopping mountpoint...\n");
+		 raise(SIGINT);
+		 return NULL;
```
bsbernd: This leaks old_name - break the look. Is the raise needed? Call fuse_session_exit()?

Is raise() needed? Set se->error() before fuse_session_exit()?

bsbernd (changes requested):

example/notify_inval_entry.c
bsbernd: As you point out yourself, fuse_session_exit() can be used. I'm personally not too happy about thee lookup_cnt checks - this is fragile and not under libfuse control. Kernel side is free to release cached inodes whenever it wants - will happen under memory pressure and on echo 3 >/proc/sys/vm/drop_caches. I.e. the unit test might spuriously fail.

example/notify_inval_entry.c
HereThereBeDragons: your `tfs_init` only checks if `expire` is supported. otherwise it aborts the program.

example/notify_inval_entry.c
HereThereBeDragons: hi @bsbernd 
thanks for your feedback. you are correct with your arguments but i also think for this example those edge cases are a bit of an overkill. this example only creates/modifies a single (empty) file in a given update interval (default 1 sec) + you can set the timeout of the kernel caches (default 5 sec).

i guess if you really want to protect against drop_caches you need to change this line
https://github.com/libfuse/libfuse/blob/8b77a82d3fc957a4662f9deb0d6b569428a955f1/example/notify_inval_entry.c#L281-L283

to something like
```c++
if (old_lookup_cnt >= lookup_cnt) {
  old_lookup_cnt = lookup_cnt;
  assert(ret == -ENOENT);
} 
```

include/fuse_lowlevel.h
```diff
@@ -1675,6 +1676,24 @@ int fuse_lowlevel_notify_poll(struct fuse_pollhandle *ph);
...
+ * 
+ *
+ * @param se the session object
+ * @param parent inode number
+ * @param name file name
+ * @param namelen strlen() of file name
+ * @param flags flags to control if the entry should be expired or invalidated
+ * @return zero for success, -errno for failure
+*/
+int fuse_lowlevel_notify_entry(struct fuse_session *se, fuse_ino_t parent,
```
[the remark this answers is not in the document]
HereThereBeDragons: i personally would want to make it available to users to use it if they want a little bit more performance and dont care about that the fallback behavior is `invalidate`. 
but lets see what nikolaus says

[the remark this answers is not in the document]
Nikratio: Sorry, I missed this question. 

I'd say make it static for now:
- I don't think that the fallback behavior is very useful
- We can always make it public later on (but not the other way)
- It reduces the API surface (less complexity)
- I can't imagine where performance of this function could become a bottleneck

HereThereBeDragons:
I implemented the change requests as good as possible. 
For the `lookup_counter` i added some `current_counter` so that it is only accessed once within the loop.
Ending the thread was not so easy. Just `fuse_session_exit()` is not enough, you also need `fuse_session_unmount()` and if not done carefully you run into `double frees` when having the multithreaded fuse session loop running

bsbernd:
Going to look at it tomorrow.

Nikratio:
>. Ending the thread was not so easy. Just `fuse_session_exit()` is not enough, you also need `fuse_session_unmount()` and if not done carefully you run into `double frees` when having the multithreaded fuse session loop running

Can you elaborate on that? Calling `fuse_session_exit` should result in all worker threads terminating and (after that) the main thread returning from `fuse_session`. At that point, the file system should clean-up just like on a regular unmount. If that's not the case, then I think you have found a (separate) bug that we should address (which we can also do separately, but let's document it).

example/notify_inval_entry.c
```diff
@@ -253,14 +257,40 @@ static void* update_fs_loop(void *data) {
...
+                   (se, FUSE_ROOT_ID, old_name, strlen(old_name));
+
+                // no kernel support
+                if (ret == -ENOSYS) {
+                    printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
+                    printf("Exiting...\n");
+                    se->error = ret;
+
+                    fuse_session_exit(se);
+                    while(!fuse_session_exited(se)) {
```
Nikratio, line 278: This (and the code below) should not be needed.

example/notify_inval_entry.c
Nikratio: I think this code should just be removed. As Bernd point out, it's fragile (that's a pain when running a CI), and secondly I still believe that this tests the kernel rather than libfuse (the return value comes from the kernel).

Nikratio (changes requested):
Thanks for working on this! This now looks good overall, but I think a few clean-ups are needed (reducing the size of the patch).

example/notify_inval_entry.c
HereThereBeDragons: this seems correct

HereThereBeDragons:
about `fuse_session_exit`

i rebuild 2 scenarios i tried before (i tried more but could not remember them all)

1) remove `fuse_session_unmount(se);` within `update_fs_loop()`
both in multithreaded/singlethreaded fuse loop:
- in `update_fs_loop()`: `while(!fuse_session_exited(se))` breaks correctly
- `return NULL; `is called at end of function
BUT `fuse_session_loop()` never returns in main thread.. got stuck

here some output (i added some prints to know where we are):
```
// multithreaded
fuse_lowlevel_notify_expire_entry not supported by kernel
Exiting...
Break update_fs_loop()
Return NULL update_fs_loop()
^CReturn main thread: fuse_session_loop_mt() 2

// Single threaded
fuse_lowlevel_notify_expire_entry not supported by kernel
Exiting...
Break update_fs_loop()
Return NULL update_fs_loop()
^CReturn main thread: fuse_session_loop_mt() 2
```

2) Add within `update_fs_loop()` the functions `fuse_remove_signal_handlers(se);` and `fuse_session_destroy(se);`
Main thread should jump directly to `err_out1`

```
// Works for single threaded:
// Single threaded
fuse_lowlevel_notify_expire_entry not supported by kernel
Exiting...
fuse_session_unmount() in update_fs_loop()
Return main thread: fuse_session_loop_mt() -38

//Occasionally works for multi-threaded:
// Multi-threaded
fuse_lowlevel_notify_expire_entry not supported by kernel
Exiting...
fuse_session_unmount() in update_fs_loop()
Return main thread: fuse_session_loop_mt() -38


//and sometimes you get this:
// Multi-threaded
fuse_lowlevel_notify_expire_entry not supported by kernel
Exiting...
fuse_session_unmount() in update_fs_loop()
fuse_remove_signal_handlers() in update_fs_loop()
fuse_session_destroy() in update_fs_loop()
Break update_fs_loop()
Return NULL update_fs_loop()
Return main thread: fuse_session_loop_mt() 928
fuse: failed to unmount P����U: No such file or directory
free(): invalid pointer
Aborted
```

i have no clue what kind of error 928 should be. (the number is either -38 (= FUSE_DESTROY) or 928)
The failures afterwards should be due to trying to umount again as it does goto err_out1

HereThereBeDragons:
Test system is `5.19.0-43-generic #44~22.04.1-Ubuntu`
I read somewhere it could be related to `SA_RESTART` signal that might be differently set on different systems?
( https://stackoverflow.com/questions/8903448/libfuse-exiting-fuse-session-loop ) 
(I did not try it out myself to work with that)

Also `raise(SIGINT)` or `raise(SIGTERM)` at the end of `update_fs_loop()` does nothing. Only `raise(SIGKILL)` kills the program but leaves a broken mountpoint

example/notify_inval_entry.c
HereThereBeDragons: i dont see how we can remove it. maybe we can add some locking around the manipulation of the counter.

when we `expire` `forget()` is not called. therefore the counter is not decreased. as the counter is not decreased the `if` within the `while`-loop is always called and as such always `expire`  is called: on existing or old non-existing files. only because of the counter we know if we should expect and error or a successful execution of `expire`.

if you have any better idea or if i mistunderstand something, please let me know

HereThereBeDragons:
For whatever reason my reply is not shown to your comment about the counters:
https://github.com/libfuse/libfuse/pull/789/files#r1246384851
> i dont see how we can remove it. maybe we can add some locking around the manipulation of the counter.
> 
> when we `expire` `forget()` is not called. therefore the counter is not decreased. as the counter is not decreased the `if` within the `while`-loop is always called and as such always `expire` is called: on existing or old non-existing files. only because of the counter we know if we should expect and error or a successful execution of `expire`.
> 
> if you have any better idea or if i mistunderstand something, please let me know

Nikratio:
> > i dont see how we can remove it. maybe we can add some locking around the manipulation of the counter.
> > when we `expire` `forget()` is not called. therefore the counter is not decreased. as the counter is not decreased the `if` within the `while`-loop is always called and as such always `expire` is called: on existing or old non-existing files. only because of the counter we know if we should expect and error or a successful execution of `expire`.
> > if you have any better idea or if i mistunderstand something, please let me know

My suggestion is to remove all counting (i.e., get rid of `lookup_count` and `old_lookup_count` variables) and instead just do this:

```
// The kernel may have expired the entry on its own already, in which case
// ENOENT is the expected return.
assert(ret == 0 || ret == -ENOENT);
```

Nikratio:
> about `fuse_session_exit`
> 
> i rebuild 2 scenarios i tried before (i tried more but could not remember them all)
> 
>     1. remove `fuse_session_unmount(se);` within `update_fs_loop()`
>        both in multithreaded/singlethreaded fuse loop:
> 
> 
>     * in `update_fs_loop()`: `while(!fuse_session_exited(se))` breaks correctly
> 
>     * `return NULL; `is called at end of function
>       BUT `fuse_session_loop()` never returns in main thread.. got stuck
> 
> 
> here some output (i added some prints to know where we are):
> 
> ```
> // multithreaded
> fuse_lowlevel_notify_expire_entry not supported by kernel
> Exiting...
> Break update_fs_loop()
> Return NULL update_fs_loop()
> ^CReturn main thread: fuse_session_loop_mt() 2

Thanks for explaining! I have a hypothesis: fuse_session_exit() just sets a flag that causes worker threads to "exit on the next opportunity".

Could you try to issue a request to the mountpoint (e.g. just run `ls`) on it by hand after `fuse_session_exit` has been called? If that helps, then I have some ideas for how to solve this properly.

HereThereBeDragons:
> Could you try to issue a request to the mountpoint (e.g. just run ls) on it by hand after fuse_session_exit has been called? If that helps, then I have some ideas for how to solve this properly.

Yes, that works.

```
// fuse mountpoint terminal
> sudo example/notify_inval_entry  -f --update-interval=1 --timeout=5 --only-expire /mnt/test/

// other terminal
> ls /mnt/test/
Time_is_13h_44m_29s

// fuse mountpoint terminal
fuse_lowlevel_notify_expire_entry not supported by kernel
Exiting...
update_fs_loop() fuse_session_exit()
update_fs_loop() break
update_fs_loop() return NULL

// stuck here
// other terminal
> ls /mnt/test/
ls: cannot open directory '/mnt/test/': Software caused connection abort

// fuse mountpoint terminal
main Thread: ret -38

**Comment by HereThereBeDragons** [comment]
> My suggestion is to remove all counting (i.e., get rid of `lookup_count` and `old_lookup_count` variables) and instead just do this:

ok will do

**Comment by Nikratio** [comment]
> > Could you try to issue a request to the mountpoint (e.g. just run ls) on it by hand after fuse_session_exit has been called? If that helps, then I have some ideas for how to solve this properly.
> 
> Yes, that works.

Great! I have just committed a change that better documents this behavior.

As a workaround, you can send `SIGPIPE` to the main thread after calling fuse_session_exit(). E.g. like this:

```
diff --git a/example/notify_inval_entry.c b/example/notify_inval_entry.c
index 8af31bc..dde69bd 100644
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -88,6 +88,7 @@
 #include <stddef.h>
 #include <unistd.h>
 #include <pthread.h>
+#include <signal.h>
 
 #define MAX_STR_LEN 128
 static char file_name[MAX_STR_LEN];
@@ -212,6 +213,7 @@ static int reply_buf_limited(fuse_req_t req, const char *buf, size_t bufsize,
         return fuse_reply_buf(req, NULL, 0);
 }
 
+static pthread_t main_thread;
 static void tfs_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
                         off_t off, struct fuse_file_info *fi) {
     (void) fi;
@@ -253,6 +255,7 @@ static void* update_fs_loop(void *data) {
     struct fuse_session *se = (struct fuse_session*) data;
     char *old_name;
 
+    int i=0;
     while(1) {
         old_name = strdup(file_name);
         update_fs();
@@ -267,6 +270,11 @@ static void* update_fs_loop(void *data) {
         }
         free(old_name);
         sleep(options.update_interval);
+        if  (i++ == 5) {
+            printf("Setting exit flag\n");
+            fuse_session_exit(se);
+            pthread_kill(main_thread, SIGPIPE);
+        }
     }
     return NULL;
 }
@@ -325,6 +333,7 @@ int main(int argc, char *argv[]) {
     fuse_daemonize(opts.foreground);
 
     /* Start thread to update file contents */
+    main_thread = pthread_self();
     ret = pthread_create(&updater, NULL, update_fs_loop, (void *)se);
     if (ret != 0) {
         fprintf(stderr, "pthread_create failed with %s\n",

```

**Comment by bsbernd** [comment]
I guess the fuse_session_exit issue is read/splice from /dev/fuse?
Isn't it more obvious to let this test use a single threaded loop only? And maybe we should make fuse_session_exit should send a signal to all worker threads (SIGINT?)?

File: example/notify_inval_entry.c (hunk 3 above)
Comment (bsbernd) [reply example/notify_inval_entry.c]: This is kind of a test for libfuse + kernel and kerne sidel is not entirely returning what is actually does - this code is hacking it around and that certainly works most of the time.

File: example/notify_inval_entry.c [hunk 10]
```diff
@@ -275,21 +274,15 @@ static void* update_fs_loop(void *data) {
                     se->error = ret;
 
                     fuse_session_exit(se);
-                    while(!fuse_session_exited(se)) {
-                        sleep(1);
-                    }
-                    fuse_session_unmount(se);
+                    // needed to correctly exit fuse_session_loop()
+                    pthread_kill(main_thread, SIGPIPE);
```
bsbernd, line 278: I don't have a strong opinion, but I wonder if this could be avoided, if we would convert to 3.12 API and then only allow 1 thread, even without "opts.singlethread".

bsbernd (commented):
Thanks for your updates! I think the latest version looks good, I'm going to download it as single patch in the morning to get the full picture.

example/notify_inval_entry.c
Nikratio: Yes, but let's not block on this. We've given HereThereBeDragons enough pain with this as-is, I think.

Nikratio:
> I guess the fuse_session_exit issue is read/splice from /dev/fuse? Isn't it more obvious to let this test use a single threaded loop only?

That won't help, we still need to wake up that single thread. (There will always be a second, non-FUSE thread running the update_fs_loop).

> And maybe we should make fuse_session_exit should send a signal to all worker threads (SIGINT?)?

Yes, I was thinking about that. However, fuse_session_exit() may be called in signal context itself and I couldn't immediately tell if that's a problem. Also, the session object does not give access to the worker thread ids. So this is stuff for a separate pull request.

example/notify_inval_entry.c
```diff
@@ -78,13 +78,16 @@
 
 #define FUSE_USE_VERSION 34
 
+#include <fuse_i.h>
```
Nikratio, line 81: Why is this needed? That should only be used internally in libfuse, not in an example.

example/notify_inval_entry.c
```diff
@@ -253,14 +257,33 @@ static void* update_fs_loop(void *data) {
...
+                   (se, FUSE_ROOT_ID, old_name, strlen(old_name));
+
+                // no kernel support
+                if (ret == -ENOSYS) {
+                    printf("fuse_lowlevel_notify_expire_entry not supported by kernel\n");
+                    printf("Exiting...\n");
+                    se->error = ret;
+
+                    fuse_session_exit(se);
+                    // needed to correctly exit fuse_session_loop()
```
Nikratio, line 277: Suggest to phrase this as `// Make sure to exit now, rather than on next request from userspace`. There is nothing incorrect about not sending SIGPIPE.

example/notify_inval_entry.c
```diff
@@ -324,6 +347,9 @@ int main(int argc, char *argv[]) {
 
     fuse_daemonize(opts.foreground);
 
+    // needed to correctly exit fuse_session_loop()
```
Nikratio, line 350: No, this is needed to sent a signal to the main thread so it gets worked up and notices that it should exit. It is not about how to exit, but when to exit.

Nikratio (changes requested):

example/notify_inval_entry.c
HereThereBeDragons: needed it for `se->error = ret;`
 https://github.com/libfuse/libfuse/blob/0cc6fe0d9b560786e63a075c77d5014e7bca1590/example/notify_inval_entry.c#L270-L276
 
i will remove it as with the current design it is not needed anymore.

example/notify_inval_entry.c
```diff
@@ -78,13 +78,15 @@
 
 #define FUSE_USE_VERSION 34
 
+#include <fuse_kernel.h>
```
Nikratio, line 81: What's this needed for? I think this should only be needed when communicating directly with the kernel, not when going through libfuse.

HereThereBeDragons: think that was part of the `fuse_i.h` needed to be able to set `se->error` which is not needed anymore

Nikratio (approved):

Nikratio:
Thanks for working on this!

HereThereBeDragons:
thanks also for all the feedback and improvements!

PR #790, opened 2023-05-08
Do not pass unsupported mount options to the kernel.

Nikratio:
The filesystem daemon is responsible for implementing eg. st_atime updates, so passing options like relatime to the kernel results in them being silently ignored. Instead, such options need to be interpreted (and filtered out) by the filesystem daemon.


bsbernd (approved):
Thanks Nikolaus, looks good to me.

PR #793, opened 2023-05-18
Add minimal token permissions

### LINKED ISSUE #792: Set read-only workflow permissions
Currently, libfuse's workflows run with write-all permissions. This is dangerous, since it opens the project up to supply-chain attacks. [GitHub itself](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets) recommends ensuring all workflows run with minimal permissions.

I've taken a look at the workflows and they don't need anything other than read permissions to the repository.

This issue can be solved in two ways:
- add top-level read-only permissions to both workflows; and/or
- set the default token permissions to read-only in the repo settings.

I'll be sending a PR along with this issue that sets the top-level permissions. If you instead (or also) wish to modify the default token permissions:

1. Open the repo settings
2. Go to [Actions > General](https://github.com/libfuse/libfuse/settings/actions)
3. Under "Workflow permissions", set them to "Read repository contents and packages permissions"

---

**Disclosure:** My name is Pedro and I work with Google and the [Open Source Security Foundation (OpenSSF)](https://www.openssf.org/) to improve the supply-chain security of the open-source ecosystem.

pnacht:
Closes #792.

As mentioned in the issue above, this PR ensures the workflows run with read-only permissions. This protects the project from supply-chain attacks.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
     branches: [master]
 
+permissions:
+  contents: read
+
 jobs:
   codespell:
     name: Check for spelling errors
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches:
       - master
+
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: '${{ matrix.os }}'


bsbernd:
Hi @pnacht I have one question - why isn't that default in github? Shouldn't it be the other way around - write permission should get a config option?

pnacht:
Hey @bsbernd, as far as I understand it, it's a historical artifact. When GHA was released, they hadn't implemented permissioning yet, so everything was effectively write-all.

When they released a permissions system, they kept the default as write-all for backwards compatibility. But as I mention in the linked issue, GitHub itself [recommends](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets) changing the default (or at least ensuring Actions declare minimal permissions). 

Thankfully earlier this year, [GitHub's changed the default for new projects](https://github.blog/changelog/2023-02-02-github-actions-updating-the-default-github_token-permissions-to-read-only/). So at least anything since February 2023 will run with safer defaults, but the rest of us are still in the wild west.

Let me know if there's anything else I can help you with!

bsbernd:
Thanks @pnacht, very helpful! So you are looking from project to project and notifying all of them? Tthat is quite some work! I'm going to send a patch to libfabric tomorrow.

PR #806, opened 2023-06-19
Improve wording of user_allow_other usage instructions

waldyrious:
While I appreciate the detailed instructions for `user_allow_other`, IMO these small grammatical, punctuation and wording adjustments improve the clarity of the content.

In addition, I took the opportunity to remove a dated comment from the `mount_max` usage instructions immediately below.


Nikratio:
Thanks!

PR #807, opened 2023-06-27
passthrough_hp benchmark mode

bsbernd:
This adds a read/write benchmark mode for passthrough_hp. Minor commits are the forgotten handler for clone_fd and adding in parallel dio writes.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -81,6 +81,12 @@ using namespace std;
 #define SFS_DEFAULT_THREADS "-1" // take libfuse value as default
 #define SFS_DEFAULT_CLONE_FD "0"
 
+/* In read/write bypass mode typically a buffer copy is done, with buf_copy = 2
+ * that is avoided as well. For read() that is not possible right now,
+ * though.
+ */
+#define SFS_BYPASS_BUF_COPY 2
+
 /* We are re-using pointers to our `struct sfs_inode` and `struct
    sfs_dirp` elements as inodes and file handles. This means that we
    must be able to store pointer a pointer in both a fuse_ino_t
@@ -158,6 +164,12 @@ struct Fs {
     size_t num_threads;
     bool clone_fd;
     std::string fuse_mount_options;
+
+    /* for fuse benchmarking, to avoid the need of an indefinitely fast
+     * backend file system - will skip file writes/reads
+     */
+    bool bypass_rw;
+    char *bypass_buf;
 };
 static Fs fs{};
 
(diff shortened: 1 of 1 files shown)


example/passthrough_hp.cc
```diff
@@ -928,8 +942,22 @@ static void do_read(fuse_req_t req, size_t size, off_t off, fuse_file_info *fi)
         FUSE_BUF_IS_FD | FUSE_BUF_FD_SEEK);
     buf.buf[0].fd = fi->fh;
     buf.buf[0].pos = off;
+    enum fuse_buf_copy_flags copy_flags = FUSE_BUF_COPY_FLAGS;
+
+    if (fs.bypass_rw) {
+        /* special benchmark, to see how fast fuse works - just copy
+         * from a memory buffer
```
Nikratio, line 949: Have you considered splicing the data from /dev/zero instead? That way you could avoid the buffer copy.

example/passthrough_hp.cc
```diff
@@ -946,8 +974,30 @@ static void do_write_buf(fuse_req_t req, size_t size, off_t off,
...
+        if (bypass == SFS_BYPASS_BUF_COPY) {
+            /* the received buffer is not used at all */
+            size_t res = size;
+            fuse_reply_write(req, (size_t)res);
+            return;
+        }
 
-    auto res = fuse_buf_copy(&out_buf, in_buf, FUSE_BUF_COPY_FLAGS);
+        /* the received buffer is used for another memory copy */
+        out_buf.buf[0].fd = -1;
```
Nikratio, line 994: Instead of overwriting the previous definiton of out_buf, can we move the first definition into an `else` clause?

example/passthrough_hp.cc
```diff
@@ -1245,6 +1298,14 @@ static cxxopts::ParseResult parse_options(int argc, char **argv) {
     fs.nosplice = options.count("nosplice") != 0;
     fs.num_threads = options["num-threads"].as<int>();
     fs.clone_fd = options["clone-fd"].as<bool>();
+
+    fs.bypass_rw = options["bypass-rw"].as<int>();
+    if (fs.bypass_rw) {
+        std::cout << "===== Warning RW IO bypassed! Data corruption expected! ====="
+                  << std::endl;
+        fs.bypass_buf = reinterpret_cast<char *>(new uint8_t [1024 * 1024 * 2]);
```
Nikratio, line 1306: Can we document why this is the right buffer size, and check for out of bounds writes and reads in the callbacks?

Nikratio (changes requested):
Thanks! Looks good overall, just a few minor comments.

bsbernd:
Thanks for your comments, I will try to update today or tomorrow

Nikratio:
@bsbernd Any update on this?

bsbernd:
Yes sorry, got busy with kernel side again. Let me close for now and I will send out a new version once I'm ready (hopefully during the next days).

Nikratio:
Let's keep this one open so the comments don't get lost. I've converted it to draft instead.

bsbernd:
Part of memfs_ll now and not needed in passthrough_hp anymore.

Issue #717, opened 2022-12-21
Strip `remount` option in mount.fuse


chasecaleb:
I encountered the same issue as #211, except in this case it's for option `remount`.  Based on that issue and the fix in #221, maybe this should be solved the same way with a one line addition of `remount` to the ignored options list? If that's the correct solution then I'm willing to make a PR -- although it's such a trivial change it might be easier for a maintainer to do it instead of dealing with the PR process. I'm open to whatever works best for you.

---

Here's how I encountered this on NixOS with an [open-vm-tools](https://github.com/vmware/open-vm-tools/) (VMware guest tools) mount:

```
$ sudo systemctl cat run-vmblock\\x2dfuse.mount
# /etc/systemd/system/run-vmblock\x2dfuse.mount
[Unit]
ConditionVirtualization=vmware
Description=VMware vmblock fuse mount
Documentation=https://github.com/vmware/open-vm-tools/blob/master/open-vm-tools/vmblock-fuse/design.txt

[Mount]
Options=subtype=vmware-vmblock,default_permissions,allow_other
Type=fuse
What=/nix/store/1y562cdla1a6jxvs96z14vl7lqps5mgd-test-open-vm-tools-12.1.5/bin/vmware-vmblock-fuse
Where=/run/vmblock-fuse

$ sudo systemctl reload run-vmblock\\x2dfuse.mount; journalctl -u run-vmblock\\x2dfuse.mount --since "-5s"
Job failed. See "journalctl -xe" for details.
Dec 21 13:49:57 dev systemd[1]: Reloading VMware vmblock fuse mount...
Dec 21 13:49:57 dev mount[172663]: fuse: unknown option(s): `-o remount'
Dec 21 13:49:57 dev systemd[1]: run-vmblock\x2dfuse.mount: Mount process exited, code=exited, status=3/NOTIMPLEMENTED
Dec 21 13:49:57 dev systemd[1]: Reload failed for VMware vmblock fuse mount.
```

And some version info:
```
$ systemctl --version
systemd 252 (252.1)
[...]

PR #811, opened 2023-07-05
xfstests example: Use export in local.config and remove comment

bsbernd:
mount.fuse.passthrough had a comment explaining that these variables
do not work if specified in local.config. Actually the variables
just had to be exported - remove the comment and export the example
variables.

### DIFF
--- a/xfstests/local.config
+++ b/xfstests/local.config
@@ -7,6 +7,13 @@ export FUSE_SUBTYP=.passthrough
 export MOUNT_OPTIONS=""
 export TEST_FS_MOUNT_OPTS=""
 
-export PASSTHROUGH_PATH=<path-to-libfuse-build>/example/passthrough_hp
+# If PASSTHROUGH_PATH is unset, the mount helper is going to look
+# for the binary '${FUSE_SUBTYP}', though omitting the leading dot '.'.
+# Example:
+#       with FUSE_SUBTYP=".passthrough", the mount helper is called
+#       'mount.fuse.passthrough' and that would try to run
+#       'passthrough'.
+# export PASSTHROUGH_PATH=<path-to-libfuse-build>/example/passthrough_hp
+
 export SCRATCH_SOURCE=/mnt/src/scratch
 export TEST_SOURCE=/mnt/src/test
--- a/xfstests/local.config
+++ b/xfstests/local.config
@@ -7,6 +7,6 @@ export FUSE_SUBTYP=.passthrough
 export MOUNT_OPTIONS=""
 export TEST_FS_MOUNT_OPTS=""
 
-PASSTHROUGH_PATH=/home/nikratio/libfuse/build/example/passthrough_hp
-SCRATCH_SOURCE=/mnt/src/scratch
-TEST_SOURCE=/mnt/src/test
+export PASSTHROUGH_PATH=<path-to-libfuse-build>/example/passthrough_hp
+export SCRATCH_SOURCE=/mnt/src/scratch
+export TEST_SOURCE=/mnt/src/test
--- a/xfstests/local.config
+++ b/xfstests/local.config
@@ -1,7 +1,9 @@
-export TEST_DEV=non1
+export TEST_DEV=source:/mnt/src/test
 export TEST_DIR=/mnt/test
-export SCRATCH_DEV=non2
+
+export SCRATCH_DEV=source:/mnt/src/scratch
 export SCRATCH_MNT=/mnt/scratch
+
 export FSTYP=fuse
 export FUSE_SUBTYP=.passthrough
 export MOUNT_OPTIONS=""
@@ -14,6 +16,3 @@ export TEST_FS_MOUNT_OPTS=""
 #       'mount.fuse.passthrough' and that would try to run
 #       'passthrough'.
 # export PASSTHROUGH_PATH=<path-to-libfuse-build>/example/passthrough_hp
-
-export SCRATCH_SOURCE=/mnt/src/scratch
-export TEST_SOURCE=/mnt/src/test
--- a/xfstests/mount.fuse.passthrough
+++ b/xfstests/mount.fuse.passthrough
@@ -12,17 +12,14 @@ mntopts="$1"
 shift
 
 # source can be provided as NFS style device (e.g. TEST_DEV=source:/${TEST_SOURCE})
-# or by exporting the {TEST,SCRATCH}_SOURCE env vars
 # and/or it can already be inside mount options (passthrough_ll style)
 if ( echo "$mntopts" | grep -q "source=" ) ; then
 	# Don't pass source as position argument
 	source=
 elif [[ "$dev" == "source:"* ]]; then
 	source="${dev#"source:"}"
-elif [ "$dev" = "${SCRATCH_DEV}" ]; then
-	source="${SCRATCH_SOURCE}"
-elif [ "$dev" = "${TEST_DEV}" ]; then
-	source="${TEST_SOURCE}"
+else
+    >&2 echo "passthrough source is undefined, aborting!"
 fi
 
 if ( echo "$mntopts" | grep -q remount ) ; then
--- a/xfstests/mount.fuse.passthrough
+++ b/xfstests/mount.fuse.passthrough
@@ -2,16 +2,40 @@
 
 ulimit -n 1048576
 
-# It would be easier if we could just set SCRATCH_DEV and TEST_DEV to the source directory
-# path in local.options. Unfortunately, setting these variables to a path seems get
-# xfstests all worked up (even though it should treat these as opaque values), and it
-# refuses to even start running any tests).
 dev="$1"
 shift
-if [ "$dev" = "${SCRATCH_DEV}" ]; then
-  source="${SCRATCH_SOURCE}"
-else
-  source="${TEST_SOURCE}"
+mnt="$1"
+shift
+# -o
+shift
+mntopts="$1"
+shift
+
+# source can be provided as NFS style device (e.g. TEST_DEV=source:/${TEST_SOURCE})
+# or by exporting the {TEST,SCRATCH}_SOURCE env vars
+# and/or it can already be inside mount options (passthrough_ll style)
+if ( echo "$mntopts" | grep -q "source=" ) ; then
+	# Don't pass source as position argument
+	source=
+elif [[ "$dev" == "source:"* ]]; then
+	source="${dev#"source:"}"
+elif [ "$dev" = "${SCRATCH_DEV}" ]; then
+	source="${SCRATCH_SOURCE}"
+elif [ "$dev" = "${TEST_DEV}" ]; then
+	source="${TEST_SOURCE}"
 fi
 
-exec "$PASSTHROUGH_PATH" -o fsname=$dev,allow_other "${source}" "$@"
+if ( echo "$mntopts" | grep -q remount ) ; then
+	exec mount -i "$dev" "$mnt" -o "$mntopts"
+fi
+
+# set default to SUBTYPE (extracted from this script name)
+# example:
+#   Copy or link this script to /sbin/mount.fuse.passthrough_hp
+#   If xfstests local.config does not set PASSTHROUGH_PATH,
+#   PASSTHROUGH_PATH will be set to 'passthrough_hp' and exec below
+#   will look that up from $PATH
+
+[ -n "$PASSTHROUGH_PATH" ] || PASSTHROUGH_PATH=${0#*mount.fuse.}
+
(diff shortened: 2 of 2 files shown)


xfstests/mount.fuse.passthrough
```diff
@@ -2,10 +2,6 @@
 
 ulimit -n 1048576
 
-# It would be easier if we could just set SCRATCH_DEV and TEST_DEV to the source directory
```
Nikratio, line 5: I think there's some confusion between SCRATCH_DEV and SCRATCH_SOURCE. What this comment says is that it would be great if we could set SCRATCH_DEV to the value that we currently store in SCRATCH_SOURCE, and get rid of SCRATCH_SOURCE completely.

Are you saying that you were able to run xfstests with this change? In this case we should get rid of the comment, but also of the SCRATCH_SOURCE variable.

Or are you saying that as-is, xfstests was not working for you unless you added the `export`? In that case we should keep the comment and add the exports.

Nikratio (changes requested):

xfstests/mount.fuse.passthrough
bsbernd: Ah ok, sorry, but I'm not able to extract the meaning from the current comment :) I can test/change variables or comment when I'm back from vacation in 2 weeks.

Nikratio: What the comment says is that `local.config` should look like this:

```
export TEST_DEV=/mnt/src/test
export TEST_DIR=/mnt/test
export SCRATCH_DEV=/mnt/src/scratch
export FSTYP=fuse
export FUSE_SUBTYP=.passthrough
export MOUNT_OPTIONS=""
export TEST_FS_MOUNT_OPTS=""
PASSTHROUGH_PATH=<path-to>/libfuse/build/example/passthrough_hp
```

and mount.fuse.passthrough like this:

```
#!/bin/bash
ulimit -n 1048576
exec "$PASSTHROUGH_PATH" -o fsname=$dev,allow_other "$@"
```

I believe this should work in theory and is much simpler.

Unfortunately, this causes xfstests to get very confused and nothing works (at least at the time of writing this comment).

bsbernd: Can we try that when I'm back from vacation? I need to pack things and still submit some kernel patches (while I actually shouldn't be in front of my monitor ;) ).

Nikratio: @bsbernd Any update on this?

bsbernd: I will come back to this later on today or tomorrow, once I have sent out new atomic open kernel patches.

Nikratio:
This patch looks at least at first glance totally unlike the first one.

Could you explain what problem is solved by it / what additional functionality enabled?

amir73il:
> This patch looks at least at first glance totally unlike the first one.
> 
> Could you explain what problem is solved by it / what additional functionality enabled?
@Nikratio , you wrote in the comment in mount helper:

> # It would be easier if we could just set SCRATCH_DEV and TEST_DEV to the source directory
> # path in local.options. Unfortunately, setting these variables to a path seems get
> # xfstests all worked up (even though it should treat these as opaque values), and it
> # refuses to even start running any tests).

This is exactly what my patch does.
As written in the new comment, now you can still set TEST/SCRATCH_SOURCE vars,
but you can instead set TEST_DEV in format source://<source path> instead -
xfstest will treat them as opaque values, but the mount helper parses the source path to use it.

A third way is to include -osource=<source path> in mount options the way that 
passthrough_ll is expecting to get the source path argument and then the same
mount helper works for both passthrough_hp and passthrough_ll (I tested).
That is also why PASSTHROUGH_PATH default to the helper name.
You can install this helper as /sbin/passthough_ll and /sbin/passthrough_hp

Another unrelated change is to support -oremount mount option in the mount helper,
which means that PASSTHROUGH_PATH program need not be executed at all.
This helps pass a few tests that try to remount ro/rw.

Nikratio:
> Another unrelated change is to support -oremount mount option in the mount helper, which means that PASSTHROUGH_PATH program need not be executed at all. This helps pass a few tests that try to remount ro/rw.

Are you sure this is safe? 

My impression is that a remount operation for fuse filesystems requires active cooperation of the userspace daemon (since mount options like `relatime` need to be implemented there, and might be changed during the remount). See also issue #717 for more discussion of this. Users may also reasonably expect that "remount" means re-exec'ing the userspace daemon (to e.g. update to  a new version in-place). So I think the safest option is to always refuse a remount as unsupported.

I guess if this makes some xfstests pass, then by definition the tested remount functionality works. But I'm wondering if it may give the wrong impression that remount in general is supported...

xfstests/local.config
```diff
@@ -7,6 +7,6 @@ export FUSE_SUBTYP=.passthrough
 export MOUNT_OPTIONS=""
 export TEST_FS_MOUNT_OPTS=""
 
-PASSTHROUGH_PATH=/home/nikratio/libfuse/build/example/passthrough_hp
-SCRATCH_SOURCE=/mnt/src/scratch
-TEST_SOURCE=/mnt/src/test
+export PASSTHROUGH_PATH=<path-to-libfuse-build>/example/passthrough_hp
+export SCRATCH_SOURCE=/mnt/src/scratch
+export TEST_SOURCE=/mnt/src/test
```
Nikratio, line 12: From your comment in the main discussion thread:

> As written in the new comment, now you can still set TEST/SCRATCH_SOURCE vars,
> but you can instead set TEST_DEV in format source://'source path' instead -
> xfstest will treat them as opaque values, but the mount helper parses the source path to use it.

If that's the case, why not do it here? (remove `SCRATCH_SOURCE` and `TEST_SOURCE` and use `TEST_DEV` and `SCRATCH_DEV` instead). If that works, then I see no reason to keep the workaround variables around.

xfstests/local.config
```diff
@@ -7,6 +7,6 @@ export FUSE_SUBTYP=.passthrough
 export MOUNT_OPTIONS=""
 export TEST_FS_MOUNT_OPTS=""
 
-PASSTHROUGH_PATH=/home/nikratio/libfuse/build/example/passthrough_hp
-SCRATCH_SOURCE=/mnt/src/scratch
-TEST_SOURCE=/mnt/src/test
+export PASSTHROUGH_PATH=<path-to-libfuse-build>/example/passthrough_hp
```
Nikratio, line 10: If this is now optional (which I think it is), can we add a comment that explains what the default value is? Is the following correrct?

```
# If this not set, attempt to execute mount.fuse.<helper_name> from $PATH
PASSTHROUGH_PATH=...
```

amir73il: Not exactly.
Mount helper to execute is determined from the -t argument to mount command,
So its going to be /sbin/mount.fuse.$FUSE_SUBTYP.
The fuse server binary to execute is attempted to be extracted from the mount helper name. 
So if the server binary can be found is $PATH and it's name is $FUSE_SUBTYPE,
No need to specify the server path explicitly.
In the local.config example, we can change the subtype to passthrough_hp

The downside of this approach is that a mount helper needs to be installed per subtype when testing different passthrough servers. 
Otoh maybe the right approach is to maintain different mount helpers per passthrough example because as you can see the source path argument is passed differently to passthrough_ll and I had to do something in the mount helper to work around that

amir73il:
> > Another unrelated change is to support -oremount mount option in the mount helper, which means that PASSTHROUGH_PATH program need not be executed at all. This helps pass a few tests that try to remount ro/rw.
> 
> Are you sure this is safe?
> 
> My impression is that a remount operation for fuse filesystems requires active cooperation of the userspace daemon (since mount options like `relatime` need to be implemented there, and might be changed during the remount). See also issue #717 for more discussion of this. Users may also reasonably expect that "remount" means re-exec'ing the userspace daemon (to e.g. update to a new version in-place). So I think the safest option is to always refuse a remount as unsupported.
> 
> I guess if this makes some xfstests pass, then by definition the tested remount functionality works. But I'm wondering if it may give the wrong impression that remount in general is supported...

The thing is that a mount helper is not mandatory to mount fuse.
If you just run the passthough_hp executable, you could happily remount,ro/rw
as far as I can tell the only thing that kernel will do in this case in fuse_reconfigure() is sync_filesystem() to flush out all the write cache.

As to remount,noatime/relatime it also works and this is actually a per bind mount option,
so fuse server cannot be responsible for it no matter what, because fuse server does not have the bind mount context.

I have seen that in xfstests generic/003 fails due to `fuse: unknown option(s): '-o relatime'` (it is not a remount) so I guess the mount helper needs to filter out and ignore those per bind mount options that are not its concern - not sure.

xfstests/local.config
Nikratio: Let's drop the old ways then :-).

xfstests/local.config
Nikratio: Can we add something like "If unset, defaults to looking for a binary named after the filesystem subtype in $PATH"? (I'd do these changes myself, but it seems I don't have edit permissions on the pull request).

Nikratio:
> As to remount,noatime/relatime it also works and this is actually a per bind mount option, so fuse server cannot be responsible for it no matter what, because fuse server does not have the bind mount context.

I am quite certain that this is not true, although I agree that it *should* work that way. Unfortunately I don't remember where we last discussed this, but I'm pretty sure it was either with Miklos or Bernd. Proper support of atime-related mount options currently requires help from userspace.

That said, this is a tangential issue and I don't want to overcomplicate this simple patch. So I'm fine with keeping that change in. Let's just remove the old variables and add a comment explaining the default for the server binary.

bsbernd:
@Nikratio I don't know how to grant you write permissions. 

Please check the updated commits, I tried to add explaining comments.

bsbernd:
Regarding atime, I had submitted a patch to add related mount options. You had then discussed with Miklos (during LSFMM) and we also continued to discuss and eventually you removed the patch again. Issue is that fuse kernel is not going to send atime updates - which is 'noatime'.

xfstests/local.config
Nikratio: @bsbernd Could you drop the definitions of `SCRATCH_SOURCE` and `TEST_SOURCE` please?

xfstests/mount.fuse.passthrough
```diff
@@ -2,16 +2,40 @@
...
+# and/or it can already be inside mount options (passthrough_ll style)
+if ( echo "$mntopts" | grep -q "source=" ) ; then
+	# Don't pass source as position argument
+	source=
+elif [[ "$dev" == "source:"* ]]; then
+	source="${dev#"source:"}"
+elif [ "$dev" = "${SCRATCH_DEV}" ]; then
+	source="${SCRATCH_SOURCE}"
+elif [ "$dev" = "${TEST_DEV}" ]; then
+	source="${TEST_SOURCE}"
```
Nikratio, line 25: @bsbernd Please remove the references to `*_SOURCE` here as well.

Nikratio:
> Regarding atime, I had submitted a patch to add related mount options. You had then discussed with Miklos (during LSFMM) and we also continued to discuss and eventually you removed the patch again. Issue is that fuse kernel is not going to send atime updates - which is 'noatime'.

Ah, thanks for remembering! Yes, currently the kernel does not update {cma}-time (unless writeback caching is active, in which case at least mtime is updated). This means that the userspace daemon is responsible for updating attributes on read/write/setattr, and thus also needs to interpret the related mount options. Thus, remounting as currently implemented in the xfstest script will not work correctly if these mount options are changed.

xfstests/mount.fuse.passthrough
bsbernd: Thanks, I was already in the process to do that. Can you please check the new  additional patch?

Nikratio:
Thanks Bernd! @amir73il, does that look good to you too?

amir73il:
LGTM thanks @bsbernd for posting my patch :)

amir73il:
> > Regarding atime, I had submitted a patch to add related mount options. You had then discussed with Miklos (during LSFMM) and we also continued to discuss and eventually you removed the patch again. Issue is that fuse kernel is not going to send atime updates - which is 'noatime'.
> 
> Ah, thanks for remembering! Yes, currently the kernel does not update {cma}-time (unless writeback caching is active, in which case at least mtime is updated). This means that the userspace daemon is responsible for updating attributes on read/write/setattr, and thus also needs to interpret the related mount options. Thus, remounting as currently implemented in the xfstest script will not work correctly if these mount options are changed.

I agree that the *atime mount options don't wont correctly, because the server is in control of atime, but try to do remount without a mount helper and see what happens... Specifically try to create a bind mount of fuse and change just the bind mount to noatime with remount and see what happens. Just the bind mount options are changed. 

On the plus side, I did not see any xfstest that tries to change *atime options with remount. generic/003 does a unmount/mount cycle to change the *atime mount options - this test fails with fuse passthrough.

Nikratio (approved):

PR #813, opened 2023-07-06
Atomic open for 6.4

bsbernd:


bsbernd:
New kernel side patch is upcoming.

youenn98:
I'm Yuan, working on Google's crosvm project, which utilizes FUSE for virtiofs. I'm particularly interested in the "fuse: full atomic open and atomic-open-revalidate" kernel patch and have been closely following the v9 kernel patch. However, I've noticed some ambiguity regarding the behavior of atomic_open for symbolic links.

The previous behavior for fuse_create_open() will only do lookup but not open the symbolic link. But, with the full atomic open kernel patch. My understanding is that when the kernel performs an atomic_open operation on a symbolic link, the dentry returned from the FUSE server contains the inode pointing to the opened symbolic link. However, after atomic_open() is called, the may_open() function in namei.c checks the node's i_mode and identifies it as a symbolic link, resulting in an ELOOP error. 

My question is: what is the expected behavior for opening a symbolic link, both on the kernel side and the server side? Is it possible for the fuse server to return the dentry containing the inode of the link destination instead of the inode of the symbolic link itself?

bsbernd:
@youenn98 would it be possible to ask your question as reply to the kernel patches? I.e. on linux-fsdevel? This is here the libfuse part - possibly there also needs to be updated something, but kernel part is more complex and the core of behavior. I also do not exclude that I missed something with atomic open - please post as reply to kernel patches and I will also (later on today) look into symlink open.

youenn98:
I made the question to the kernel thread patches, thank you a lot.

bsbernd:
Closing it for now, until we find time to update kernel patches.

PR #822, opened 2023-07-31
Pass cache_readdir and keep_cache from high level API

### LINKED ISSUE #821: Can't use readdir cache from high-level API
Hello, I am unable to use readdir cache from the high-level API: 

As suggested in documentation, I am setting `cache_readdir = 1` and `keep_cache = 1` from my opendir handler, however, those flags are never passed to the kernel. And I never see readdir requests cached.

Here is what I think is happening:
`fuse_lib_opendir` is called with `fuse_file_info *llfi`. This function then creates new `fuse_file_info fi` and calls `fuse_fs_opendir` with `fi`. In the end, `fuse_reply_open(req, llfi)` is called, meaning that any changes I made to `fi` - are discarded.

I would love to fix this issue, but I am not sure which way would be better:
1. [Just copy the flags](https://github.com/libfuse/libfuse/commit/8bb0ba1ffc0c1a748d20c92c1b418158b251d8be).
2. Alternatively, we can add a new `fuse_config` option which would control this behaviour automatically.

I have a slight preference over (1), since it gives more control, but I would love to hear your suggestions.

AM5800:
Currently `cache_readdir` and `keep_cache` flags are not passed from high-level api to low-level. This means that high-level API users can not use readdir kernel caching.

This commit simply passes the above mentioned flags to low-level.

Fixes #821

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,8 @@
+libfuse next.release.version (xxxx-xx-xx)
+===========================
+
+* Readdir kernel cache can be enabled from high-level API.
+
 libfuse 3.15.1 (2023-07-05)
 ===========================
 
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks! Could you document the change in ChangeLog.rst please?

AM5800:
Done!

PR #839, opened 2023-09-15
Passthrough hp dio

bsbernd:
This adds the --direct-io option to passthrough-hp and also fixes --clone-fd

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -158,6 +158,7 @@ struct Fs {
     size_t num_threads;
     bool clone_fd;
     std::string fuse_mount_options;
+    bool direct_io;
 };
 static Fs fs{};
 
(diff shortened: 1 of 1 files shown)


example/passthrough_hp.cc
```diff
@@ -20,7 +20,8 @@
  *
  * If --nocache is specified, the source directory may be changed
  * directly even while mounted and the filesystem will continue
- * to work correctly.
+ * to have correct file attributes. For data coherency --direct-io is
```
Nikratio, line 23: Could you expand on this? It's not clear to me what's meant by that (alternatively, omitting it is also fine).

bsbernd: Hmm, what I mean is that --nocache just disables the attribute cache, but still data inconsistencies might come up.

PR #847, opened 2023-10-10
Enabled parallel direct IO writes for passthrough examples

bsbernd:
All these passthrough examples don't need writes to be serialized.

Actually, most file systems probably handle non serialized parallel direct writes - the FOPEN_PARALLEL_DIRECT_WRITES flag is just to avoid a regression for those file system that rely on serialized DIO writes in fuse kernel. Passthrough file system forward the IO to another file system, which actually handles that internally - serialized in fuser kernel is not needed.

### DIFF
--- a/example/passthrough.c
+++ b/example/passthrough.c
@@ -285,6 +285,7 @@ static int xmp_create(const char *path, mode_t mode,
 		return -errno;
 
 	fi->fh = res;
+	fi->parallel_direct_writes = 1;
 	return 0;
 }
 
@@ -297,6 +298,7 @@ static int xmp_open(const char *path, struct fuse_file_info *fi)
 		return -errno;
 
 	fi->fh = res;
+	fi->parallel_direct_writes = 1;
 	return 0;
 }
 
--- a/example/passthrough_fh.c
+++ b/example/passthrough_fh.c
@@ -366,6 +366,7 @@ static int xmp_create(const char *path, mode_t mode, struct fuse_file_info *fi)
 		return -errno;
 
 	fi->fh = fd;
+	fi->parallel_direct_writes = 1;
 	return 0;
 }
 
@@ -378,6 +379,7 @@ static int xmp_open(const char *path, struct fuse_file_info *fi)
 		return -errno;
 
 	fi->fh = fd;
+	fi->parallel_direct_writes = 1;
 	return 0;
 }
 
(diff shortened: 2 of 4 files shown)


Nikratio:
Hmm. After merging, this broke a test: https://github.com/libfuse/libfuse/actions/runs/6479444187

Flaky test or real problem?

bsbernd:
I wouldn't know how this can break a test - right now this flag only has an effect when fi->direct_io / FOPEN_DIRECT_IO is set. Once fuse dio work is merged (https://lwn.net/Articles/944881/) it will also allow parallel plain O_DIRECT. Hmm, although the error is during compilation?


What does this actually mean
```
 File "/tmp/libfuse-build-zw7t97/build-gcc/test/conftest.py", line 80, in _check
    raise AssertionError('Suspicious output to stderr (matched "%s")'
```

bsbernd:
No issue in my 22.04 chroot, though
```
bschubert2@ubuntu22.04@imesrv6 build-ubuntu>ninja clean
[2/2] Cleaning
Cleaning... 78 files.
bschubert2@ubuntu22.04@imesrv6 build-ubuntu>ninja
[73/73] Linking target example/passthrough_hp
bschubert2@ubuntu22.04@imesrv6 build-ubuntu>git show HEAD |head -n6
commit 806b16a99bf8e2fe5a7b4e6ea3cb8b62935f2a7c
Author: Bernd Schubert <bschubert@ddn.com>
Date:   Tue Oct 10 13:18:36 2023 +0200

    Enabled parallel direct IO writes for passthrough examples
    
bschubert2@ubuntu22.04@imesrv6 build-ubuntu>
```

Can you retrigger the test?

Nikratio:
> I wouldn't know how this can break a test - right now this flag only has an effect when fi->direct_io / FOPEN_DIRECT_IO is set. Once fuse dio work is merged (https://lwn.net/Articles/944881/) it will also allow parallel plain O_DIRECT. Hmm, although the error is during compilation?
> 
> What does this actually mean
> 
> ```
>  File "/tmp/libfuse-build-zw7t97/build-gcc/test/conftest.py", line 80, in _check
>     raise AssertionError('Suspicious output to stderr (matched "%s")'
> ```

It means that the binary that the Python script forked produced an error message. You can see the error message below (write size not having expected value). Will re-run the test.

Nikratio:
Seems to have passed on second try, so guess it's a flaky test.

PR #850, opened 2023-10-11
passthrough_hp: Fix clone-fd option

SteveY4ng:
The clone-fd option is set valued but not used in the context. Use it in the code.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -1357,6 +1357,8 @@ int main(int argc, char *argv[]) {
     if (fs.num_threads != -1)
         fuse_loop_cfg_set_idle_threads(loop_config, fs.num_threads);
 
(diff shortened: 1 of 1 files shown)


bsbernd (approved):
Thanks, yeah I had forgotten it that. Actually I think have a patch like this in one of the branches - looks like I forgot to submit it. Thanks for fixing it!

bsbernd:
Hmm, I wonder why it is stuck at "Waiting for status to be reported ". @Nikratio any idea?

SteveY4ng:
> Hmm, I wonder why it is stuck at "Waiting for status to be reported ". @Nikratio any idea?

Hello! I think the reason is that  the maintainers need to run some workflows on pull requests from contributors that require approval. I don't have the write access to this repository. And the workflows need to be started by the maintainers manually. 
You can see this [Link](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) for more information.

PR #852, opened 2023-10-18
fuse_clone_chan: avoid additional operations to set FD_CLOEXEC flag.

legezywzh:
Since open(2) has been called with O_CLOEXEC flag.

### DIFF
--- a/lib/fuse_loop_mt.c
+++ b/lib/fuse_loop_mt.c
@@ -259,7 +259,9 @@ static struct fuse_chan *fuse_clone_chan(struct fuse_mt *mt)
 			strerror(errno));
 		return NULL;
 	}
+#ifndef O_CLOEXEC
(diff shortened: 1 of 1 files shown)


bsbernd (commented):
The only risk is that someone might run libfuse with a really old kernel that does not have O_CLOEXEC

Nikratio:
> > The only risk is that someone might run libfuse with a really old kernel that does not have O_CLOEXEC

In that case, shouldn't the open call fail with `EINVAL`?

bsbernd:
Although we can probably assume that O_CLOEXEC is known, as FUSE_DEV_IOC_CLONE was introduced much later. The ifdef around O_CLOEXEC makes sense if compiled on a ancient userspace system that does not have these flags. I just wonder if we shouldn't add meson tests to enforce more recent systems.

PR #858, opened 2023-11-08
Remove unnecessary write_header_size check

legezywzh:
IIUC, in fuse_session_receive_buf_int(), if fuse_buf is set with FUSE_BUF_IS_FD, fuse_buf's size would be greater than "sizeof(struct fuse_in_header) + sizeof(struct fuse_write_in) + pagesize)", then write_header_size must be less than fuse_buf's size, so we can remove write_header_size check safely.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2658,9 +2658,6 @@ void fuse_session_process_buf_int(struct fuse_session *se,
 	int res;
 
 	if (buf->flags & FUSE_BUF_IS_FD) {
-		if (buf->size < tmpbuf.buf[0].size)
-			tmpbuf.buf[0].size = buf->size;
-
 		mbuf = malloc(tmpbuf.buf[0].size);
 		if (mbuf == NULL) {
 			fuse_log(FUSE_LOG_ERR, "fuse: failed to allocate header\n");
@@ -2740,7 +2737,7 @@ void fuse_session_process_buf_int(struct fuse_session *se,
 			fuse_reply_err(intr, EAGAIN);
 	}
 
-	if ((buf->flags & FUSE_BUF_IS_FD) && write_header_size < buf->size &&
+	if ((buf->flags & FUSE_BUF_IS_FD) &&
 	    (in->opcode != FUSE_WRITE || !se->op.write_buf) &&
 	    in->opcode != FUSE_NOTIFY_REPLY) {
 		void *newmbuf;


Nikratio:
> IIUC, in fuse_session_receive_buf_int(), if fuse_buf is set with FUSE_BUF_IS_FD, fuse_buf's size would be greater than "sizeof(struct fuse_in_header) + sizeof(struct fuse_write_in) + pagesize)", 

How do you know that? `fuse_buf` is passed into `fuse_session_receive_buf_int` as a parameter, so it could be pretty much anything , no?

legezywzh:
> > IIUC, in fuse_session_receive_buf_int(), if fuse_buf is set with FUSE_BUF_IS_FD, fuse_buf's size would be greater than "sizeof(struct fuse_in_header) + sizeof(struct fuse_write_in) + pagesize)",
> 
> How do you know that? `fuse_buf` is passed into `fuse_session_receive_buf_int` as a parameter, so it could be pretty much anything , no?

I may miss something. Look at codes in fuse_session_receive_buf_int():
```
tmpbuf = (struct fuse_buf) {
                .size = res,
                .flags = FUSE_BUF_IS_FD,
                .fd = llp->pipe[0],
        }; 

        if (res < sizeof(struct fuse_in_header) +
            sizeof(struct fuse_write_in) + pagesize) {
                struct fuse_bufvec src = { .buf[0] = tmpbuf, .count = 1 };
                struct fuse_bufvec dst = { .count = 1 };

                if (!buf->mem) {
                        buf->mem = malloc(se->bufsize);
                        if (!buf->mem) {
                                fuse_log(FUSE_LOG_ERR,
                                        "fuse: failed to allocate read buffer\n");
                                return -ENOMEM; 
                        } 
                }
                buf->size = se->bufsize;
                buf->flags = 0;
                dst.buf[0] = *buf;

                res = fuse_buf_copy(&dst, &src, 0);
                if (res < 0) {
                        fuse_log(FUSE_LOG_ERR, "fuse: copy from pipe: %s\n",
                                strerror(-res));
                        fuse_ll_clear_pipe(se);
                        return res;
                }
                if (res < tmpbuf.size) {
                        fuse_log(FUSE_LOG_ERR, "fuse: copy from pipe: short read\n");
                        fuse_ll_clear_pipe(se);
                        return -EIO;
                }
                assert(res == tmpbuf.size);

        } else {
                /* Don't overwrite buf->mem, as that would cause a leak */
                buf->fd = tmpbuf.fd;
                buf->flags = tmpbuf.flags;
        }
        buf->size = tmpbuf.size;
```

Seems if FUSE_BUF_IS_FD has been set, fuse_buf's size would come frome 'res' or tmpbuf's size , but res is greater than " sizeof(struct fuse_in_header) + sizeof(struct fuse_write_in) + pagesize) "

lib/fuse_lowlevel.c
```diff
@@ -2658,9 +2658,6 @@ void fuse_session_process_buf_int(struct fuse_session *se,
 	int res;
 
 	if (buf->flags & FUSE_BUF_IS_FD) {
-		if (buf->size < tmpbuf.buf[0].size)
```
Nikratio, line 2661: How do you know this condition will never be fulfilled? `buf->size` depends on the caller, and `tmpbuf[0].size` has size `write_header_size` (which is independent of `buf->size`).

Nikratio:
Sorry, I don't understand how your reply relates to my question. I've tried to rephrase it and attach it to the specific line of code it relates to, maybe that helps.

lib/fuse_lowlevel.c
legezywzh: > How do you know this condition will never be fulfilled? `buf->size` depends on the caller, and `tmpbuf[0].size` has size `write_header_size` (which is independent of `buf->size`).

I had thought buf->size would only be assigned by fuse_session_receive_buf_int, which is the only caller,  then if I understand correctly, buf->size will be greater than "sizeof(struct fuse_in_header) + sizeof(struct fuse_write_in) + pagesize" if  FUSE_BUF_IS_FD is flagged.

legezywzh:
> Sorry, I don't understand how your reply relates to my question. I've tried to rephrase it and attach it to the specific line of code it relates to, maybe that helps.

Really never mind, thanks for your kindness.

lib/fuse_lowlevel.c
Nikratio: This might well be true, yes. However, making such assumptions is not a good idea. This is because if someone changes the code of `fuse_session_receive_buf_int` in the future, they are unlikely to be aware that `fuse_session_process_buf_int` cannot receive an arbitrary `buf` parameter. The same goes if someone wants to call `fuse_session_process_buf_int` in a new place.

Nikratio:
No worries. Thanks for submitting the pull request. I hope you understand why I will rather not make change this change.

PR #862, opened 2023-11-19
Clarify fuse_lowlevel poll() docstring

thirtythreeforty:
The original docstring was confusing; it was not clear that the ph must
be retained indefinitely, nor was it clear that the client *also* needs
to reply to the poll call immediately.

Clarify this by explaining that it is only necessary to retain a single
handle, that the client must retain ph, and that it must immediately
call reply.


Nikratio:
Thanks!

PR #870, opened 2023-12-04
F_OPEN_DIRECT_IO allow mmap

bsbernd:


bsbernd:
@amir73il I guess that is what you also have? @szmi I'm not entire sure what to do about the the pending flag rename.

szmi:
See git://git.kernel.org/pub/scm/linux/kernel/git/mszeredi/fuse.git#for-next

amir73il:
> @amir73il I guess that is what you also have? @szmi I'm not entire sure what to do about the the pending flag rename.

I just had a local patch in my build to forcefully add the init flag

Nikratio:
Is this ready for merging and supported by a released kernel?

bsbernd:
@Nikratio in linux-6.7 the flag is called FUSE_DIRECT_IO_RELAX (but just because Miklos didn't have to send a pull request). fuse-next branch already has the rename (and given the flag was already released, also has a compat define). I already added in that rename here.

bsbernd:
Looks like there is a generic issue with this test
```
ERROR: fuse_lowlevel_notify_store() failed with No such file or directory (2)
```

Going to look into it later.

bsbernd:
@Nikratio I created a different pull request for the notification test failure, I think easier to review and merge with separate pull requests.

bsbernd:
@Nikratio any reason to hold this back? linux-6.7 is released now, which also includes commit c55e0a55b165202f18cbc4a20650d2e1becd5507

```
commit c55e0a55b165202f18cbc4a20650d2e1becd5507
Author: Tyler Fanelli <tfanelli@redhat.com>
Date:   Tue Sep 19 22:40:00 2023 -0400

    fuse: Rename DIRECT_IO_RELAX to DIRECT_IO_ALLOW_MMAP
    
    Although DIRECT_IO_RELAX's initial usage is to allow shared mmap, its
    description indicates a purpose of reducing memory footprint. This
    may imply that it could be further used to relax other DIRECT_IO
    operations in the future.
    
    Replace it with a flag DIRECT_IO_ALLOW_MMAP which does only one thing,
    allow shared mmap of DIRECT_IO files while still bypassing the cache
    on regular reads and writes.
    
    [Miklos] Also Keep DIRECT_IO_RELAX definition for backward compatibility.    
```

Any reasons not to merge this pull request?

Nikratio:
Apologies. No reason other than lack of time for libfuse on my part. Unfortunately this lack of time also means that when I finally  find some time I'm increasingly lacking context, so the reviews take even longer because I need to refresh my mind. I'll take your word for this one.

PR #880, opened 2024-01-10
Don't set FUSE_CAP_PARALLEL_DIROPS by default

szmi:
Allowing parallel dir operations could result in a crash in a filesystem
implementation that is not prepared for this.

To be safe keep this flag off by default (this is not a regression, since
there was no public release where this flag wasn't ignored).

If the filesystem wants better perforance, then it should set this flag
explicitly.

Fixes: c9905341ea34 ("Pass FUSE_PARALLEL_DIROPS to kernel (#861)")

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -336,8 +336,6 @@ struct fuse_loop_config_v1 {
  * is unset, the FUSE kernel module will ensure that lookup() and
  * readdir() requests are never issued concurrently for the same
  * directory.
- *
- * This feature is enabled by default when supported by the kernel.
  */
 #define FUSE_CAP_PARALLEL_DIROPS        (1 << 18)
 
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2053,7 +2053,6 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 	if ((cond) && (se->conn.capable & (cap))) \
 		se->conn.want |= (cap)
 	LL_SET_DEFAULT(1, FUSE_CAP_ASYNC_READ);
-	LL_SET_DEFAULT(1, FUSE_CAP_PARALLEL_DIROPS);
 	LL_SET_DEFAULT(1, FUSE_CAP_AUTO_INVAL_DATA);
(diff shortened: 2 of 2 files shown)


bsbernd:
@szmi Could we also updated the comment to something like this? From the existing comment I would have guessed that a read-write lock is used (like readdir takes write- and lookup takes read- lock).

```
diff --git a/include/fuse_common.h b/include/fuse_common.h
index 3ff14ec9..c24e24ee 100644
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -335,9 +335,8 @@ struct fuse_loop_config_v1 {
  * Indicates support for parallel directory operations. If this flag
  * is unset, the FUSE kernel module will ensure that lookup() and
  * readdir() requests are never issued concurrently for the same
- * directory.
- *
- * This feature is enabled by default when supported by the kernel.
+ * directory. This is a kernel directory inode mutex lock - parallel
+ * lookup requests to the same directory are also serialized.
  */
 #define FUSE_CAP_PARALLEL_DIROPS        (1 << 18)
``

**Comment by bsbernd** [comment]
Hi @szmi,

yeah I meant that serialization of lookup vs lookup is not clear. If I would have read that comment without reading kernel code, I would have assumed lookup is using a read lock and readdir would use a write lock.

**Comment by szmi** [comment]
Hmm, but with that logic using write lock for lookup and read lock for readdir is equally plausible.  How about

```
/**
 * Indicates support for parallel directory operations. If this flag
 * is unset, the FUSE kernel module will ensure that all lookup() and
 * readdir() requests are serialized on a particular directory.
 */
#define FUSE_CAP_PARALLEL_DIROPS        (1 << 18)

```

bsbernd:
Sounds good to me!

szmi:
Test is failing for a reason beyond this PR (passed on v1 and failed on v2 with the only difference being in the comment).   I don't seem to find a way to trigger a re-run of the test.

@Nikratio any ideas?

bsbernd:
Arg, I thought I had fixed that by introducing a clean shutdown in that test. All my attempts to reproduce locally never worked and for my patch multiple github runs had succeeded. I will try again to fix it. @szmi  closing/re-opening the pull request will retrigger, but not sure if this is really needed here.

PR #886, opened 2024-01-29
Add extra check for high bit inargflags

HereThereBeDragons:
Based on this comment here by @bsbernd  https://github.com/libfuse/libfuse/pull/871#discussion_r1469229721
I added checks for the 2 inargflags that are in the high bits ( > 32)

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2019,9 +2019,11 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 				bufsize = max_bufsize;
 			}
 		}
-		if (inargflags & FUSE_DIRECT_IO_ALLOW_MMAP)
+		if ((inargflags & FUSE_INIT_EXT) &&
+				(inargflags & FUSE_DIRECT_IO_ALLOW_MMAP))
 			se->conn.capable |= FUSE_CAP_DIRECT_IO_ALLOW_MMAP;
-		if (arg->minor >= 38 || (inargflags & FUSE_HAS_EXPIRE_ONLY))
+		if (arg->minor >= 38 ||
+				((inargflags & FUSE_INIT_EXT) && (inargflags & FUSE_HAS_EXPIRE_ONLY)))
 			se->conn.capable |= FUSE_CAP_EXPIRE_ONLY;
 	} else {
 		se->conn.max_readahead = 0;


bsbernd:
I don't understand this. The high bits of inargflags are 0 as the flag is initialized as 0, unless FUSE_INIT_EXT is set - this extra check is not needed... Nor is checking for the version number.

HereThereBeDragons:
ok now i am a bit confused: first you write this here `FUSE_HAS_EXPIRE_ONLY is (1ULL << 35>), so is in the high bits, which is only set when FUSE_INIT_EXT is also set. Otherwise we couldn't trust these flags at all.` and now you say it is not needed.

and the check for version is needed because the first kernel support of this feature did not have a flag to check. It only was later introduced to check for the flag (I guess when we asked for a backport from RedHat). So this is why "if version is high enough OR flag is found".

I will close this now as you are right  in `do_init()`  `inargflags` is set to 0 before values are assigned

bsbernd:
What I meant is to not trust the version, i.e. to remove the check for the version. I'm going to submit a separate patch for that in a bit and are going to request your review.

HereThereBeDragons:
ah ok. i guess i interpreted your comment wrongly.

in your PR please also consider that the feature should also be support for kernels between those 2 commits:
https://github.com/torvalds/linux/commit/4f8d37020e1fd0bf6ee9381ba918135ef3712efd (kernel 6.2 rc1 - introduction of the feature)
https://github.com/torvalds/linux/commit/5cadfbd5a11e5495cac217534c5f788168b1afd7 (kernel 6.5 rc3 - introduction of the flag `FUSE_HAS_EXPIRE_ONLY`)

bsbernd:
Hmm, that is ugly https://github.com/torvalds/linux/commit/5cadfbd5a11e5495cac217534c5f788168b1afd7 got picked up in stable kernels in 6.4, but didn't make it into 6.2 and 6.3. I'm going to send a notification for ubuntu-hwe. My personal opinion is that https://github.com/torvalds/linux/commit/4f8d37020e1fd0bf6ee9381ba918135ef3712efd is buggy and cannot be trusted without the fix.

bsbernd:
I think the issue was that there is a merge conflict (rather simple) and then linux-stable didn't pick it up automatically. In the mean time these kernels are not maintained anymore. @HereThereBeDragons Do you have distributions/kernels in your mind that use 6.2 and 6.3 and need this flag? I would really like to avoid having a version based check in libfuse that doesn't work with fuse back ports.

HereThereBeDragons:
We had a short discussion in our team. We do not mind if the check is reduced to only on the capability flag and not also, as other option, the header version - even if this excludes some OS versions for a bit longer. We are mainly interested in RHEL and for RHEL8 and RHEL9 i just confirmed the backports for this `expire`-feature.

Out of personal curiosity is there anything "Best Practice Guide" or so to read about why checking for the header version for some available functionality is a bad thing? (For backport i understand that you need the capability flag)

bsbernd:
Another example regarding version check, let's say in a later kernel version one wants to disable that feature at all (let's assume something like a design issue) and one needs a totally new struct. So kernel would introduce something like FUSE_HAS_EXPIRE_ONLY_V2 and would not set FUSE_HAS_EXPIRE_ONLY anymore. With version check just not possible to ever disable the old feature.

Going to check RHEL8 and RHEL9 kernel code later on.  I have filled this ubuntu bug report:
https://bugs.launchpad.net/ubuntu/+source/linux-hwe-6.2/+bug/2051700

bsbernd:
Ah also forgot to write, my idea is to add an env variable here that would allow a version check for this specific feature.

HereThereBeDragons:
ah true, that makes sense that there a version check might not be wise long term.

thanks for taking care of writing to linux-hwe and sounds good. looking forward to your pr!

PR #892, opened 2024-02-19
Retry tests/test_write_cache in write back mode

bsbernd:
This test can fail whenever a something (kernel, userspace) decides to flush in between of two 2048B writes. These two writes are supposed to be merged into a single 4906 byte write by the kernel writeback cache, but _sometimes_ the test fails because 2048 byte writes get through.

### DIFF
--- a/test/test_write_cache.c
+++ b/test/test_write_cache.c
@@ -21,6 +21,7 @@
 #include <unistd.h>
 #include <sys/stat.h>
 #include <pthread.h>
+#include <stdatomic.h>
 
 #ifndef __linux__
 #include <limits.h>
@@ -42,6 +43,8 @@ struct options {
     .delay_ms = 0,
 };
 
+#define WRITE_SYSCALLS 64
+
 #define OPTION(t, p)                           \
     { t, offsetof(struct options, p), 1 }
 static const struct fuse_opt option_spec[] = {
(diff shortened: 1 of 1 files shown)


test/test_write_cache.c
```diff
@@ -51,6 +54,7 @@ static const struct fuse_opt option_spec[] = {
     FUSE_OPT_END
 };
 static int got_write;
+static atomic_int write_cnt;
```
ashleypittman, line 57: Is this used anywhere?

ashleypittman (commented):
Does the commit message need updating?

I've been testing this as well, the commit message talks about using two writes to check coalescing, using a greater number of writes has a higher chance of success.

test/test_write_cache.c
bsbernd: It has to - is a core requirement of this patch

```
assert(write_cnt < WRITE_SYSCALLS);
```

bsbernd: Complete

```
    if (options.writeback)
        assert(write_cnt < WRITE_SYSCALLS);
    else
        assert(write_cnt == WRITE_SYSCALLS);
``

ashleypittman: Ah yes, that makes much more sense, I'm not sure how I missed that.

ashleypittman (approved):
I was looking at the PR comment.

PR #893, opened 2024-02-19
remove duplicated fuse_chan_put() 

yangyun50:
remove duplicated fuse_chan_put() in do_interrupt().
In the pull request [#649 ](https://github.com/libfuse/libfuse/pull/649), it invokes fuse_chan_put(req->ch) twice if curr->ctr is zero, one is in the find_interrupted()(line 1717), another is in the do_interrupt()(line 1747). The fuse_chan_put() in the find_interrupted() is not needed.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -1740,8 +1740,6 @@ static int find_interrupted(struct fuse_session *se, struct fuse_req *req)
 			pthread_mutex_lock(&se->lock);
 			curr->ctr--;
 			if (!curr->ctr) {
-				fuse_chan_put(req->ch);
-				req->ch = NULL;
 				destroy_req(curr);
 			}
 


lib/fuse_lowlevel.c
```diff
@@ -1740,8 +1740,6 @@ static int find_interrupted(struct fuse_session *se, struct fuse_req *req)
 			pthread_mutex_lock(&se->lock);
 			curr->ctr--;
 			if (!curr->ctr) {
-				fuse_chan_put(req->ch);
-				req->ch = NULL;
```
bsbernd, line 1744: To be honest, I don't understand why this acts on req->ch and not on `cur->ch`. `req` is interrupting `curr` and also destroys it. How will it ever put the reference of it now? So either I miss something or current code is wrong and the patch then goes into the wrong direction.

bsbernd (changes requested):
Is this trying to fix a real live issue or just something spotted on code review or by automatic tools?

lib/fuse_lowlevel.c
yangyun50: I think fuse_chan_put does not need in this position,  whatever the argument is `req->ch` or `cur->ch`.  
1. The reason for `req->ch` is described in the [first comment](https://github.com/libfuse/libfuse/pull/893#issue-2142610769) of this pull request.  As what you said, `req` is interrupting `curr`. When `req` found the request to be interrupted (i.e., `curr`), it  marks `curr->interrupted` and executes some interrupt operations. And then the life of `req` is over and `req` would be destroyed in the following position, not in this position. https://github.com/libfuse/libfuse/blob/67d28fb4b762989d895732f8b1aec9a4f69c0d56/lib/fuse_lowlevel.c#L1771-L1776
2. Ths reason for `cur->ch` is described below. 
For every `request`, it invokes `fuse_chan_get()`  only once in the code below.
https://github.com/libfuse/libfuse/blob/67d28fb4b762989d895732f8b1aec9a4f69c0d56/lib/fuse_lowlevel.c#L2713 

For every `request` which is FUSE_INTERRUPT request, it invokes `fuse_chan_put()` either in the `do_interrupt` function,
https://github.com/libfuse/libfuse/blob/67d28fb4b762989d895732f8b1aec9a4f69c0d56/lib/fuse_lowlevel.c#L131-L146
or in the `check_interrupt` function.
https://github.com/libfuse/libfuse/blob/67d28fb4b762989d895732f8b1aec9a4f69c0d56/lib/fuse_lowlevel.c#L1781-L1804

And for every `request` which is not FUSE_INTERRUPT request, it invokes `fuse_chan_put()` only once in the function `fuse_free_req()`. 
https://github.com/libfuse/libfuse/blob/67d28fb4b762989d895732f8b1aec9a4f69c0d56/lib/fuse_lowlevel.c#L131-L146

If `fuse_chan_put()` is invoked in this position for `curr`, it would be invoked twice for `curr`, one is in this position and another is in `fuse_free_req()`.

Sorry for my expression if you are confused.

yangyun50:
Thanks for your advice!
Recently I'm focused on fuse interrupt mechanism, and this is what I found in the code.  And this situation is difficult to reproduce since its complicated reproduce situation (e.g., interrupt, lock contention).

lib/fuse_lowlevel.c
bsbernd: >If fuse_chan_put() is invoked in this position for curr, it would be invoked >twice for curr, one is in this position and another is in fuse_free_req().

Please take a closer look for this part.

```
			if (!curr->ctr) {
				fuse_chan_put(req->ch);
				req->ch = NULL;
				destroy_req(curr);
```

--> This calls `destroy_req(curr)` - if there would be another call of fuse_free_req() for the same req, it would be use-after-free.

bsbernd (changes requested):

lib/fuse_lowlevel.c
yangyun50: These two `destroy_req()` can only be called once for the same req. Because only if the `req->ctr` (i.e., refcount) is zero, it will call `destroy_req()` .
In the code
https://github.com/libfuse/libfuse/blob/67d28fb4b762989d895732f8b1aec9a4f69c0d56/lib/fuse_lowlevel.c#L1742-L1746

this part  has the condition `if(!curr->ctr)`

and for `fuse_free_req()`
https://github.com/libfuse/libfuse/blob/67d28fb4b762989d895732f8b1aec9a4f69c0d56/lib/fuse_lowlevel.c#L131-L146

it has the condition `if (!ctr)`.

however `fuse_chan_put` cannot be promised only invoked once.

bsbernd: Well look

```
			curr->ctr--;
			if (!curr->ctr) {
				fuse_chan_put(req->ch);
				req->ch = NULL;
				destroy_req(curr);
			}
```

It decreases `curr->ctr` and when zero destroys the request. Whatever `fuse_free_req` would do here is not relevant, because if  `fuse_free_req`  would called for this request we would have a use-after-free.
And neither can one call `fuse_chan_put()` on freed request, which clearly point out a reference leak.
Obviously `fuse_chan_put(req->ch)` is wrong, but this should be `fuse_chan_put(curr->ch)`

bsbernd (approved):

bsbernd:
@yangyun50 Thank you for your contribution and patience, I first had to understand the code.

PR #898, opened 2024-03-05
fusermount: Fix head-buffer-overflow in extract_x_options

bsbernd:
Commit 74b1df2e introduced a heap-buffer-overflow, as allocated memory was not initialized and extract_x_options was also not checking for the remaining buffer size. Fix is to initialize the buffer and to also not exceed the buffer size. Actually not exceeding buffer size is rather complex with C and introduced quite some code changes.

Also fixed is a memory leak of allocated buffers in the commit mentioned above.

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -24,6 +24,7 @@
 #include <mntent.h>
 #include <sys/wait.h>
 #include <sys/stat.h>
+#include <sys/param.h>
 
 #include "fuse_mount_compat.h"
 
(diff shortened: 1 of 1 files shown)


MatthewCash (approved):
Everything looks good, the x-options are passed through correctly. 

I'm struggling to see where a buffer overflow could occur since extract_x_options never copies more than the length of the original options?

bsbernd:
Thanks for looking at it. The issue was here:

`if (strlen(opt_buf) > 1)` in `extract_x_options()`, opt_buf was set by

`		opt_buf = strncmp(opt, "x-", 2) ? non_x_opts : x_opts;`

so the overflow happened with non_x_opts / x_opts. And with these come from

```
do_mount_opts = malloc(opts_size);
x_opts = malloc(opts_size);
```

Issue is that these were not zeroed - strlen() did not terminate. I switched to calloc() to zero it and also added additional buf len checks. The simple fix was just using calloc, but I _think_ the updated code is easier to follow and with more checks. Although in general, I start to wonder if fusermount should be switched to another language - parsing string in C is hard to get safe...

MatthewCash:
Oh wow, I don't think I would have ever caught that and I'm seriously impressed that you found it. I guess in my testing I got lucky that the buffer was already zeroed out (in hindsight this was probably unlucky). The extra checks probably aren't technically needed but I'm all for them since this function is proving itself to be more dangerous that I imagined.

bsbernd:
It was ASAN that found it (see here [asan report]( https://github.com/libfuse/libfuse/pull/894#issuecomment-1979234391))  and with its output it was actually easy to see which variable had the issue. The tests here already run with ASAN and should have found it, if the compiled fusermount would have been used. Will look into this tomorrow (I need it anyway for posix_spawn).
The not so nice part is that fusermount is typicall installed with the s-bit set - potential security issue. So we better do all string paranoia we can do.

MatthewCash:
That makes sense, certainly glad it was caught at all, its unfortunate that the tests weren't working before merge. I definitely agree that a memory-safe language would be worthwhile to explore for fusermount.

PR #901, opened 2024-03-05
Switch from fork to posix_spawn

bsbernd:
Client code might allocate a lot of memory before starting the mount.
Fork is slow for processes that are using a lot of memory.  But
posix_spawn fixes that.

Another issue with fork is if the process is also doing RDMA - this
might lead to data corruption, as least if memory used for RDMA
is not marked with MADV_DONTFORK.  At least with linux kernels
before 5.12.
Also see https://blog.nelhage.com/post/a-cursed-bug/ for more details

### DIFF
--- a/test/ci-build.sh
+++ b/test/ci-build.sh
@@ -18,6 +18,9 @@ umask 0022
 # readable/executable.
 SOURCE_DIR="$(readlink -f .)"
 TEST_DIR="$(mktemp -dt libfuse-build-XXXXXX)"
+
+PREFIX_DIR="$(mktemp -dt libfuse-install-XXXXXXX)"
+
 chmod 0755 "${TEST_DIR}"
 cd "${TEST_DIR}"
 echo "Running in ${TEST_DIR}"
@@ -47,13 +50,17 @@ for CC in gcc gcc-9 gcc-10 clang; do
     else
         build_opts=''
     fi
-    meson setup -D werror=true ${build_opts} "${SOURCE_DIR}" || (cat meson-logs/meson-log.txt; false)
+    meson setup -Dprefix=${PREFIX_DIR} -D werror=true ${build_opts} "${SOURCE_DIR}" || (cat meson-logs/meson-log.txt; false)
     ninja
+    sudo ninja install
 
-    sudo chown root:root util/fusermount3
-    sudo chmod 4755 util/fusermount3
+    # libfuse will first try the install path and then system defaults
+    sudo chmod 4755 ${PREFIX_DIR}/bin/fusermount3
+
     ${TEST_CMD}
     popd
+    rm -fr build-${CC}
+    sudo rm -fr ${PREFIX_DIR}
 done
 
 sanitized_build()
@@ -62,7 +69,7 @@ sanitized_build()
 
     mkdir build-san; pushd build-san
 
-    meson setup -D werror=true\
+    meson setup -Dprefix=${PREFIX_DIR} -D werror=true\
            "${SOURCE_DIR}" \
            || (ct meson-logs/meson-log.txt; false)
     meson configure $SAN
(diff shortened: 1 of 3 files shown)


test/ci-build.sh
```diff
@@ -26,7 +30,7 @@ export CC
 # Standard build
 for CC in gcc gcc-9 gcc-10 clang; do
     echo "=== Building with ${CC} ==="
-    mkdir build-${CC}; cd build-${CC}
+    mkdir build-${CC}; pushd build-${CC}
```
matthiasgoergens, line 33: You could probably also use a sub shell for the loop body, instead of manually keeping pushd and popd in sync?

(Oh, but I see that we are using pushd and popd already elsewhere in the script. So probably better not to change that in this PR.)

matthiasgoergens:
You might want to describe in the PR description _why_ we are doing this?

bsbernd:
It is all in the commit messages,  I update the PR description later on.

test/ci-build.sh
bsbernd: All individual commits - just should be explained in the commit message.

bsbernd:
@matthiasgoergens seems to pass. I can also move the ci-updates into its own PR, if you prefer. The installation of fusermount3 is a dependency for this commit, as fusermount3 gets a new parameter - old binary will not work anymore (which gives me a bit headache for some other reasons).

matthiasgoergens:
If we merge PRs with keeping individual commits intact (ie no squashing) then moving into its own PR wouldn't really bring any benefit.

I've only looked over the code very briefly. But I can look a bit more in depth, if you want a proper review from me?

bsbernd:
I definitely want to keep the individual commits, all of these have a commit message and explain what they are for. Yes, please review - first commit is yours - oops, I just see that the author changed - changing it back to you now.

ashleypittman:
> Also see https://blog.nelhage.com/post/a-cursed-bug/ for more details

That's a great read.

I don't think we can change the interface to fusermount so readily if I'm understanding the commit message correctly, I know that we often build just the library rather than the binary as end users and then rely on the installed package for the mounting, and i know that for a while we were even compiling fuse3 and then using the fuse2 fusermount program to do the mounting so medium term or backwards compatibility if the binary is essential IMO.

bsbernd:
Actually, I think we do not need to copy env. We can just do keep (or re-add) the existing setenv and libc does actually all the work for us.

https://github.com/lattera/glibc/blob/master/stdlib/setenv.c

The function `__add_to_environ()` does all we need and we can just pass the environ pointer.

bsbernd:
@ashleypittman Please check again. It was always first using the install location and then system default - no change in behavior. It now *adds* the new parameter to fusermount, but does not use it. Instead it sets the env variable - as before and just passes through 'environ'. Zero change in behavior regarding fusermount - I actually test if an old fusermount still works.

bsbernd:
I.e. see new commit 3 in the series: https://github.com/libfuse/libfuse/pull/901/commits/81d55f99267996f88629ba199d868075f333c0a6

bsbernd:
Yeah, I want to keep them as separate commits. But that doesn't hinder us to fix patch 1 in patch3 - this is supposed to get merged in one step. At some later time in the future we can then just revert patch3.

bsbernd:
I could also squash patch3 into patch1, I don't have a strong preference here.

ashleypittman (commented):
I hadn't checked the whole PR, just one commit.

ashleypittman (approved):

ashleypittman:
> I could also squash patch3 into patch1, I don't have a strong preference here.

as long as they go in together I don't think it makes a big difference.  I can see that someone might want to force the option and therefore having it as a commit that they could revert would be convenient but I'd probably squash patch3 into patch1 purely to have a cleaner and shorter commit history and clearer commit message.

bsbernd:
Fair enough, squashed into patch1 and commit message updated. Thanks for your review @ashleypittman

PR #902, opened 2024-03-13
Replace fork by posix_spawn in mount_util.c

bsbernd:
This completes commit bb9cecbf67341e03e53dd36b8f69520f3a26f834
("Use posix_spawn instead of fork+exec") and replaces remaining
fork by posix_spawn.

Not modified is mount_bsd.d as I cannot test it.

Another commit adds the env FUSE_NO_MTAB_UPDATE to be able to
disable mtab updates without recompilation.

### DIFF
--- a/lib/mount_util.c
+++ b/lib/mount_util.c
@@ -295,9 +295,8 @@ static int add_mount(const char *progname, const char *fsname,
 	if (res == -1)
 		fprintf(stderr, "%s: waitpid of %d: %s\n", progname,
 			pid, strerror(errno));
-
-	if (status != 0)
-		res = -1;
+	else
+		res = (WIFEXITED(status) && WEXITSTATUS(status) == 0) ? 0 : -1;
 
 out_restore:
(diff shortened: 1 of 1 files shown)


lib/mount_util.c
```diff
@@ -81,13 +91,45 @@ static int mtab_needs_update(const char *mnt)
...
+static int fuse_mount_posix_spawn_attr(posix_spawnattr_t *attr, const char *cmd,
+				       sigset_t mask)
+{
+	int res;
+
+	posix_spawnattr_init(attr);
+
+	int flags = POSIX_SPAWN_RESETIDS | POSIX_SPAWN_SETSIGMASK;
+	res = posix_spawnattr_setflags(attr, flags);
+	if (res) {
```
trapexit, line 103: Why the difference between this check and the one on line 110? From the man pages both appear to return 0 on success and an errno on error.

bsbernd: Thanks,  the check below for (res < 0) needs to be fixed. Although I see it in line 103

Issue #629, opened 2021-10-31
Example passthrough, fuse_fill_dir_flags and default value in cpp/c


latot:
Hi hi, I was trying to start a project in c++ using the passthrough example, and I notice something very curious, I'll put the different sections of code for context:

From fuse.h
```
enum fuse_fill_dir_flags {
	FUSE_FILL_DIR_PLUS = (1 << 1)
};
...
typedef int (*fuse_fill_dir_t) (void *buf, const char *name,
				const struct stat *stbuf, off_t off,
				enum fuse_fill_dir_flags flags);
```

From passthrough.c (only lines of interest)
```
static int fill_dir_plus = 0;
....
		if (!strcmp(argv[i], "--plus")) {
			fill_dir_plus = FUSE_FILL_DIR_PLUS;
		} else {
```

Now, this type of line is used very common:

```
fuse_fill_dir_t filler (declared in function)
(filler(buf, de->d_name, &st, 0, fill_dir_plus))
```

Is mixing types everywhere the example, ```fill_dir_plus``` var, should have the type ```fuse_fill_dir_flags``` not int, but at the same time, the enum ```fuse_fill_dir_flags``` only have one value, so we can't run the first line of the example where is set to 0.

I still don't know very well how fuse works internally, but I think ```fuse_fill_dir_flags``` should have one more flag for the 0 value.

Thx.

Nikratio:
In C, `enum` types are `int`. This is why "mixing" types this way is fine. Still, if there is a way to change this so that it also compiles without warnings in C++, then I'm happy to accept pull requests.

I do not understand what you mean with "use_fill_dir_flags should have one more flag for the 0 value" - could you elaborate?

latot:
Hi, enum estore values as int, but declaring ```enum fuse_fill_dir_flags flags```, even parsing an int can fail, depending of the compiler and the project (I test this with a C++ project), the passthrough example don't store a ```fuse_fill_dir_flags flags```, store ```int``.

The ```user_full_dir_flags``` only have one definition, ```FUSE_FILL_DIR_PLUS```, with the value 2 (1<<1), in the passthrough example the value is set to 0, so I can't change the integer type to enum.

So, to pull a request, for the ```fuse_fill_dir_flags flags```, what would means the 0 value? maybe ```FUSE_READDIR_NOT_PLUS```?

Nikratio:
Let's use `FUSE_READDIR_DEFAULTS`.

Nikratio:
[...]

PR #903, opened 2024-03-14
1. fix enum fuse_readdir_flags and enum fuse_fill_dir_flags in cpp

FredyVia:
ref #629 , add this simple enum item to adapt cpp


bsbernd (changes requested):
Could you please comment what this is for? This is an unused flag now.

bsbernd:
Could you please add failing compiler output to the commit message? Also would help if this flag would get used by the project.
The current way your commit message is very confusing, the title says 
"1. fix enum fuse_readdir_flags&fuse_fill_dir_flags in cpp"

What does "1." mean - what would "2.", "3.", etc?
And without explaining what is fixed and without an actual user of the change, this part is not right either "fix enum". I.e. it is less the code, but more the way you are communicating the change that needs updated.

FredyVia:
```
fredyvia@LittleBee ~$ gcc -Wall ~/libfuse/example/hello.c `pkg-config fuse3 --cflags --libs` -o hello // gcc compile ok!
fredyvia@LittleBee ~$ g++ -Wall ~/libfuse/example/hello.c `pkg-config fuse3 --cflags --libs` -o hello // g++ error bellow
/home/fredyvia/libfuse/example/hello.c: In function ‘int hello_readdir(const char*, void*, fuse_fill_dir_t, off_t, fuse_file_info*, fuse_readdir_flags)’:
/home/fredyvia/libfuse/example/hello.c:94:35: error: invalid conversion from ‘int’ to ‘fuse_fill_dir_flags’ [-fpermissive]
   94 |         filler(buf, ".", NULL, 0, 0);
      |                                   ^
      |                                   |
      |                                   int
/home/fredyvia/libfuse/example/hello.c:95:36: error: invalid conversion from ‘int’ to ‘fuse_fill_dir_flags’ [-fpermissive]
   95 |         filler(buf, "..", NULL, 0, 0);
      |                                    ^
      |                                    |
      |                                    int
/home/fredyvia/libfuse/example/hello.c:96:48: error: invalid conversion from ‘int’ to ‘fuse_fill_dir_flags’ [-fpermissive]
   96 |         filler(buf, options.filename, NULL, 0, 0);
      |                                                ^
      |                                                |
      |                                                int
/home/fredyvia/libfuse/example/hello.c: In function ‘int hello_read(const char*, char*, size_t, off_t, fuse_file_info*)’:
/home/fredyvia/libfuse/example/hello.c:121:20: warning: comparison of integer expressions of different signedness: ‘off_t’ {aka ‘long int’} and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
  121 |         if (offset < len) {
      |             ~~~~~~~^~~~~
/home/fredyvia/libfuse/example/hello.c: At global scope:
/home/fredyvia/libfuse/example/hello.c:137:1: error: designator order for field ‘fuse_operations::getattr’ does not match declaration order in ‘const fuse_operations’
  137 | };
```
We only need to add an enumeration item FUSE_FILL_DIR_DEFAULTS. If we do not add this enum item representing 0, we cannot use the filler function directly in the C++ program.

bsbernd:
So then please add that part to the commit message and please use the new enum value in example/hello.c

bsbernd:
Thanks, much better, please remove all debug printf. Could you please also change the commit message to something like 

```
Add FUSE_FILL_DIR_DEFAULTS enum

In order to use the fuse_fill_dir_t function in a C++ program, add the enum item: 
FUSE_FILL_DIR_DEFAULTS

Without this change g++ compilation failed with
fredyvia@LittleBee ~$ gcc -Wall ~/libfuse/example/hello.c `pkg-config fuse3 --cflags --libs` -o hello // gcc compile ok!
fredyvia@LittleBee ~$ g++ -Wall ~/libfuse/example/hello.c `pkg-config fuse3 --cflags --libs` -o hello // g++ error bellow
/home/fredyvia/libfuse/example/hello.c: In function ‘int hello_readdir(const char*, void*, fuse_fill_dir_t, off_t, fuse_file_info*, fuse_readdir_flags)’:
/home/fredyvia/libfuse/example/hello.c:94:35: error: invalid conversion from ‘int’ to ‘fuse_fill_dir_flags’ [-fpermissive]
   94 |         filler(buf, ".", NULL, 0, 0);
      |                                   ^
      |                                   |
      |                                   int


```

Please feel free to change the text, but please limit subject line to 69 characters. Commit message should be around max 79 chars, except of pasted compiler output.

bsbernd:
Sorry for the delay, there is an out-of-the-sudden test failure - it was working the days/commits before and now reproducibly stalls. Tested old commits - same issue. 
The patch looks good to me, but I would like to get tests past it - right now cannot merge anything until I have tracked down what is the issue.

PR #904, opened 2024-03-14
Add back s-bit for compiled fusermount

bsbernd:
This is also needed by some tests and was accidentally removed in commit aab146eea8877ee744a1b5a0da8bbbf31d14bad1

### DIFF
--- a/example/poll.c
+++ b/example/poll.c
@@ -33,6 +33,7 @@
 #include <time.h>
 #include <pthread.h>
 #include <poll.h>
+#include <stdbool.h>
 
 /*
  * fsel_open_mask is used to limit the number of opens to 1 per file.
@@ -51,6 +52,9 @@ static pthread_mutex_t fsel_mutex;	/* protects notify_mask and cnt array */
 static unsigned fsel_poll_notify_mask;	/* poll notification scheduled? */
 static struct fuse_pollhandle *fsel_poll_handle[FSEL_FILES]; /* poll notify handles */
 static unsigned fsel_cnt[FSEL_FILES];	/* nbytes stored in each file */
+static _Atomic bool fsel_stop = false;
+static pthread_t fsel_producer_thread;
+
 
 static int fsel_path_index(const char *path)
 {
@@ -59,6 +63,15 @@ static int fsel_path_index(const char *path)
 	if (strlen(path) != 2 || path[0] != '/' || !isxdigit(ch) || islower(ch))
 		return -1;
 	return ch <= '9' ? ch - '0' : ch - 'A' + 10;
+}
+
+static void fsel_destroy(void *private_data)
+{
+	(void)private_data;
+
+	fsel_stop = true;
+
+	pthread_join(fsel_producer_thread, NULL);
 }
 
 static int fsel_getattr(const char *path, struct stat *stbuf,
@@ -205,6 +218,7 @@ static int fsel_poll(const char *path, struct fuse_file_info *fi,
 }
 
 static const struct fuse_operations fsel_oper = {
+	.destroy        = fsel_destroy,
 	.getattr	= fsel_getattr,
 	.readdir	= fsel_readdir,
 	.open		= fsel_open,
@@ -220,7 +234,7 @@ static void *fsel_producer(void *data)
 
 	(void) data;
 
-	while (1) {
+	while (!fsel_stop) {
 		int i, t;
 
 		pthread_mutex_lock(&fsel_mutex);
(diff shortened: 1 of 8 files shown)


example/poll.c
```diff
@@ -51,6 +52,9 @@ static pthread_mutex_t fsel_mutex;	/* protects notify_mask and cnt array */
 static unsigned fsel_poll_notify_mask;	/* poll notification scheduled? */
 static struct fuse_pollhandle *fsel_poll_handle[FSEL_FILES]; /* poll notify handles */
 static unsigned fsel_cnt[FSEL_FILES];	/* nbytes stored in each file */
+static _Atomic bool fsel_stop = false;
```
ashleypittman, line 55: The _Atomic keyword here says this is an atomic...

example/poll.c
```diff
@@ -61,6 +65,15 @@ static int fsel_path_index(const char *path)
 	return ch <= '9' ? ch - '0' : ch - 'A' + 10;
 }
 
+static void fsel_destroy(void *private_data)
+{
+	(void)private_data;
+
+	fsel_stop = true;
```
ashleypittman, line 72: ... and if this is an atomic then I think you need to access it via the appropriate atomic_load function.

In practice I don't think this makes a difference on modern CPUs (and less so for a bool) but you're using the atomic api in one place and not in others and that's technically undefined behaviour.

example/poll.c
```diff
@@ -220,7 +234,7 @@ static void *fsel_producer(void *data)
 
 	(void) data;
 
-	while (1) {
+	while (!fsel_stop) {
```
ashleypittman, line 237: Again, this should be atomic_fetch

test/ci-build.sh
```diff
@@ -78,11 +88,18 @@ sanitized_build()
     # bug, cf. https://groups.google.com/forum/#!topic/mesonbuild/tgEdAXIIdC4
     meson configure -D b_lundef=false
 
-    meson configure
+    # additional options
+    meson configure "$@"
+
+    meson configure --no-pager
```
ashleypittman, line 94: We just had an interesting issue on my project, `meson configure` is probably not doing what you expect it to here.

`meson setup` will apply any config options and setup the build, `meson configure` will save configuration options but not setup the build.  The only reason this code works is that the subsequent `ninja` call will check if the options file is newer than the output of setup, and if so it'll re-run `meson setup`, but this depends on file age checks working which itself depends on timing, how quick the script runs and the backing filesystem etc.

After the last configure call and before calling `ninja` you need to add a explicit call to `meson setup --reconfigure "${SRC_DIR"` or similar.

example/poll.c
bsbernd: I might be wrong, but AFAIK it is sufficient to specify _Atomic for the type for C11 atomic operations 

https://developers.redhat.com/blog/2016/01/14/toward-a-better-use-of-c11-atomics-part-1#solutions_adopted_in_c11

See "Intuitive Syntax For Operations On Atomic Types"

test/ci-build.sh
bsbernd: The last "meson configure --no-pager" is there to see the resulting options. 
See the current tests - the entire branch is here because tests out of the sudden started to stall. I went back to 3.16.2 and tested that - same stall. 
Anyway, from the output I actually exactly get what I'm expecting.

ashleypittman: Yes, I agree.  Meson will have saved the configuration options correctly and will show them to you, but it won't have used them to re-generate the ninja config file, if you look at the output the first line from Ninja will be something like "regenerating config" which is ninja detecting the meson config is newer than the meson config and running "meson reconfigure" itself - but this is explicitly racey as it depends on file ages being different.

example/poll.c
ashleypittman: I may have been working from an older version of the spec.  This is great if true.

test/ci-build.sh
bsbernd: Thanks, I had thought about it again and had just added it. Thanks for your feedback. 

Also added backtraces in the signal handler, as I can't figure from github output why tests are failing. I see that `test_write_cache.c` fails when mnt_dir is missing, but I ensured that it is there.

ashleypittman: Thank you.

Consider the output at https://github.com/libfuse/libfuse/actions/runs/8357627990/job/22877175705 vs https://github.com/libfuse/libfuse/actions/runs/8357528585/job/22876857707
You can see the reconfigure happening both times but in the second case it's triggered by the ninja command itself, and this configure does need to happen for the options to be applied.

This is a thorny aspect of the meson interface, we should add the extra reconfigure line to https://github.com/libfuse/libfuse/blob/master/README.md?plain=1#L76 as well IMO.

bsbernd: Done.

PR #910, opened 2024-03-21
reset got_init after handling FUSE_DESTROY message

legezywzh:
User may still need to mount same fuse filesystem after umounting it(In this case, the userspace filesystem server needs to keep live), and after handling FUSE_DESTROY message, new FUSE_INIT message may come, so need to reset got_init to be zero.
This change is necessary when exporting existing fuse filesystems as virtiofs devices, users may mount or umount as they wish.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2204,6 +2204,7 @@ static void do_destroy(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 	(void) inarg;
 
 	se->got_destroy = 1;
+	se->got_init = 0;
 	if (se->op.destroy)
 		se->op.destroy(se->userdata);
 


bsbernd (approved):

bsbernd:
@legezywzh I would actually like to revert this , see https://github.com/libfuse/libfuse/pull/1326 and https://github.com/libfuse/libfuse/issues/1296 for reasons. 

The easiest solution for https://github.com/libfuse/libfuse/issues/1296 is to run single threaded before FUSE_INIT is complete. But that will not work when FUSE_INIT is received multiple times - right now I don't see how that might happen at all. Back when this was commit there was no harm to reset got_init, but in mean time with allowance of larger buffers, it is.

I also do not see at all how, FUSE_INIT could be send again - could you please explain that detail? If that is really needed we would basically needed to switch to single  threaded on handling FUSE_DESTROY.

PR #911, opened 2024-03-22
Example FS: High-perf passthrough over UDS

kbieganski:
This MR adds an example filesystem based on the `passthrough_hp.cc` example, with IO through Unix domain sockets.

The changes are similar to the ones in `hello_ll.c` -> `hello_ll_uds.c`.

Another difference from the original example is the option to provide a path to where the UDS should be created.
It defaults to `/tmp/libfuse-passthrough-hp.sock`.


example/passthrough_hp_uds.cc
```diff
@@ -0,0 +1,1406 @@
...
+        goto err_out2;
+
+    // Don't apply umask, use modes exactly as specified
+    umask(0);
+
+    // Mount and run main loop
+    loop_config = fuse_loop_cfg_create();
+
+    if (fs.num_threads != -1)
+        fuse_loop_cfg_set_idle_threads(loop_config, fs.num_threads);
```
bsbernd: That is a deprecated option, please use `_max_threads`. Unless you want to get persistent thread creation/destruction overhead. 
Hmm, I see, looks like I had accidentally also used "_idle_threads" in passthrough_hp.cc - needs to be fixed.

bsbernd (commented):
I will look at it more detailed later on. With all the code dup, I wonder if we could create a common file (like passthrough_hp_common.cc) and then just have the differences in passthrough_hp.cc and passthrough_uds.cc? Besides that it would simplify review, it would also simply to see for others what are the differences?

example/passthrough_hp_uds.cc
kbieganski: Note that the first commit simply copies an existing example.

bsbernd: We don't need to propagate that error even further. As I wrote, I would prefer a common file (feel free to copy it for that) so that we avoid unnecessary code dup.
I will submit a patch later on to fix the existing _idle_threads issue.

jmatuszewska: This might be a good idea.
I will split the logic as described and try to keep changes simple as they are still examples.

trapexit:
Could some documentation on why you'd want to use UDS be added?

jmatuszewska:
We've added the documentation about usage of UDS in the example.
Also  for `*.cc` files are added to Doxygen

bsbernd:
Thank you for your changes @jmatuszewska @kbieganski . The hard part to review is that it first duplicates code and then creates a common helper out of it - creates huge patches and more than one time. I will need a bit time and maybe will sort patches differently.

bsbernd:
@kbieganski I'm sorry for the long delay. Are you still interested in this? Also note that I would like to disable UDS by default, because there are safety issues with it - libfuse code does not verify what the kernel sends.

kbieganski:
We can close this for now, if we get back to it we'll open a new PR. Thanks.

bsbernd:
Thanks @kbieganski . Note that I'm going to merge https://github.com/libfuse/libfuse/pull/1602 for security reasons.

PR #912, opened 2024-03-23
Merge CMake changes

Smit-tay:
This merge is required because of the nature of this fork.

Not only is CMake build capability introduced, but some minor code clean-up is performed.


bsbernd (changes requested):
No, I don't see a reason why we should add another build system. Meson is perfectly fine.

Smit-tay:
Incorrectly submitted - intended to merge from my own branch into my master

Issue #819, opened 2023-07-26
Hash-pin workflow dependencies and keep them updated


pnacht:
Hey, it's Pedro and I'm back (see #793 and #797) with another security suggestion!

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

These dependencies can be kept up-to-date with dependabot. Whenever a new version of an Action is released, you'll receive a PR updating both its hash and the version comment (see [this repo](https://github.com/pnacht/logr/pulls) as an example).

You currently only have two workflow dependencies (`actions/checkout` and `codespell-project/actions-codespell`). Looking at their release history, this will probably lead to ~1 version bump PR per month.

I'll send a PR pinning the dependencies and adding dependabot along with this issue.

PR #913, opened 2024-03-23
Use major version numbers for actions.

ashleypittman:
Signed-off-by: Ashley Pittman <ashley@pittman.co.uk>

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -7,8 +7,7 @@ on:
   pull_request:
     branches: [master]
 
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
   codespell:
@@ -17,6 +16,6 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
+        uses: actions/checkout@v4
       - name: Codespell
-        uses: codespell-project/actions-codespell@94259cd8be02ad2903ba34a22d9c13de21a74461 # v2.0
+        uses: codespell-project/actions-codespell@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/codespell.yml
```diff
@@ -7,8 +7,7 @@ on:
   pull_request:
     branches: [master]
 
-permissions:
-  contents: read
+permissions: {}
```
bsbernd, line 11: The permission change needs to be documented in the commit message. I see that it disables all. This comes from https://github.com/libfuse/libfuse/issues/792 - what speaks against read permissions?

.github/workflows/codespell.yml
```diff
@@ -17,6 +16,6 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
+        uses: actions/checkout@v4
```
bsbernd, line 19: See this commit please
```
commit 0d7f43f110ec2a246bda906d4190767f8626446c
Author: Pedro Kaj Kjellerup Nacht <pnacht@google.com>
Date:   Wed Jul 26 13:48:07 2023 +0000

    Hash-pin workflow Actions
    
    Signed-off-by: Pedro Kaj Kjellerup Nacht <pnacht@google.com>
```

And the reason is documented here: https://github.com/libfuse/libfuse/issues/819 (should have better been part of the commit message.

Also read this please
https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

So I think we better keep the current way.

bsbernd (changes requested):

.github/workflows/codespell.yml
ashleypittman: Whilst this check does read from the repository it does so anonymously without using the token so my takeaway was that the token could be granted with no permissions.

.github/workflows/codespell.yml
ashleypittman: That makes a very good point.

PR #914, opened 2024-03-23
Install all test/build python packages from requirements.txt

ashleypittman:
Have GitHub actions setup python on runners and install
all python dependencies via pip.

Signed-off-by: Ashley Pittman <ashley@pittman.co.uk>

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -23,8 +23,11 @@ jobs:
     steps:
       - name: Install dependencies (Ubuntu)
         if: runner.os == 'Linux'
-        run: sudo apt-get update && sudo apt-get install -y clang doxygen gcc gcc-10 gcc-9 libstdc++-10-dev libstdc++-9-dev ninja-build python3-pip python3-setuptools valgrind
+        run: sudo apt-get update && sudo apt-get install -y clang doxygen gcc gcc-10 gcc-9 valgrind
       - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
-      - run: test/ci-install.sh
+      - uses: actions/setup-python@v5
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -0,0 +1,4 @@
+meson
+ninja
+looseversion
+pytest
```
bsbernd, line 4: Does this file support comments? If this stays in the project root, it needs to explain what it is for. Otherwise should be moved to .github/workflows.

.github/workflows/pr-ci.yml
```diff
@@ -23,8 +23,11 @@ jobs:
     steps:
       - name: Install dependencies (Ubuntu)
         if: runner.os == 'Linux'
-        run: sudo apt-get update && sudo apt-get install -y clang doxygen gcc gcc-10 gcc-9 libstdc++-10-dev libstdc++-9-dev ninja-build python3-pip python3-setuptools valgrind
```
bsbernd, line 26: I wonder if this needs to add `g++-10 g++-9`

bsbernd: Maybe not, as it passes tests without. Although I'm actually wondering why apt is installing g++ without a request for it - g++ might be in the 'Suggested:' section, which is then config dependent.

requirements.txt
ashleypittman: Yes it does.  requirements.txt is standard practice for anyone using python so I'd expect most people to recognise/understand it but as you say a comment at the top of the file would help.

.github/workflows/pr-ci.yml
ashleypittman: I wondered that.  It may be that it's already installed in the actions host as well.  Looking at the range of compilers is what I was going to look at next.

Issue #838, opened 2023-09-13
passthrough_hp: xfstests generic/477 triggers AddressSanitizer: heap-buffer-overflow 


bsbernd:
While running xfstests I noticed that generic/477 triggers a crash, debugging with ASAN points out

```
Sep 13 23:11:21 sq1 fusetests: run xfstest generic/477
Sep 13 23:11:22 sq1 systemd[1]: Started /bin/bash -c test -w /proc/self/oom_score_adj && echo 250 > /proc/self/oom_score_adj; exec ./tests/generic/477.
Sep 13 23:11:24 sq1 systemd[770]: mnt-test.mount: Succeeded.
Sep 13 23:11:24 sq1 systemd[1]: mnt-test.mount: Succeeded.
Sep 13 23:11:24 sq1 systemd[1340]: mnt-test.mount: Succeeded.
Sep 13 23:11:25 sq1 kernel: [  684.277452] sh (2650): drop_caches: 3
Sep 13 23:11:25 sq1 passthrough_hp: =================================================================
Sep 13 23:11:25 sq1 passthrough_hp: ==2646==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60c000012718 at pc 0x55d3efa07868 bp 0x7fc36d7fe8
a0 sp 0x7fc36d7fe898
Sep 13 23:11:25 sq1 passthrough_hp: READ of size 4 at 0x60c000012718 thread T1
Sep 13 23:11:25 sq1 passthrough_hp:     #0 0x55d3efa07867 in get_inode ../example/passthrough_hp.cc:176
Sep 13 23:11:25 sq1 passthrough_hp:     #1 0x55d3efa07ad8 in get_fs_fd ../example/passthrough_hp.cc:185
Sep 13 23:11:25 sq1 passthrough_hp:     #2 0x55d3efa09a04 in do_lookup ../example/passthrough_hp.cc:323
Sep 13 23:11:25 sq1 passthrough_hp:     #3 0x55d3efa0c783 in sfs_lookup ../example/passthrough_hp.cc:408
Sep 13 23:11:25 sq1 passthrough_hp:     #4 0x7fc371f572b2 in do_lookup ../lib/fuse_lowlevel.c:1103
Sep 13 23:11:25 sq1 passthrough_hp:     #5 0x7fc371f72351 in fuse_session_process_buf_int ../lib/fuse_lowlevel.c:2771
[...]

PR #917, opened 2024-03-27
Add more documentation for FUSE_CAP_EXPORT_SUPPORT

bsbernd:
Also remove the flag from passthrough_ll.c and passthrough_hp.cc as these implementations do _not_ handle that flag. They just cast fuse_ino_t to an inode and cause a heap buffer overflow for unknown objects (simplest reproducer are the examples in "man 2 open_by_handle_at", but to unmount/mount the file system after name_to_handle_at and before open_by_handle_at).

Fixes https://github.com/libfuse/libfuse/issues/838

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -185,7 +185,8 @@ struct fuse_loop_config_v1 {
  *
  * When this flag is set, the filesystem must be prepared to receive requests
  * for invalid inodes (i.e., for which a FORGET request was received or
- * which have been used in a previous instance of the filesystem daemon).
+ * which have been used in a previous instance of the filesystem daemon) and
+ * must not re-use node-ids (even when setting generation numbers).
  *
  * This feature is disabled by default.
  */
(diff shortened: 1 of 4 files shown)


bsbernd:
@Nikratio I think you had set these flags in passthrough_ll and passthrough_hp - more for your information, that you don't accidentally set it in other file system.

include/fuse_common.h
```diff
@@ -182,6 +182,16 @@ struct fuse_loop_config_v1 {
 
 /**
  * Indicates that the filesystem supports lookups of "." and "..".
+ * Setting this flag can be dangerous and cause heap-buffer-overflows,
+ * if the file system server just uses the received node-id (fuse_ino_t)
+ * and casts it from uint64_t to a server object (as an inode object).
```
Nikratio, line 187: This seems overly specific for a general API documentation. I would suggest to remove this and instead say something like "When this flag is set, the filesystem must be prepared to receive requests for invalid inodes (i.e., for which a FORGET request was received or which have been used in a previous instance of the filesystem daemon).

bsbernd: Thanks, Nikolaus! I will take over your suggestion!

Nikratio:
I've edited the pull request to expand a bit more on this. Hope that's okay.

bsbernd:
Thanks a lot @Nikratio, sounds good to me.

PR #919, opened 2024-03-29
Enable passthrough mode for read/write operations

amir73il:
Add support for filesystem passthrough read/write of files.

When the FUSE_PASSTHROUGH capability is enabled, the FUSE server may decide, while handling the "open" or "create" requests, if the given file can be accessed by that process in "passthrough" mode, meaning that all the further read and write operations would be forwarded by the kernel directly to the backing file rather than to the FUSE server. All requests other than read or write are still handled by the server.

This allows for an improved performance on reads and writes, especially in the case of reads at random offsets, for which no (readahead) caching mechanism would help, reducing the performance gap between FUSE and native filesystem access.

Extend also the passthrough_hp example with the new passthrough feature. This example opens a kernel backing file per FUSE inode on the first FUSE file open of that inode and closes the backing file on the release of the last FUSE file on that inode.

All opens of the same inode passthrough to the same backing file. A combination of fi->direct_io and fi->passthrough is allowed. It means that read/write operations go directly to the server, but mmap is done on the backing file.

This allows to open some fds of the inode in passthrough mode and some fd of the same inode in direct_io/passthrough_mmap mode.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -123,6 +123,7 @@ struct Inode {
     dev_t src_dev {0};
     ino_t src_ino {0};
     int generation {0};
+    int backing_id {0};
     uint64_t nopen {0};
     uint64_t nlookup {0};
     std::mutex m;
@@ -159,6 +160,7 @@ struct Fs {
     bool clone_fd;
     std::string fuse_mount_options;
     bool direct_io;
+    bool passthrough;
 };
 static Fs fs{};
 
@@ -190,7 +192,15 @@ static int get_fs_fd(fuse_ino_t ino) {
 
 static void sfs_init(void *userdata, fuse_conn_info *conn) {
     (void)userdata;
-    if (fs.timeout && conn->capable & FUSE_CAP_WRITEBACK_CACHE)
+
+    if (fs.passthrough && conn->capable & FUSE_CAP_PASSTHROUGH)
+        conn->want |= FUSE_CAP_PASSTHROUGH;
+    else
+        fs.passthrough = false;
+
+    /* Passthrough and writeback cache are conflicting modes */
+    if (fs.timeout && !fs.passthrough &&
+        conn->capable & FUSE_CAP_WRITEBACK_CACHE)
         conn->want |= FUSE_CAP_WRITEBACK_CACHE;
 
     if (conn->capable & FUSE_CAP_FLOCK_LOCKS)
@@ -810,6 +820,30 @@ static void sfs_releasedir(fuse_req_t req, fuse_ino_t ino, fuse_file_info *fi) {
 }
 
 
+static void do_passthrough_open(fuse_req_t req, fuse_ino_t ino, int fd,
+                                fuse_file_info *fi) {
+    Inode& inode = get_inode(ino);
+    /* Setup a shared backing file on first open of an inode */
+    if (inode.backing_id) {
+        if (fs.debug)
+            cerr << "DEBUG: reusing shared backing file "
+                 << inode.backing_id << " for inode " << ino << endl;
+        fi->backing_id = inode.backing_id;
+    } else if (!(inode.backing_id = fuse_passthrough_open(req, fd))) {
+        cerr << "DEBUG: fuse_passthrough_open failed for inode " << ino
+             << ", disabling rw passthrough." << endl;
+        fs.passthrough = false;
+    } else {
+        if (fs.debug)
+            cerr << "DEBUG: setup shared backing file "
+                 << inode.backing_id << " for inode " << ino << endl;
+        fi->backing_id = inode.backing_id;
+    }
+    /* open in passthrough mode must drop old page cache */
+    if (fi->backing_id)
+        fi->keep_cache = false;
+}
+
 static void sfs_create(fuse_req_t req, fuse_ino_t parent, const char *name,
                        mode_t mode, fuse_file_info *fi) {
     Inode& inode_p = get_inode(parent);
@@ -845,6 +879,8 @@ static void sfs_create(fuse_req_t req, fuse_ino_t parent, const char *name,
     Inode& inode = get_inode(e.ino);
     lock_guard<mutex> g {inode.m};
     inode.nopen++;
+    if (fs.passthrough)
+        do_passthrough_open(req, e.ino, fd, fi);
     fuse_reply_create(req, &e, fi);
 }
 
@@ -914,6 +950,8 @@ static void sfs_open(fuse_req_t req, fuse_ino_t ino, fuse_file_info *fi) {
     fi->parallel_direct_writes = 1;
 
     fi->fh = fd;
+    if (fs.passthrough)
+        do_passthrough_open(req, ino, fd, fi);
     fuse_reply_open(req, fi);
 }
 
@@ -922,6 +960,19 @@ static void sfs_release(fuse_req_t req, fuse_ino_t ino, fuse_file_info *fi) {
     Inode& inode = get_inode(ino);
     lock_guard<mutex> g {inode.m};
     inode.nopen--;
+
+    /* Close the shared backing file on last file close of an inode */
+    if (inode.backing_id && !inode.nopen) {
+        if (fuse_passthrough_close(req, inode.backing_id) < 0) {
+            cerr << "DEBUG: fuse_passthrough_close failed for inode "
+                 << ino << " backing file " << inode.backing_id << endl;
+        } else if (fs.debug) {
+            cerr << "DEBUG: closed backing file " << inode.backing_id
+                 << " for inode " << ino << endl;
+        }
+        inode.backing_id = 0;
+    }
+
     close(fi->fh);
     fuse_reply_err(req, 0);
 }
@@ -960,6 +1011,11 @@ static void do_read(fuse_req_t req, size_t size, off_t off, fuse_file_info *fi) 
 static void sfs_read(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
                      fuse_file_info *fi) {
     (void) ino;
+    if (fs.passthrough && !fs.direct_io) {
+        cerr << "ERROR: fuse_passthrough read failed." << endl;
+        fuse_reply_err(req, EIO);
+        return;
+    }
     do_read(req, size, off, fi);
 }
 
@@ -983,6 +1039,11 @@ static void do_write_buf(fuse_req_t req, size_t size, off_t off,
 static void sfs_write_buf(fuse_req_t req, fuse_ino_t ino, fuse_bufvec *in_buf,
                           off_t off, fuse_file_info *fi) {
     (void) ino;
+    if (fs.passthrough && !fs.direct_io) {
+        cerr << "ERROR: fuse_passthrough write failed." << endl;
+        fuse_reply_err(req, EIO);
+        return;
+    }
     auto size {fuse_buf_size(in_buf)};
     do_write_buf(req, size, off, in_buf, fi);
 }
@@ -1232,6 +1293,7 @@ static cxxopts::ParseResult parse_options(int argc, char **argv) {
         ("help", "Print help")
         ("nocache", "Disable attribute all caching")
         ("nosplice", "Do not use splice(2) to transfer data")
+        ("nopassthrough", "Do not use pass-through mode for read/write")
         ("single", "Run single-threaded")
         ("o", "Mount options (see mount.fuse(5) - only use if you know what "
               "you are doing)", cxxopts::value(mount_options))
@@ -1239,7 +1301,6 @@ static cxxopts::ParseResult parse_options(int argc, char **argv) {
                         cxxopts::value<int>()->default_value(SFS_DEFAULT_THREADS))
         ("clone-fd", "use separate fuse device fd for each thread")
         ("direct-io", "enable fuse kernel internal direct-io");
-
 
     // FIXME: Find a better way to limit the try clause to just
     // opt_parser.parse() (cf. https://github.com/jarro2783/cxxopts/issues/146)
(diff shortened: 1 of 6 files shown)


include/fuse_common.h
```diff
@@ -447,6 +452,18 @@ struct fuse_loop_config_v1 {
...
+ * Indicates support for passthrough mode access for read/write operations.
+ *
+ * If this flag is set in the `capable` field of the `fuse_conn_info`
+ * structure, then the FUSE kernel module supports redirecting read/write
+ * operations to the backing file instead of letting them to be handled
+ * by the FUSE daemon.
+ *
+ * This feature is disabled by default.
+ */
+#define FUSE_CAP_PASSTHROUGH      (1 << 28)
```
nanonyme, line 465: This conflicts with above FUSE_CAP_DIRECT_IO_ALLOW_MMAP

nanonyme: I suppose this should simply be `1 << 29`

amir73il: yap. fixed. Thanks!

include/fuse_lowlevel.h
```diff
@@ -1381,6 +1381,19 @@ int fuse_reply_attr(fuse_req_t req, const struct stat *attr,
...
 int fuse_reply_readlink(fuse_req_t req, const char *link);
 
+/**
+ * Setup passthrough backing file for open reply
+ *
+ * Possible requests:
+ *   open, opendir, create
+ *
+ * @param req request handle
+ * @param fd backing file descrptor
```
nanonyme, line 1391: Probably want to fix this typo resulting in static analysis failure while waiting for actual maintainers to review.

amir73il: done

lib/fuse_versionscript
```diff
@@ -166,6 +166,8 @@ FUSE_3.7 {
 	global:
 		fuse_set_log_func;
 		fuse_log;
+		fuse_passthrough_open;
+		fuse_passthrough_close;
 } FUSE_3.4;
```
bsbernd, line 171: That is not right here and will cause issues for distributions. Could you please put it under FUSE_3.17? We might need to change it again to 3.18 if 3.17 gets released before (it probably will because of the ABI issue) - hmm, maybe wait a few days, I hope I get a bit time tomorrow morning to look into that.

example/passthrough_hp.cc
```diff
@@ -193,7 +195,14 @@ static void sfs_init(void *userdata, fuse_conn_info *conn) {
     if (conn->capable & FUSE_CAP_EXPORT_SUPPORT)
         conn->want |= FUSE_CAP_EXPORT_SUPPORT;
 
-    if (fs.timeout && conn->capable & FUSE_CAP_WRITEBACK_CACHE)
+    if (fs.passthrough && conn->capable & FUSE_CAP_PASSTHROUGH)
+        conn->want |= FUSE_CAP_PASSTHROUGH;
+    else
+        fs.passthrough = false;
```
bsbernd, line 201: Hmm, no option for it? I personally use passthrough_hp to benchmarks for the kernel/userspace interface performance and also use it for xfstests - I think we need an option to at least disable passthrough.

bsbernd: Ah sorry, ignore me, it gets late and I had missed the "fs.passthrough" condition.

lib/fuse_lowlevel.c
```diff
@@ -2157,6 +2189,10 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 		outargflags |= FUSE_SETXATTR_EXT;
 	if (se->conn.want & FUSE_CAP_DIRECT_IO_ALLOW_MMAP)
 		outargflags |= FUSE_DIRECT_IO_ALLOW_MMAP;
+	if (se->conn.want & FUSE_CAP_PASSTHROUGH) {
+		outargflags |= FUSE_PASSTHROUGH;
+		outarg.max_stack_depth = se->conn.max_backing_stack_depth + 1;
```
bsbernd, line 2194: Could you please add a comment here why it is +1?

lib/fuse_versionscript
amir73il: Should I use FUSE_3.17 { } FUSE_3.4;?
What does FUSE_3.4 mean in this context?

lib/fuse_lowlevel.c
amir73il: done

include/fuse_common.h
```diff
@@ -593,10 +610,21 @@ struct fuse_conn_info {
...
 	unsigned time_gran;
 
+	/**
+	 * When FUSE_CAP_PASSTHROUGH is enabled, this is the maximum allowed
+	 * stacking depth of the backing files. The default is 0, meaning that
+	 * the backing files cannot be on a stacked filesystem.
+	 * Set this to 1 if backing files are expected to be on a stacked
+	 * filesystem, such as overlayfs or another FUSE passthrough.
+	 * In current kernel, FILESYSTEM_MAX_STACK_DEPTH is 2, which includes
+	 * the FUSE layer, so max_backing_stack_depth > 1 is not supported.
```
trapexit, line 620: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fuse/inode.c?h=v6.9-rc2#n1326

If I'm reading this correctly... that is not correct. It isn't if max > 1 but max > 2. The kernel check is `arg->max_stack_depth <= FILESYSTEM_MAX_STACK_DEPTH`

nanonyme: Good point. It's important to get this exactly right if you are eg running FUSE inside Docker container so you are already on top of overlayfs. The limit of 2 is *really* low.

lib/fuse_versionscript
bsbernd: https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_mono/ld.html#SEC25

To my understanding "FUSE_3.4" is what the new version depends on. Given that we are at ABI version 3.12 it should be `FUSE_3.17 { } FUSE_3.12;`, afaik.

With the possibility that 3.17 needs to be replaced by 3.18 later this week.

include/fuse_common.h
bsbernd: I guess it relates to `In current kernel, FILESYSTEM_MAX_STACK_DEPTH is 2, which includes
	 * the FUSE layer, so max_backing_stack_depth > 1 is not supported.`

trapexit: But the literal check in FUSE is <= 2. Unless there is an additional check somewhere being done being referenced. In which case why would FUSE be checking?

nanonyme: I wonder if it would make sense to email the person who implemented the kernel feature and kindly request them to review the libfuse counterpart.

trapexit: I think I'd prefer a hybrid approach given that making it that binary choice would be awkward to undo down the road if it did change. Maybe just expand the docs to include that information? Described not too differently from what you said here.

bsbernd: @trapexit Sorry for my late reply, you might have found it already, if not, see in `do_init` 
```
if (se->conn.want & FUSE_CAP_PASSTHROUGH) {
		outargflags |= FUSE_PASSTHROUGH;
		/*
		 * outarg.max_stack_depth includes the fuse stack layer,
		 * so it is one more than max_backing_stack_depth.
		 */
		outarg.max_stack_depth = se->conn.max_backing_stack_depth + 1;```

File: lib/fuse_versionscript (hunk 3 above)
Comment (amir73il) [reply lib/fuse_versionscript:171]: ok, it is a mistake that the last section is FUSE_3.12 { } FUSE_3.4 ? 
should it not be FUSE_3.12 {} FUSE_3.7 ?

File: include/fuse_common.h (hunk 6 above)
Comment (amir73il) [reply include/fuse_common.h]: ok, added FUSE_BACKING_STACKED_{OVER,UNDER} macros (not enum) and explained them in documentation

File: lib/fuse_versionscript (hunk 3 above)
Comment (bsbernd) [reply lib/fuse_versionscript:171]: Ouch, I didn't see that, yeah, I had used the wrong dependency value. I'm not sure if I should change it now. Maybe all is fine, as the fuse_session_loop dependency symbols are defined in 3.2.

**Comment by trapexit** [comment]
Passthrough is only allowed if `capable(CAP_SYS_ADMIN)`. In my fuse server I need to seteuid to the client's uid so no CAP_SYS_ADMIN. Setting KEEP_CAPS wouldn't be appropriate so I'm guessing the best strategy would be to seteuid back to root before calling ioctl?

**Comment by amir73il** [comment]
> Passthrough is only allowed if `capable(CAP_SYS_ADMIN)`. In my fuse server I need to seteuid to the client's uid so no CAP_SYS_ADMIN. Setting KEEP_CAPS wouldn't be appropriate so I'm guessing the best strategy would be to seteuid back to root before calling ioctl?

That's an option. Or seteuid only for specific operations that need it like creating files. 
Do you need seteuid for all of operations? Considering that permissions are already enforced also on fuse later? 

I don't remember if you told me about your filesystem but it always helps to know the big picture of you want to best answer

**Comment by amir73il** [comment]
> 
> Does this mean real CAP_SYS_ADMIN? So will not be possible under unprivileged user namespaces?

That is correct.
As the comment in the kernel code says: /* TODO: relax CAP_SYS_ADMIN once backing files are visible to lsof */
This TODO is not high on my priority, so if anyone has interest to push this forward, I encourage them to try to take on this work.

**Comment by nanonyme** [comment]
> > Does this mean real CAP_SYS_ADMIN? So will not be possible under unprivileged user namespaces?
> 
> That is correct. As the comment in the kernel code says: /* TODO: relax CAP_SYS_ADMIN once backing files are visible to lsof */ This TODO is not high on my priority, so if anyone has interest to push this forward, I encourage them to try to take on this work.

So what exactly happens when filesystem requests passthrough but isn't running privileged? Will there be a silent fallback to no passthrough?

**Comment by trapexit** [comment]
> So what exactly happens when filesystem requests passthrough but isn't running privileged? Will there be a silent fallback to no passthrough?

The ioctl call is denied and the server has to determine what it will do. Passthrough is not a transparent feature. The server must support it explicitly.

**Comment by trapexit** [comment]
Not sure what you mean. The server can do whatever it wants.

File: lib/fuse_versionscript (hunk 3 above)
Comment (amir73il) [reply lib/fuse_versionscript]: @bsbernd I see the libfuse version was increased to 3.17.0.
Did you mean that the passthrough feature should increment the libfuse version to 3.18.0?

Comment (bsbernd) [reply lib/fuse_versionscript]: It is not finished yet, the version is was just increased to prepare the branch already.
I have created a separate patch/PR to add in the libfuse version an application was compiled against (https://github.com/libfuse/libfuse/pull/942), should be complete now, but I didn't have time for manual testing yet. I hope this weekend.  After that will add in auto-correct of the ABI issues for < 3.17.0 and then we should be good to release.

File: lib/fuse_versionscript [hunk 7]
```diff
@@ -187,6 +187,12 @@ FUSE_3.12 {
 		fuse_lowlevel_notify_expire_entry;
 } FUSE_3.4;
 
+FUSE_3.18 {
```
bsbernd, line 190: 3.17

bsbernd:
@amir73il needs to be rebasd against master. I guess you are busy today and this week. Shall I force push into your branch in (my) evening?

bsbernd (approved):
Thanks Amir, LGTM.

mxmlnkn:
I was looking through this PR and the [kernel patches](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6ce8b2ce0d7e3a621cdc9eb66d74436ca7d0e66e), especially the definition of `struct fuse_backing_map`. If I may ask, do I understand this correctly that pass-through only works on file descriptor granularity? What I would have desired was passthrough to a byte-range pointing inside an existing file. Such finer granularities were mentioned in this [presentation](https://youtu.be/3IjdxsFPoHE?t=97), but aside from mentioning it, it was not discussed again there.

amir73il:
> I was looking through this PR and the [kernel patches](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6ce8b2ce0d7e3a621cdc9eb66d74436ca7d0e66e), especially the definition of `struct fuse_backing_map`. If I may ask, do I understand this correctly that pass-through only works on file descriptor granularity? What I would have desired was passthrough to a byte-range pointing inside an existing file. Such finer granularities were mentioned in this [presentation](https://youtu.be/3IjdxsFPoHE?t=97), but aside from mentioning it, it was not discussed again there.

setup passthrough to byte range was not implemented.
the API could be extended to support it in the future.
I currently have no estimation about how the implementation effort and design concerns.
I also have no use case for this functionality, so am not likely to be the one driving it upstream.

mxmlnkn:
> > I was looking through this PR and the [kernel patches](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6ce8b2ce0d7e3a621cdc9eb66d74436ca7d0e66e), especially the definition of `struct fuse_backing_map`. If I may ask, do I understand this correctly that pass-through only works on file descriptor granularity? What I would have desired was passthrough to a byte-range pointing inside an existing file. Such finer granularities were mentioned in this [presentation](https://youtu.be/3IjdxsFPoHE?t=97), but aside from mentioning it, it was not discussed again there.
> 
> setup passthrough to byte range was not implemented. the API could be extended to support it in the future. I currently have no estimation about how the implementation effort and design concerns. I also have no use case for this functionality, so am not likely to be the one driving it upstream.

I see, too bad. I would use it to map byte ranges inside non-compressed archives to files. It should work fine for TAR and ZIP. SquashFS would only work easily without tail-end compression but even that might be possible to get to work somehow. I don't have any experience with kernel development. Currently, I am hoping, and need to verify, that maybe the long-planned FUSE eBPF feature could be used to implement byte ranges in the future. The idea would be something like this:

 - Add parameter pre-processing via a BPF callback before each read/write call that adds a value onto the requested file offset. I.e., when reading from byte offset 0 in the FUSE file, it would map to a read from offset 3000 in the backing file descriptor.
 - These offsets, to be added to the request offsets, could be stored in a BPF hash map keyed by the inode. Each open operation would insert the byte range information into the hash map for the opened inode.

I have never worked with eBPF, so I am not sure whether this would work. Maybe modifying BPF maps is rather expensive or there are other problems. I'll need to read up on all that and try some things out.

Thanks for all your work enabling the current passthrough!

PR #921, opened 2024-04-01
ABI issues

bsbernd:
See individual commits.

### DIFF
--- a/lib/compat.c
+++ b/lib/compat.c
@@ -51,6 +51,36 @@ int fuse_parse_cmdline(struct fuse_args *args,
 {
 	return fuse_parse_cmdline_30(args, opts);
 }
+
+
+#ifdef fuse_session_new
+#undef fuse_session_new
+#endif
+struct fuse_session *fuse_session_new_317(struct fuse_args *args,
+					  const struct fuse_lowlevel_ops *op,
+					  size_t op_size,
+					  struct libfuse_version *version,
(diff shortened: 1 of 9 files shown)


lib/compat.c
```diff
@@ -51,6 +51,36 @@ int fuse_parse_cmdline(struct fuse_args *args,
...
+
+#ifdef fuse_session_new
+#undef fuse_session_new
+#endif
+struct fuse_session *fuse_session_new_317(struct fuse_args *args,
+					  const struct fuse_lowlevel_ops *op,
+					  size_t op_size,
+					  struct libfuse_version *version,
+					  void *userdata);
+struct fuse_session *fuse_session_new(struct fuse_args *args,
```
amir73il, line 64: this one is redundant?

bsbernd: Please see above `#if (!defined(LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS))` 

With versioned symbols `fuse_session_new` is a just a redirected by glibc/ld to symbol (ABI) and by the fuse_session_new macro to fuse_session_new_317 (API). 
Without versioned symbols (what this file is about), we would have a compat issue as the new function is now called `fuse_session_new317`- we define the function here, but there is no user that can call it from the API, but it still exists for the ABI.

bsbernd:
Closing as PR #1081 has been merged.

PR #923, opened 2024-04-03
Add fuse_get_conn_info()

legezywzh:
I know that we already have fuse_parse_conn_info_opts() and fuse_apply_conn_info_opts(), but it seems a bit complicated, and will add many options to apps. With this new helper, developers can modify fuse_conn_info more easily, such as:
	struct fuse_conn_info *conn;
	conn = fuse_get_conn_info(se);
	conn->max_write = 8 * 4096;


lib/fuse_versionscript
```diff
@@ -129,6 +129,7 @@ FUSE_3.0 {
 		fuse_parse_conn_info_opts;
 		fuse_fs_lseek;
 		fuse_reply_lseek;
+		fuse_get_conn_info;
```
bsbernd, line 132: That is not right here, but needs a new `FUSE_3.17 {  } FUSE_3.12;` section, as that symbol is not part of libfuse-3.0.

 I'm going to push a patch today to increase master to 3.17.

bsbernd (changes requested):
I would like to ask to add a user for that in one of  the examples. Is my guess correct and you plan to use it in ->init()?

bsbernd:
I wouldn't like to merge this without a single user. 
There is no passthrough_ll_dpu.c right now - this patch should be part of adding that. 
If you plan to copy passthrough_ll.c to passthrough_ll_dpu.c, I would also like to ask to either add the changes as option to passthrough_ll.c or to create a passthrough_ll_common.c file that gets used by passthrough_ll.c  and passthrough_ll_dpu.c. With code duplication it is hard to see what are the differences.

legezywzh:
> I wouldn't like to merge this without a single user. There is no passthrough_ll_dpu.c right now - this patch should be part of adding that. If you plan to copy passthrough_ll.c to passthrough_ll_dpu.c, I would also like to ask to either add the changes as option to passthrough_ll.c or to create a passthrough_ll_common.c file that gets used by passthrough_ll.c and passthrough_ll_dpu.c. With code duplication it is hard to see what are the differences.

OK, I understand your concerns and will try to submit a new version patch containing a use case, seems example/hello_ll.c is a much simple case.

legezywzh:
> Personally I find `fuse_parse_conn_info_opts()` and ` fuse_apply_conn_info_opts()` confusing. I guess I wouldn't have created these functions at all and would have left it to file systems adding their own options for that. I wonder how many users there are for these functions and if we could mark those functions as deprecated to avoid new users for these (using `__attribute__((deprecated))`). What I mean is that struct fuse_conn_info is exposed in ->init() to file systems, so why are there fuse helper functions needed...?
> 
> Regarding limiting se->conn.max_write, yeah, I guess you are right. Kind of. In the end it is used only further down in do_init() to set the kernel limit and kernel side has its own checks for it. From my point of view the sanity checks in libfuse could be removed. Just makes the code complex and gives false impressions, imho.

Totally agree with you.

lib/fuse_versionscript
legezywzh: > That is not right here, but needs a new `FUSE_3.17 { } FUSE_3.12;` section, as that symbol is not part of libfuse-3.0.
> 
> I'm going to push a patch today to increase master to 3.17.

Hi, have you finished this patch? I need to rebase my patch on top of it, thanks.

legezywzh:
hello, I have pushed a new version, please help to review if you have some free time.

example/hello_ll.c
```diff
@@ -242,11 +263,19 @@ int main(int argc, char *argv[])
...
+
 	se = fuse_session_new(&args, &hello_ll_oper,
 			      sizeof(hello_ll_oper), NULL);
 	if (se == NULL)
 	    goto err_out1;
 
+	if (lo.max_write) {
+		conn = fuse_get_conn_info(se);
+		conn->max_write = lo.max_write;
+	}
```
bsbernd, line 277: Thank you, the intent is clear now, though now I don't understand why can't do the same in the ->init() method, which actually gets passed conn

In `do_init()`
```
	if (se->op.init)
		se->op.init(se->userdata, &se->conn);
```

legezywzh: Yes, we can do it in ->init() method, but as i said before, there is a bug there:
```
diff --git a/example/hello_ll.c b/example/hello_ll.c
index bea2112..007d4f8 100644
--- a/example/hello_ll.c
+++ b/example/hello_ll.c
@@ -201,8 +201,15 @@ static void hello_ll_removexattr(fuse_req_t req, fuse_ino_t ino, const char *nam
                fuse_reply_err(req, ENOTSUP);
        }
 }
+static void hello_ll_init(void *userdata, struct fuse_conn_info *conn)
+{
+       (void)userdata;
+       conn->max_write = 512 * 4096;
+       return;
+}
 
 static const struct fuse_lowlevel_ops hello_ll_oper = {
+       .init = hello_ll_init,
        .lookup = hello_ll_lookup,
        .getattr = hello_ll_getattr,
        .readdir = hello_ll_readdir,
index d9844b0..3f13f2a 100644
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2189,6 +2189,7 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
                fuse_log(FUSE_LOG_DEBUG, "   max_readahead=0x%08x\n",
                        outarg.max_readahead);
                fuse_log(FUSE_LOG_DEBUG, "   max_write=0x%08x\n", outarg.max_write);
+               fuse_log(FUSE_LOG_DEBUG, "   fuse_session_size=0x%08x\n", se->bufsize);
                fuse_log(FUSE_LOG_DEBUG, "   max_background=%i\n",
                        outarg.max_background);
                fuse_log(FUSE_LOG_DEBUG, "   congestion_threshold=%i\n",
```
I tried above codes, it'll fail:
[lege@fedora build]$ sudo ./example/hello_ll -d /home/lege/mntpoint/ 
FUSE library version: 3.16.2
unique: 2, opcode: INIT (26), nodeid: 0, insize: 104, pid: 0
INIT: 7.33
flags=0x53fffffb
max_readahead=0x00020000
   INIT: 7.39
   flags=0x40409029
   max_readahead=0x00020000
   max_write=0x00200000
   fuse_session_size=0x00101000
   max_background=0
   congestion_threshold=0
   time_gran=1
   unique: 2, success, outsize: 80
fuse: reading device: Invalid argument

The reason is that "conn->max_write + FUSE_BUFFER_HEADER_SIZE" does not match se->bufsize. This bug could be fixed by below diff:
```
index d9844b0..99c6075 100644
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2079,13 +2079,13 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
        }
        se->bufsize = bufsize;
 
-       if (se->conn.max_write > bufsize - FUSE_BUFFER_HEADER_SIZE)
-               se->conn.max_write = bufsize - FUSE_BUFFER_HEADER_SIZE;
-
        se->got_init = 1;
        if (se->op.init)
                se->op.init(se->userdata, &se->conn);
 
+       if (se->conn.max_write > bufsize - FUSE_BUFFER_HEADER_SIZE)
+               se->conn.max_write = bufsize - FUSE_BUFFER_HEADER_SIZE;
+
```
then
[lege@fedora build]$ sudo ./example/hello_ll -d /home/lege/mntpoint/ 
FUSE library version: 3.16.2
unique: 2, opcode: INIT (26), nodeid: 0, insize: 104, pid: 0
INIT: 7.33
flags=0x53fffffb
max_readahead=0x00020000
   INIT: 7.39
   flags=0x40409029
   max_readahead=0x00020000
   max_write=0x00100000
   fuse_session_size=0x00101000
   max_background=0
   congestion_threshold=0
   time_gran=1
   unique: 2, success, outsize: 80

No error occurs, and it seems that libfuse will limit conn->max_write to not greater than se->bufsize, in fuse_session_new(), it sets "se->bufsize = FUSE_MAX_MAX_PAGES * getpagesize() + FUSE_BUFFER_HEADER_SIZE" initially.

So far, I only care about whether conn->max_write can be set correctly, so I'm ok to close this pr(not mergerd), but
above do_init()'s bug needs to be fixed, I can submit a patch, what do you think?
And make conn->max_write not greater than se->bufsize is libfuse's design intent? Can users set conn->max_write to be greater than the default " FUSE_MAX_MAX_PAGES * getpagesize()"?  whih may need explict documents, thanks.

bsbernd:
Yeah, would be great if you could create PR for the diff above. I think we will need to do more about that in a follow-up. Basically se->bufsize is updated in `do_init()` - kernel side might reduce what libfuse tells is to use, but the failure you got probably comes from kernel `fuse_dev_do_read`

```
	 */
			   fc->max_write))
		return -EINVAL;
```

I.e. `se->bufsize ` was updated, but then not honored correctly.

bsbernd:
@legezywzh Are you ok to close this?

PR #924, opened 2024-04-05
fusermount: Fix use of uninitialized x_mnt_opts

bsbernd:
x_mnt_opts was not initialized to 0, but strncat was done

Spotted by cppcheck running on our ddn branch.


ashleypittman (approved):
calloc should be standard IMO.  None of this is performance critical.

PR #929, opened 2024-04-18
passthrough_ll: fix fd leaks in lo_destroy()

legezywzh:
By virtio-fs and libfuse fuse_custom_io, passthrough_ll could be a virtio filesystem device backend, this bug was found when doing mount, fsstress and umount repeatedly.

### DIFF
--- a/example/passthrough_ll.c
+++ b/example/passthrough_ll.c
@@ -190,6 +190,7 @@ static void lo_destroy(void *userdata)
 	while (lo->root.next != &lo->root) {
 		struct lo_inode* next = lo->root.next;
 		lo->root.next = next->next;
+		close(next->fd);
 		free(next);
 	}
 }


bsbernd:
@legezywzh Out of interest, did you observe a real leak? In principle kernel side should first send FUSE_FORGOT and once all inodes are release it should send FUSE_DESTRPOY? Though I need to check for "umount -f", I'm not sure if it sends FORGET in that case.

legezywzh:
> @legezywzh Out of interest, did you observe a real leak? In principle kernel side should first send FUSE_FORGOT and once all inodes are release it should send FUSE_DESTRPOY? Though I need to check for "umount -f", I'm not sure if it sends FORGET in that case.


@bsbernd  yeah, firstly I observed real fd leaks in 5.10 kernel, later even in 6.9-rc4, I also found fd leaks.
You're right that kernel virtio-fs or fuse would evict all inodes in the process of umount,
->virtio_kill_sb
-->virtio_fs_conn_destroy->fuse_conn_destroy->fuse_abort_conn, which will set fiq->connected to be zero.
-->generic_shutdown_super
--->evict_inodes
----> fuse_evict_inode
-----> fuse_queue_forget
```
void fuse_queue_forget(struct fuse_conn *fc, struct fuse_forget_link *forget,
                       u64 nodeid, u64 nlookup)
{
        struct fuse_iqueue *fiq = &fc->iq;
        
        forget->forget_one.nodeid = nodeid; 
        forget->forget_one.nlookup = nlookup;
     
        spin_lock(&fiq->lock);
        if (fiq->connected) {
                fiq->forget_list_tail->next = forget;
                fiq->forget_list_tail = forget;
                fiq->ops->wake_forget_and_unlock(fiq);
        } else {
                kfree(forget);
                spin_unlock(&fiq->lock);
        }
}
```
but wake_forget_and_unlock() is only called when fiq->connected is true., and fuse_abort_conn has set it to be false before.

PR #930, opened 2024-04-18
fuse_session_custom_io: Add a struct size field

bsbernd:
As in commit 73cd124d 'struct fuse_custom_io' might get extended and in order to stay ABI compatible libfuse has to zero all extended fields. In order to do so it has to know the size of the struct the application knows about.

### DIFF
--- a/example/hello_ll_uds.c
+++ b/example/hello_ll_uds.c
@@ -341,7 +341,7 @@ int main(int argc, char *argv[])
 	if (cfd == -1)
 		goto err_out3;
 
-	if (fuse_session_custom_io(se, &io, cfd) != 0)
+	if (fuse_session_custom_io(se, &io, sizeof(io), cfd) != 0)
 		goto err_out3;
 
 	/* Block until ctrl+c */
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -2043,7 +2043,10 @@ struct fuse_session *fuse_session_new(struct fuse_args *args,
  * @param se session object
  * @param io Custom io to use when retrieving/sending requests/responses
  * @param fd file descriptor for the session
- *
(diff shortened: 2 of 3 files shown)


ashleypittman (approved):
This looks good.

As I say it might be possible to support two versions in the same way we do at
https://github.com/libfuse/libfuse/blob/master/include/fuse_lowlevel.h#L1979-L1992 but I'd need to look into that.  This solves the problem for master and going forward at least.

bsbernd:
Yeah, we can make it versioned (ABI and API).

ashleypittman:
> Yeah, we can make it versioned (ABI and API). I cannot work on that right now, probably earliest tomorrow morning. I would also happily let you do the versioning part - if you want :)

sorry I missed this.   I'll have a go in the morning.

bsbernd:
Thanks Ashley! My patch to get the libfuse version passed into fuse_session_new is growing and got me busy for spare time over the weekend. Difficulty is the highlevel API. I hope I'm close to get that done, but then didn't have any time for this issue here.

legezywzh:
This patch looks good to me, thanks for your time. I didn't realize this compatibility issue before, sorry.

bsbernd:
@legezywzh The patch is unfortunately not complete - we need to ensure API and ABI compatibility and that makes it more complex.

legezywzh:
> @legezywzh The patch is unfortunately not complete - we need to ensure API and ABI compatibility and that makes it more complex.


So the correct solution is https://github.com/libfuse/libfuse/pull/935, right? I'll take a look.

bsbernd:
Yeah, but it is not complete yet.

bsbernd:
Already addressed here: https://github.com/libfuse/libfuse/pull/953

PR #935, opened 2024-04-23
Make the changes to custom_io backwards compabible.

ashleypittman:
Pass in a size field to session_custom_io so that new entries
can be added without breaking the ABI.

Make the function prototype backwards compatible.

Signed-off-by: Ashley Pittman <ashley.m.pittman@intel.com>

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,11 @@
+libfuse 3.17.x (tbd)
+====================
+
+When using -DFUSE_USE_VERSION=317 then
+* Added new fuse_session_custom_io_317 function to allow extra callbacks.
+
+* Use clone_fd callback in struct fuse_custom_io
+
 libfuse 3.16.2 (2023-10-10)
 ===========================
 
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -136,7 +136,10 @@ struct fuse_custom_io {
 	ssize_t (*splice_send)(int fdin, off_t *offin, int fdout,
 				     off_t *offout, size_t len,
 			           unsigned int flags, void *userdata);
+
+#if FUSE_USE_VERSION >= FUSE_MAKE_VERSION(3, 17)
 	int (*clone_fd)(int master_fd);
+#endif
 };
 
 /**
@@ -2042,6 +2045,7 @@ struct fuse_session *fuse_session_new(struct fuse_args *args,
  *
  * @param se session object
  * @param io Custom io to use when retrieving/sending requests/responses
+ * @param io_size sizeof(struct fuse_custom_io)
  * @param fd file descriptor for the session
  *
  * @return 0  on success
@@ -2050,8 +2054,14 @@ struct fuse_session *fuse_session_new(struct fuse_args *args,
  * @return -errno  if failed to allocate memory to store `io`
  *
  **/
+#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 17)
 int fuse_session_custom_io(struct fuse_session *se,
-				   const struct fuse_custom_io *io, int fd);
+			   const struct fuse_custom_io *io, int fd);
+#else
+int fuse_session_custom_io_317(struct fuse_session *se,
+			       const struct fuse_custom_io *io, size_t io_size, int fd);
+#define fuse_session_custom_io(_SE, _IO, _FD) fuse_session_custom_io_317(_SE, _IO, (sizeof(_IO)), _FD)
+#endif
 
 /**
  * Mount a FUSE file system.
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -3116,9 +3116,16 @@ out1:
 	return NULL;
 }
 
-int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io,
-			   int fd)
+int fuse_session_custom_io_317(struct fuse_session *se, const struct fuse_custom_io *io,
+			       size_t io_size, int fd)
 {
+
+	if (sizeof(struct fuse_custom_io) < io_size)
+	{
+		fuse_log(FUSE_LOG_ERR, "fuse: warning: library too old, some operations may not work\n");
+		io_size = sizeof(struct fuse_custom_io);
+	}
+
 	if (fd < 0) {
 		fuse_log(FUSE_LOG_ERR, "Invalid file descriptor value %d passed to "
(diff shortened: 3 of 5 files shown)


ChangeLog.rst
```diff
@@ -1,3 +1,11 @@
+libfuse 3.16.x (tbd)
+====================
```
bsbernd, line 2: Should be 3.17

include/fuse_lowlevel.h
```diff
@@ -2050,8 +2051,14 @@ struct fuse_session *fuse_session_new(struct fuse_args *args,
...
  *
  **/
+#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 17)
 int fuse_session_custom_io(struct fuse_session *se,
-				   const struct fuse_custom_io *io, int fd);
+			   const struct fuse_custom_io *io, int fd);
+#else
+int fuse_session_custom_io_317(struct fuse_session *se,
+			       const struct fuse_custom_io *io, size_t io_size, int fd);
+#define fuse_session_custom_io(_SE, _IO, _FD) fuse_session_custom_io_317(_SE, _IO, (sizeof(_IO)), _FD)
```
bsbernd, line 2063: This makes it API compatible, but redirects with macro to the fixed function name _317. To make it ABI compatible with versioned symbols, like this
```
int fuse_session_loop_mt_312(struct fuse_session *se, struct fuse_loop_config *config);
FUSE_SYMVER("fuse_session_loop_mt_312", "fuse_session_loop_mt@@FUSE_3.12")
int fuse_session_loop_mt_312(struct fuse_session *se, struct fuse_loop_config *config)
````

File: ChangeLog.rst (hunk 1 above)
Comment (ashleypittman) [reply ChangeLog.rst:2]: Yes, I'd already spotted that and fixed in the next push.

File: include/fuse_lowlevel.h (hunk 2 above)
Comment (ashleypittman) [reply include/fuse_lowlevel.h:2063]: I think this is probably correct as-is but I agree this should use versioned symbols, I'm still trying to get my head around them completely.

Comment (ashleypittman) [reply include/fuse_lowlevel.h:2063]: Are you sure?  The FUSE_SYMVER macro is not available to the public headers.

Comment (bsbernd) [reply include/fuse_lowlevel.h:2063]: This is a libfuse internal file, so it doesn't need to be in public headers. The public header basically has to define the API for 

```
#if LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS
   #if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3,17)
       int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io,
			   int fd);
  #else
       int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io,
			   size_t fuse_custom_io_sz, int fd);
  #endif
#else
   /* macro solution as you already did */
#endif
```

Comment (ashleypittman) [reply include/fuse_lowlevel.h:2063]: This is the public header.

Comment (bsbernd) [reply include/fuse_lowlevel.h:2063]: Ah now I see what you mean, FUSE_SYMVER() needs to be done inside of the c file, as with the other compat symbols.

File: lib/fuse_lowlevel.c [hunk 3]
```diff
@@ -3138,18 +3145,29 @@ int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io
...
 
 	se->fd = fd;
-	*se->io = *io;
+	memcpy(&se->io, io, io_size);
 	return 0;
 }
 
+#undef fuse_session_custom_io
+
+int fuse_session_custom_io(struct fuse_session *se,
```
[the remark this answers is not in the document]
legezywzh: > Just for your information - I'm not sure if needed here, but I had to fuse_parse_cmdline() to compat.c (for libc without versioned symbols) and the ABI symbol in the corresponding c file.

I wonder why fuse_session_loop_mt doesn't need same actions in lib/compat.c.

legezywzh:
@ashleypittman hi, would you like to push a new one according to bsbernd's comments, seems that the only not resolved issue is whether to do same actions as fuse_parse_cmdline in lib/compat.c

lib/fuse_lowlevel.c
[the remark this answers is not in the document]
bsbernd: See fuse_lowlevel.h
```
#if (defined(LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS))
int fuse_parse_cmdline(struct fuse_args *args,
               struct fuse_cmdline_opts *opts);
#else
#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
int fuse_parse_cmdline_30(struct fuse_args *args,
               struct fuse_cmdline_opts *opts);
#define fuse_parse_cmdline(args, opts) fuse_parse_cmdline_30(args, opts)
#else
int fuse_parse_cmdline_312(struct fuse_args *args,
               struct fuse_cmdline_opts *opts);
#define fuse_parse_cmdline(args, opts) fuse_parse_cmdline_312(args, opts)
#endif
#endif 

```

So for LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS no issue and compat.c not needed. But for the else branch we need to redirect fuse_parse_cmdline() for *API* to fuse_parse_cmdline_30 and fuse_parse_cmdline_312. For *ABI we still need to provide a  fuse_parse_cmdline *symbol* - if defined in the same header this clashes with the macro. I probably could have left fuse_parse_cmdline() inside of fuse_lowlevel.c, but I _think_ it is cleaner to have it in compat.c

legezywzh:
@bsbernd @ashleypittman if you don't have free time, do you agree that I take over this pr, do some rebase and rework according to current patch, your previous discussions and libfuse code base?

bsbernd:
@legezywzh I would have time for it earliest next week. You might want to look at my recent commit (58f85bfa9b7dca9a216cd0bb4e38e9cdf4b661da), which also adds new compat symbols.

ashleypittman:
> @bsbernd @ashleypittman if you don't have free time, do you agree that I take over this pr, do some rebase and rework according to current patch, your previous discussions and libfuse code base?

No issue from me.  Other than the fact this PR needs merging with master I believe it's right from a C perspective but as per the comments needs extra handling for versioned symbols.  This isn't something I've needed to touch before so I need to come up to speed on this which is taking me a while to be able to schedule some time for.  If you want to take over this then I have no issue with it.

legezywzh:
@bsbernd @ashleypittman since this compatibility issue is introduced by my patch :) I'll prepare a patch this week, thnaks.

PR #939, opened 2024-04-26
fuse_common.h: fix warning on _Static_assert()

CismonX:
When building a project that uses libfuse with musl libc, gcc gives the following warning if provided with the `-std=c99 -Wpedantic` options:

```
/usr/include/fuse3/fuse_common.h:938:1: warning: ISO C99 does not support '_Static_assert' [-Wpedantic]
  938 | _Static_assert(sizeof(off_t) == 8, "fuse: off_t must be 64bit");
      | ^~~~~~~~~~~~~~
```

Note: not reproducible with glibc, since glibc's `sys/cdefs.h` (which is included by `features.h`, and the latter is included by some system headers, e.g. `sys/types.h`, used by libfuse) defines `_Static_assert` when it's not available:

```C
#if (!defined _Static_assert && !defined __cplusplus \
     && (defined __STDC_VERSION__ ? __STDC_VERSION__ : 0) < 201112 \
     && (!(__GNUC_PREREQ (4, 6) || __clang_major__ >= 4) \
         || defined __STRICT_ANSI__))
# define _Static_assert(expr, diagnostic) \
    extern int (*__Static_assert_function (void)) \
      [!!sizeof (struct { int __error_if_negative: (expr) ? 2 : -1; })]
#endif
```

Since `_Static_assert` is introduced in C11, it would be better to use a more standard-conformant way to check for its availability.

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -957,7 +957,7 @@ void fuse_loop_cfg_convert(struct fuse_loop_config *config,
  * On 32bit systems please add -D_FILE_OFFSET_BITS=64 to your compile flags!
  */
 
-#if defined(__GNUC__) && (__GNUC__ > 4 || __GNUC__ == 4 && __GNUC_MINOR__ >= 6) && !defined __cplusplus
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
 _Static_assert(sizeof(off_t) == 8, "fuse: off_t must be 64bit");
 #else
 struct _fuse_off_t_must_be_64bit_dummy_struct \


bsbernd:
Thanks for pointing this out, I would prefer to enforce C11 https://github.com/libfuse/libfuse/pull/940. Can you please check if that works for you?

bsbernd:
Fixed by using gnu11.

bsbernd:
@CismonX before going this direction I would first like to understand if there is a real need for it or if this a theoretical request? I don't say we should right now already enforce C23, but C11 is 13 years old..
Which stack still does depend on C99 and with that going further, why C99 and not C89...

If wanted to go that C99 direction, if would Static_assert would need to be defined like FUSE_STATIC_ASSERT() and become a generic macro - I wouldn't like to have such ifdefs all over the  code.
Furthermore, I also recently added _Atomic - which is C11 as well. Means if we would like to avoid C11, we would need to have a full compat atomics part as well - test if compiler intrinsic work and if not fall back to pthread locking. Personally I wouldn't like to spend my time on that. If you would like to do that work - other projects like libfabric have it and it might be just a copy over - but needs to check if wouldn't violate the license (libfabric is GPL/BSD, I think).

bsbernd:
Hmm, ok, I'm kind of fine to avoid C11 in public headers. Now the issues is that specific line is in there since 2012. And what gives me even more headache is that this specific assert is responsible to avoid data corruption. I.e. a 32bit project forgets to set -D_FILE_OFFSET_BITS=64 (see for example recent activity in unionfs-fuse) and now let's assume this project also enforces C99 - data corruption.
For me that means we cannot just disable the flag is a project wants C99, i.e. in this specific case you would need to find a C99 replacement for the assert first.

ashleypittman:
For me that means we cannot just disable the flag if a project wants C99, i.e. in this specific case you would need to find a C99 replacement for the assert first.

Does the code not already have this?

Disabling 12 year old code in order to support a 25 year old standard isn't ideal but it does sound like this PR would solve a problem and the current check of checking an ancient gcc versions isn't correct either.

I haven't checked if we have one but it would be good to have a test that we're not losing any c++ support with this PR.

bsbernd:
> For me that means we cannot just disable the flag if a project wants C99, i.e. in this specific case you would need to find a C99 replacement for the assert first.
> 
> Does the code not already have this?

Oh right I had missed the 'else'.

> Disabling 12 year old code in order to support a 25 year old standard isn't ideal but it does sound like this PR would solve a problem and the current check of checking an ancient gcc versions isn't correct either.

Given there is already a fall back, fine with me.

> I haven't checked if we have one but it would be good to have a test that we're not losing any c++ support with this PR.

Maybe I read it wrong again in the evening, but it says `&& !defined __cplusplus`, i.e. currently incorrectly disables C++ (C++11 has the assert). Maybe something like

```#if defined (__cpp_static_assert) || (defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L))```

(untested)

bsbernd:
I have not tested it yet, but `__cpp_static_assert` should also work 
https://en.cppreference.com/w/cpp/feature_test#cpp_static_assert

And yeah, a bit a pity that C++ and C syntax differ (fixed in C23).

bsbernd:
@CismonX thanks for testing it! I guess you are right and we should use `__cplusplus`. Do you want to submit an updated patch or shall I merge this one and update it on my own?

PR #942, opened 2024-04-28
Add in the libfuse version a program was compiled with

bsbernd:
The API stays the same, the libfuse version comes from inlined functions, which are defined fuse_lowlevel.h and fuse.h.

fuse_session_new() -> static inlinei, in the application _fuse_session_new -> inside of libfuse

fuse_new() -> static inline, in the application
_fuse_new() -> inside of libfuse

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -854,6 +854,22 @@ struct fuse_context {
 	mode_t umask;
 };
 
+#if (defined(LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS))
+/**
+ * The real main function
+ *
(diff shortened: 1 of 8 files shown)


include/fuse.h
```diff
@@ -913,7 +913,20 @@ struct fuse_context {
   void *private_data);
 */
 #define fuse_main(argc, argv, op, private_data)				\
-	fuse_main_real(argc, argv, op, sizeof(*(op)), private_data)
+({									\
```
bsbernd, line 916: I guess we can also make this a static inline.

bsbernd:
At least according to nm seems to be correct.

Issue #114, opened 2016-11-10
Review default capabilities


Nikratio:
There are several capability flags that should always be enabled, and that are optional only to ensure backwards compatibility. For libfuse3, we should ensure that we enable them all by default, and maybe even remove some of them completely (i.e., force them to be enabled).

Nikratio:
Forcing them to be enabled would mean that either libfuse3 doesn't work with old kernels, or that file systems wouldn't get the same semantics when running on old kernels.

I think the best solution is to enable them by default, and checking that all requested flags are supported after `init()` returns. That way, filesystems get decent behavior by default, but can opt-in into supporting old semantics.

Nikratio:
Fixed in commit 08ce854a739d25a.

PR #949, opened 2024-05-10
Handle NO_OPEN/NO_OPENDIR support automatically

josefbacik:
If the file system doesn't provide a ->open or an ->opendir, and the kernel supports FUSE_CAP_NO_OPEN_SUPPORT or FUSE_CAP_NO_OPENDIR_SUPPORT, return ENOSYS so that the kernel no longer sends OPEN or OPENDIR requests, and properly caches the directories.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -1351,6 +1351,8 @@ static void do_open(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 
 	if (req->se->op.open)
 		req->se->op.open(req, nodeid, &fi);
+	else if (req->se->conn.want & FUSE_CAP_NO_OPEN_SUPPORT)
+		fuse_reply_err(req, ENOSYS);
 	else
 		fuse_reply_open(req, &fi);
 }
@@ -1507,6 +1509,8 @@ static void do_opendir(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 
 	if (req->se->op.opendir)
 		req->se->op.opendir(req, nodeid, &fi);
+	else if (req->se->conn.want & FUSE_CAP_NO_OPENDIR_SUPPORT)
+		fuse_reply_err(req, ENOSYS);
 	else
 		fuse_reply_open(req, &fi);
 }
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -325,8 +325,10 @@ struct fuse_loop_config_v1 {
  * kernel. (If this flag is not set, returning ENOSYS will be treated
  * as an error and signaled to the caller).
  *
- * Setting (or unsetting) this flag in the `want` field has *no
- * effect*.
+ * Setting this flag in the `want` field enables this behavior automatically
+ * within libfuse for low level API users. If non-low level users wish to have
+ * this behavior you must return `ENOSYS` from the open() handler on supporting
+ * kernels.
  */
 #define FUSE_CAP_NO_OPEN_SUPPORT	(1 << 17)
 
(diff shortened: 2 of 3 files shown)


lib/fuse_lowlevel.c
```diff
@@ -1351,6 +1351,8 @@ static void do_open(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 
 	if (req->se->op.open)
 		req->se->op.open(req, nodeid, &fi);
+	else if (req->se->conn.capable & FUSE_CAP_NO_OPEN_SUPPORT)
+		fuse_reply_err(req, ENOSYS);
 	else
```
bsbernd, line 1356: Hmm, I think we might get a bit of different behavior. Basically before we had a success if open/opendir was not supported. No idea if there are file systems out there that depend on or want that, if there are they might get broken by this change. 
I think the right way is to check for req->se->conn.want and then file systems need to set that in their -> init() method.

bsbernd (changes requested):

lib/fuse_lowlevel.c
josefbacik: The documentation here

https://libfuse.github.io/doxygen/structfuse__lowlevel__ops.html#afa15612c68f7971cadfe3d3ec0a8b70e

explicitly says if you don't provide ->opendir() (and the same exists for ->open as well) and we have FUSE_CAP_NO_OPENDIR_SUPPORT that we won't get subsequent ->opendir (or ->open) calls from the kernel.  I was very surprised when I was tracing things to see thousands of these calls coming in.

So if we want to make it so the filesystem has to set something that's fine, but the documentation has said this since this behavior has been there, and it's clear what the kernel intended the result to be.  It's a super annoying surprise for the documentation to say one thing but the behavior be something else.

Additionally, the code for req->se->conn.want does this

        if (se->conn.want & (~se->conn.capable)) {
                fuse_log(FUSE_LOG_ERR, "fuse: error: filesystem requested capabilities "
                        "0x%x that are not supported by kernel, aborting.\n",
                        se->conn.want & (~se->conn.capable));
                fuse_reply_err(req, EPROTO);

I want this to just work no matter what.  If the kernel doesn't support my capabilities then fine, do a normal open, but otherwise take advantage of this pretty significant performance win.

bsbernd: I understand your reasoning, but now let's assume a file system depends on the current behavior to just return success without any action? The file system would now out of the sudden need to create an open/opendir method and just return success. Especially missing opendir is rather common.

A rather important side effect of no-open / no-open-dir is that kernel side enables FOPEN_CACHE_DIR  / FOPEN_KEEP_CACHE for no_open / no_opendir.

So my point of view, if it would have been like this from the beginning, your patch as it is would have been right. But now it implies a change of behavior and documentation needs to be corrected?

bsbernd: This part 
```
if (se->conn.want & (~se->conn.capable)) {
            fuse_log(FUSE_LOG_ERR, "fuse: error: filesystem requested capabilities "
                    "0x%x that are not supported by kernel, aborting.\n",
                    se->conn.want & (~se->conn.capable));
            fuse_reply_err(req, EPROTO); 
```

is debatable - it makes file systems to have
```
if (conn->capable & FLAG)
    conn->want |= FLAG;
```

I wonder if that could be removed, I think kernel side just ignores unknown flags - I don't see why libfuse needs to be that strict here.

bsbernd: Looks like it comes from https://github.com/libfuse/libfuse/issues/114

```
Forcing them to be enabled would mean that either libfuse3 doesn't work with old kernels, or that file systems wouldn't get the same semantics when running on old kernels.

I think the best solution is to enable them by default, and checking that all requested flags are supported after init() returns. That way, filesystems get decent behavior by default, but can opt-in into supporting old semantics.
```

But I don't understand where the first paragraph comes from.

josefbacik: Ok agreed, it is a behavior change for what's been going on for years, I'll rework it your way and update all the documentation.  I will just continue to have fuse_reply_err(req, ENOSYS) in our implementation since there's no telling which version of libfuse we'll have at any given moment.

bsbernd (approved):
Thank you very much, LGTM.

PR #953, opened 2024-05-27
Fix compatibility issue around fuse_custom_io->clone_fd

legezywzh:
Fixes: 73cd124d0408 ("Add clone_fd to custom IO (#927)")

### DIFF
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -24,6 +24,7 @@
 
 #include "fuse_common.h"
 
+#include <stddef.h>
 #include <utime.h>
 #include <fcntl.h>
 #include <sys/types.h>
@@ -2092,6 +2093,13 @@ fuse_session_new(struct fuse_args *args,
 	return _fuse_session_new(args, op, op_size, &version, userdata);
 }
 
+/*
+ * This should mostly not be called directly, but instead the
+ * fuse_session_custom_io() should be used.
+ */
+int fuse_session_custom_io_317(struct fuse_session *se,
+			const struct fuse_custom_io *io, size_t op_size, int fd);
+
 /**
  * Set a file descriptor for the session.
  *
@@ -2119,8 +2127,20 @@ fuse_session_new(struct fuse_args *args,
  * @return -errno  if failed to allocate memory to store `io`
  *
  **/
-int fuse_session_custom_io(struct fuse_session *se,
-				   const struct fuse_custom_io *io, int fd);
+#if FUSE_MAKE_VERSION(3, 17) <= FUSE_USE_VERSION
+static inline int fuse_session_custom_io(struct fuse_session *se,
+					const struct fuse_custom_io *io, size_t op_size, int fd)
+{
+	return fuse_session_custom_io_317(se, io, op_size, fd);
+}
+#else
+static inline int fuse_session_custom_io(struct fuse_session *se,
+					const struct fuse_custom_io *io, int fd)
+{
+	return fuse_session_custom_io_317(se, io,
+				offsetof(struct fuse_custom_io, clone_fd), fd);
+}
+#endif
 
 /**
  * Mount a FUSE file system.
--- a/lib/compat.c
+++ b/lib/compat.c
@@ -20,6 +20,8 @@
 struct fuse_args;
 struct fuse_cmdline_opts;
 struct fuse_cmdline_opts;
+struct fuse_session;
+struct fuse_custom_io;
 
 
 /**
@@ -41,6 +43,17 @@ int fuse_parse_cmdline(struct fuse_args *args,
 		       struct fuse_cmdline_opts *opts)
 {
 	return fuse_parse_cmdline_30(args, opts);
+}
+
+int fuse_session_custom_io_30(struct fuse_session *se,
+				const struct fuse_custom_io *io, int fd);
+int fuse_session_custom_io(struct fuse_session *se,
+				const struct fuse_custom_io *io, int fd);
+int fuse_session_custom_io(struct fuse_session *se,
+			const struct fuse_custom_io *io, int fd)
+
+{
+	return fuse_session_custom_io_30(se, io, fd);
 }
 #endif
 
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -3190,9 +3190,15 @@ struct fuse_session *fuse_session_new_30(struct fuse_args *args,
 	return _fuse_session_new_317(args, op, op_size, &version, userdata);
 }
 
-int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io,
-			   int fd)
+FUSE_SYMVER("fuse_session_custom_io_317", "fuse_session_custom_io@@FUSE_3.17")
+int fuse_session_custom_io_317(struct fuse_session *se,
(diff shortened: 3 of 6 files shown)


include/fuse_lowlevel.h
```diff
@@ -2119,8 +2130,17 @@ fuse_session_new(struct fuse_args *args,
...
+#if (defined(LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS))
+/* Do not call this directly, but only through fuse_session_custom_io() */
+int _fuse_session_custom_io(struct fuse_session *se,
+			const struct fuse_custom_io *io,
+			size_t op_size, int fd);
+#else
+#define _fuse_session_custom_io(se, io, op_size, fd)	\
+	_fuse_session_custom_io_317(se, io, op_size, fd)
+#endif
+#define fuse_session_custom_io(se, io, fd)	\
```
bsbernd, line 2142: What is the purpose of this macro?  I had introduced the inlined function `fuse_session_new` which calls into `_fuse_session_new` to encode in the libfuse version an application was compiled against. _Maybe_ I should also do that for fuse_session_custom_io(). But right now this macro here does not bring an advantage?

bsbernd:
@legezywzh Thanks for working on this, looks good to me, just a comment about the "_" name

include/fuse_lowlevel.h
legezywzh: > What is the purpose of this macro? I had introduced the inlined function `fuse_session_new` which calls into `_fuse_session_new` to encode in the libfuse version an application was compiled against. _Maybe_ I should also do that for fuse_session_custom_io(). But right now this macro here does not bring an advantage?

Do you mean change to below:
```
@@ -2119,8 +2130,13 @@ fuse_session_new(struct fuse_args *args,
  * @return -errno  if failed to allocate memory to store `io`
  *
  **/ 
+#if (defined(LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS))
 int fuse_session_custom_io(struct fuse_session *se,
                   const struct fuse_custom_io *io, int fd);
+#else
+#define fuse_session_custom_io(se, io, fd) \
+   _fuse_session_custom_io_317(se, io, sizeof(*io), fd)
+#endif
```

But original fuse_session_custom_io() API doesn't support to pass sizeof(fuse_custom_io ) info, so for LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS branch, it won't work.

bsbernd: I only see
```
#if  (defined(LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS))
int _fuse_session_custom_io(struct fuse_session *se,...
...
#endif
#define fuse_session_custom_io(se, io, fd) _fuse_session_custom_io(se, io, sizeof(*io), fd)
```

I.e. redirection of `fuse_session_custom_io` to the underscored function

legezywzh: Oh, I think now I got what you mean.
Indeed we can change original fuse_session_custom_io() API to add sizeof(fuse_custom_io ) info for LIBFUSE_BUILT_WITH_VERSIONED_SYMBOLS branch, right?

legezywzh: Hmm, seems that I was wrong, API must not be changed, Could you please give some code hints.

bsbernd:
Ah right, what about making it an inlined function instead of a macro?

```
int fuse_session_custom_io_317(struct fuse_session *se, const struct fuse_custom_io *io, size_t op_size, int fd);
#if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 17)
static inline int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io, size_t op_size, int fd)
{
     return fuse_session_custom_io_317(se, io, op_size, fd);
}
#else
static inline int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io, int fd)
{
   return fuse_session_custom_io_317(se, io, offsetof(struct fuse_custom_io, clone_fd), fd);
}
#endif

```

There is also no underscore "_" and version check is "<" instead of "<=" as 3.17 is not released yet.

bsbernd:
(personally I prefer inlined functions over macros, though just my personal opinion)

legezywzh:
> Ah right, what about making it an inlined function instead of a macro?
> 
> ```
> int fuse_session_custom_io_317(struct fuse_session *se, const struct fuse_custom_io *io, size_t op_size, int fd);
> #if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 17)
> static inline int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io, size_t op_size, int fd)
> {
>      return fuse_session_custom_io_317(se, io, op_size, fd);
> }
> #else
> static inline int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io, int fd)
> {
>    return fuse_session_custom_io_317(se, io, offsetof(struct fuse_custom_io, clone_fd), fd);
> }
> #endif
> ```
> 
> There is also no underscore "_" and version check is "<" instead of "<=" as 3.17 is not released yet.


then there'll be compile error:
../lib/compat.c:59:5: error: redefinition of ‘fuse_session_custom_io’
   59 | int fuse_session_custom_io(struct fuse_session *se,
      |     ^~~~~~~~~~~~~~~~~~~~~~

See patch codes about compat.c

legezywzh:
> Ah right, what about making it an inlined function instead of a macro?
> 
> ```
> int fuse_session_custom_io_317(struct fuse_session *se, const struct fuse_custom_io *io, size_t op_size, int fd);
> #if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 17)
> static inline int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io, size_t op_size, int fd)
> {
>      return fuse_session_custom_io_317(se, io, op_size, fd);
> }
> #else
> static inline int fuse_session_custom_io(struct fuse_session *se, const struct fuse_custom_io *io, int fd)
> {
>    return fuse_session_custom_io_317(se, io, offsetof(struct fuse_custom_io, clone_fd), fd);
> }
> #endif
> ```
> 
> There is also no underscore "_" and version check is "<" instead of "<=" as 3.17 is not released yet.

Or do you mean that using this code snippet, we don't need to do versioned symbols check about fuse_session_custom_io()?

bsbernd:
We need versioned symbols for ABI, but for API I thought this code snippet would do. 

Regarding compat.c, this PR should fix your issue
https://github.com/libfuse/libfuse/pull/954

bsbernd:
I merged PR #954, you shouldn't get the compilation error after rebase.

legezywzh:
@bsbernd I have updated one new version, thanks for for your patience.

And I have one personal question, in the development of libfuse, API can be changed in the new libfuse version, right? It only requires that old ABI is still provided, and userspace filesystems codes need to adapt new API and re-compile.
Do I understand correctly?

bsbernd:
Sorry I didn't manage to look into this today (actually already past midnight for me now), will respond after getting some sleep.

legezywzh:
> Sorry I didn't manage to look into this today (actually already past midnight for me now), will respond after getting some sleep.


Really never mind, just take your time.

bsbernd (approved):

bsbernd:
@legezywzh Thank you, also for the updates! LGTM, merged.

PR #956, opened 2024-05-30
Add support for no_interrupt

yangyun50:
The function fuse_session_process_buf_int() would do much thing for FUSE_INTERRUPT requests, even there are no FUSE_INTERRUPT requests:

1. check every non-FUSE_INTERRUPT request and add these requests to the linked list(se->list) under a big lock(se->lock).
2. the function fuse_free_req() frees every request and remove them from the linked list(se->list) under a big lock(se->lock).

These operations are not meaningful when there are no FUSE_INTERRUPT requests, and have a great impact on the performance of fuse filesystem because the big lock for each request.

In some cases, FUSE_INTERRUPT requests are infrequent, even none at all. Besides, the user-defined filesystem may do nothing for FUSE_INTERRUPT requests.

And the kernel side has the option "no_interrupt" in struct fuse_conn. This kernel option can only be enabled (fc->no_interrupt = 1) by return ENOSYS in libfuse for the reply of FUSE_INTERRUPT request.  The kernel side code about this is in fuse_dev_do_write():
```      
        /* Is it an interrupt reply ID? */
        if (oh.unique & FUSE_INT_REQ_BIT) {
                __fuse_get_request(req);
                spin_unlock(&fpq->lock);

                err = 0;
                if (nbytes != sizeof(struct fuse_out_header))
                        err = -EINVAL;
                else if (oh.error == -ENOSYS)
                        fc->no_interrupt = 1;
                else if (oh.error == -EAGAIN)
                        err = queue_interrupt(req);

                fuse_put_request(req);

                goto copy_finish;
        }
```
But I don't find the code to enable the "no_interrupt" kernel option in libfuse.

So add the no_interrupt support in libfuse, and when this operaion is enabled:
1. remove the useless locking operaions and list operations.
2. return ENOSYS for the reply of FUSE_INTERRUPT request to inform the kernel to disable FUSE_INTERRUPT request.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -134,14 +134,20 @@ void fuse_free_req(fuse_req_t req)
 	int ctr;
 	struct fuse_session *se = req->se;
 
-	pthread_mutex_lock(&se->lock);
-	req->u.ni.func = NULL;
-	req->u.ni.data = NULL;
-	list_del_req(req);
-	ctr = --req->ctr;
-	fuse_chan_put(req->ch);
-	req->ch = NULL;
-	pthread_mutex_unlock(&se->lock);
+	if (se->conn.no_interrupt) {
+		ctr = --req->ctr;
+		fuse_chan_put(req->ch);
+		req->ch = NULL;
+	} else {
+		pthread_mutex_lock(&se->lock);
+		req->u.ni.func = NULL;
+		req->u.ni.data = NULL;
+		list_del_req(req);
+		ctr = --req->ctr;
+		fuse_chan_put(req->ch);
+		req->ch = NULL;
+		pthread_mutex_unlock(&se->lock);
+	}
 	if (!ctr)
 		destroy_req(req);
 }
@@ -2786,7 +2792,13 @@ void fuse_session_process_buf_int(struct fuse_session *se,
 	err = ENOSYS;
 	if (in->opcode >= FUSE_MAXOP || !fuse_ll_ops[in->opcode].func)
 		goto reply_err;
-	if (in->opcode != FUSE_INTERRUPT) {
+	/* Do not process interrupt request */
+	if (se->conn.no_interrupt && in->opcode == FUSE_INTERRUPT) {
+		if (se->debug)
+			fuse_log(FUSE_LOG_DEBUG, "FUSE_INTERRUPT: reply to kernel to disable interrupt\n");
+		goto reply_err;
+	}
+	if (!se->conn.no_interrupt && in->opcode != FUSE_INTERRUPT) {
 		struct fuse_req *intr;
 		pthread_mutex_lock(&se->lock);
 		intr = check_interrupt(se, req);
--- a/example/cuse.c
+++ b/example/cuse.c
@@ -87,6 +87,14 @@ static int cusexmp_expand(size_t new_size)
 	return 0;
 }
 
+static void cusexmp_init(void *userdata, struct fuse_conn_info *conn)
+{
+	(void)userdata;
+
+	/* Disable the receiving and processing of FUSE_INTERRUPT requests */
+	conn->no_interrupt = 1;
+}
+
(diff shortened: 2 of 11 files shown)


lib/fuse_lowlevel.c
```diff
@@ -2863,7 +2876,8 @@ void fuse_lowlevel_help(void)
 	printf(
 "    -o allow_other         allow access by all users\n"
 "    -o allow_root          allow access by root\n"
-"    -o auto_unmount        auto unmount on process termination\n");
+"    -o auto_unmount        auto unmount on process termination\n"
+"    -o no_interrupt        do not process interrupt request");
 }
```
bsbernd, line 2881: I think that would be rather confusing for end users what this actually means. My suggestion is to add a function 
`fuse_session_disable_interrupts(struct fuse_session *se)` and then to set this in the file system after creating the session. What do you think?

bsbernd: And I definitely like the idea.

yangyun50: Thanks for your suggestions!
I think it's more reasonable to use the function `fuse_session_disable_interrupts(struct fuse_session *se)` since this `no_interrupt` option is related to the fuse file system implementation, not an option for end users. 
The code is updated according to your suggestions.

lib/cuse_lowlevel.c
```diff
@@ -319,6 +319,9 @@ struct fuse_session *cuse_lowlevel_setup(int argc, char *argv[],
 	if (res == -1)
 		goto err_sig;
 
+	/* Disable the receiving and processing of FUSE_INTERRUPT requests */
+	fuse_session_disable_interrupts(se);
+
```
bsbernd, line 324: This would entirely disable for cuse? I'm not sure if this is right. Maybe add a flag  `CUSE_NO_INTERRUPT` (see `CUSE_UNRESTRICTED_IOCTL`) and then set by `struct cuse_info::flags`?

bsbernd (commented):
Thanks for the updates! I'm asking for a change for cuse so that users can request to disable on their own.

yangyun50:
Thanks for your advice. I'm sorry for that I just thought cuse does not need FUSE_INTERRUPT requests, but it is wrong in fact.

To make code more unified, I think move the "enable no_interrupt option" action into the procedure of processing FUSE_INIT request is a not bad idea. 

So that cuse can also enable no_interrupt option just like other fuse file system in `do_init`. Besides, we do not need add new function `fuse_session_disable_interrupts(struct fuse_session *se)`  to the libfuse programmer.

What do you think?  If this is not a good idea, I can revert it. Thanks!

example/printcap.c
```diff
@@ -89,6 +89,9 @@ static void pc_init(void *userdata,
 			printf("\tFUSE_CAP_HANDLE_KILLPRIV_V2\n");
 	if(conn->capable & FUSE_CAP_DIRECT_IO_ALLOW_MMAP)
 			printf("\tFUSE_CAP_DIRECT_IO_ALLOW_MMAP\n");
+	/* Disable the receiving and processing of FUSE_INTERRUPT requests */
+	conn->no_interrupt = 1;
```
bsbernd, line 93: Not needed here, it just exits the session after that anyway.

bsbernd (commented):
Thank you @yangyun50. Looks indeed much better. Would you mind to remove no_interrupt from example/printcap.c? This is definitely not performance sensitive? Some other neither, but I guess people might look at these others as implementation example.

example/printcap.c
yangyun50: Okay. Yeah, I enable the `no_interrupt` option in example directory by default since they can be the implementation example for programmers. Besides, these example file systems don't use the interruption mechanism in the code. So disable it by default.

bsbernd (approved):

bsbernd:
Slight addition here, see the commit messages please. https://github.com/libfuse/libfuse/pull/957

PR #957, opened 2024-06-04
Atomic struct fuse_req::ref_cnt

bsbernd:
Make  `struct fuse_req::ctr` an atomic type and rename to ::ref_cnt`


bsbernd:
@yangyun50 would you mind to review this?

PR #958, opened 2024-06-04
fix useless fuse_init_intr_signal in high-level api

yangyun50:
The code `fuse_init_intr_signal` in `_fuse_new_317` is useless since commit 8ee553dac ("fuse_new(): don't accept options that don't make sense for end-users") has remove the conf.intr option for end-users. The conf.intr is always zero (i.e., disabled) here.

Move `fuse_init_intr_signal` after the user-defined `init()` function, so that conf.intr and conf.intr_signal has been configured by the user.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -2604,6 +2604,8 @@ void fuse_fs_init(struct fuse_fs *fs, struct fuse_conn_info *conn,
 		fs->user_data = fs->op.init(conn, cfg);
 }
 
+static int fuse_init_intr_signal(int signum, int *installed);
+
 static void fuse_lib_init(void *data, struct fuse_conn_info *conn)
 {
 	struct fuse *f = (struct fuse *) data;
@@ -2613,9 +2615,14 @@ static void fuse_lib_init(void *data, struct fuse_conn_info *conn)
 		conn->want |= FUSE_CAP_EXPORT_SUPPORT;
 	fuse_fs_init(f->fs, conn, &f->conf);
 
-	/* Disable the receiving and processing of FUSE_INTERRUPT requests */
-	if (!f->conf.intr)
+	if (f->conf.intr) {
+		if (fuse_init_intr_signal(f->conf.intr_signal,
+				&f->intr_installed) == -1)
+			fuse_log(FUSE_LOG_ERR, "fuse: failed to init interrupt signal\n");
+	} else {
+		/* Disable the receiving and processing of FUSE_INTERRUPT requests */
 		conn->no_interrupt = 1;
+	}
 }
 
 void fuse_fs_destroy(struct fuse_fs *fs)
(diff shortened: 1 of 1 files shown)


bsbernd (commented):
I think think we should change the API with 317 (i.e. FUSE_MAKE_VERSION(3, 17) and just let ->init() return an error code. These reserved bytes in fuse_common.h are precious...

bsbernd:
In general I also don't like too much failing in ->init() as the file system is already mounted at that time and failing results in a stale mount point.

yangyun50:
You are right! I'm sorry for that I just use the reserved bytes simply.
It's indeed not good for failing in ->init(). What do you think if it just prints a warning message when fuse_init_intr_signal failed in init() , since it's not a serious fault which need to close the whole filesystem?

bsbernd:
Sure, a log message should be good.

yangyun50:
I have updated the code and hope you can review it in your free time. @bsbernd

lib/fuse.c
```diff
@@ -2616,6 +2618,12 @@ static void fuse_lib_init(void *data, struct fuse_conn_info *conn)
 	/* Disable the receiving and processing of FUSE_INTERRUPT requests */
 	if (!f->conf.intr)
 		conn->no_interrupt = 1;
+
+	if (f->conf.intr &&
+	    fuse_init_intr_signal(f->conf.intr_signal,
+			          &f->intr_installed) == -1) {
+		fuse_log(FUSE_LOG_ERR, "fuse: failed to init interrupt signal\n");
+	}
```
bsbernd, line 2626: Maybe rewrite this a little bit and merge the `f->conf.intr` conditions?
```
if (f->conf.intr) {
    if ( fuse_init_intr_signal(f->conf.intr_signal,
			          &f->intr_installed) == -1)
        fuse_log(FUSE_LOG_ERR, "fuse: failed to init interrupt signal\n");
} else {
    conn->no_interrupt = 1;
}
```

bsbernd (commented):
Thank you, nice cleanup and fix! I think it would make sense to merge the f->conf.intr conditions in fuse_lib_init, wouldn't it?

lib/fuse.c
yangyun50: Thanks for your advice！The code is updated.

bsbernd (commented):
Almost there, I just opened your change in an editor and noticed you are using spaces instead of tabs. I need to find a way to add a github ci check for that...
Could you please change to tabs? If you should be using an editor that supports clang-format, you can use the project .clang-format file.

Sorry for spotting that late.

lib/fuse.c
```diff
@@ -2613,9 +2615,14 @@ static void fuse_lib_init(void *data, struct fuse_conn_info *conn)
 		conn->want |= FUSE_CAP_EXPORT_SUPPORT;
 	fuse_fs_init(f->fs, conn, &f->conf);
 
-	/* Disable the receiving and processing of FUSE_INTERRUPT requests */
-	if (!f->conf.intr)
+	if (f->conf.intr) {
+		if (fuse_init_intr_signal(f->conf.intr_signal,
+				&f->intr_installed) == -1)
+		fuse_log(FUSE_LOG_ERR, "fuse: failed to init interrupt signal\n");
```
bsbernd, line 2621: Thanks for the update! The fuse_log() line needs an indentation (tab).

bsbernd (approved):

PR #963, opened 2024-06-07
libfuse: use --direct-io for passthrough_hp with fstests

josefbacik:
generic/095 uses a fio job that does directio, however we constantly return -EIO if we're using the passthrough functionality and don't have --direct-io set.  This causes odd problems with the test because of a bug in fio, so update the helper to use --direct-io because we do in fact want to test direct io properly.


bsbernd:
Hi Josef, 

could you please check this PR? https://github.com/libfuse/libfuse/pull/964

I'm actually testing all my kernel patches with different options using a custom script, that links different local.configs.

josefbacik:
Closing this, #964 is a better solution.

PR #964, opened 2024-06-07
xfstests: Add the possibility for extra daemon options

bsbernd:
We need to test w/o page cache, w/o passthrough and other options. I.e. we need to have the possibility to add these options via local.config


josefbacik (approved):
This approach works for me, I'll close mine.

PR #965, opened 2024-06-07
libfuse: have passthrough_hp return an error with invalid mount options

josefbacik:
generic/003 fails currently because if you specify -o relatime we will fail to mount, but not return an error, so the test thinks that we support relatime and then hilarity ensues.  Set ret so that if we get any failures while trying to mount we will properly error out.


bsbernd (approved):
Ah nice, so simple. I had added all the different *time mount options, but Nikolaus/Miklos didn't like it too much as fuse server is responsible. Your solution is much better, thank you!

PR #966, opened 2024-06-13
libfuse: null-terminate buffer in fuse_req_getgroups()

szmi:
After reading the file /proc/$PID/task/$PID/status the buffer wasn't terminated with a null character.  This could theoretically lead to buffer overrun by the subsequent strstr() call.

Since the contents of the proc file are guaranteed to contain the pattern that strstr is looking for, this doesn't happen in normal situations.

Add null termination for robustness.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -3353,6 +3353,7 @@ retry:
 		goto retry;
 	}
 
+	buf[ret] = '\0';
 	ret = -EIO;
 	s = strstr(buf, "\nGroups:");
 	if (s == NULL)


lib/fuse_lowlevel.c
```diff
@@ -3353,6 +3353,7 @@ int fuse_req_getgroups(fuse_req_t req, int size, gid_t list[])
 		goto retry;
 	}
 
+	buf[ret] = '\0';
```
bsbernd, line 3356: Thanks, we need to add in support for FUSE_CREATE_SUPP_GROUP to be able to deprecate this function.

bsbernd (approved):

lib/fuse_lowlevel.c
szmi: That's not quite enough to deprecate it, since FUSE_CREATE_SUPP_GROUP only sends a single group currently, and only in case of creates.

But now that the API infrastructure is there it would not be too hard to send all groups in all cases.  The question is whether that's useful for anybody or not.

bsbernd: I also see
```
/*
 * Code shared between mknod, mkdir, symlink and link
 */
static int create_new_entry(struct fuse_mount *fm, struct fuse_args *args,
                            struct inode *dir, struct dentry *entry,
                            umode_t mode)
```

But yeah, plain open is missing. For our file systems we typically fetch groups on the server side to avoid large rpcs - I guess we could add it for open and others when someone needs it.
What we could do on adding FUSE_CREATE_SUPP_GROUP to libfuse, is to get the groups from the request when it has it and only otherwise from /proc.

PR #974, opened 2024-06-23
add fmask and dmask options

gandalfs-cat:
dmask: umask applied to directories
fmask: umask applied to non-directories
see https://github.com/libfuse/libfuse/issues/975

to get "typical" permission bits for regular files (0644) and directories (0755), a single umask option is not sufficient (or well, it isn't the way fuse implements it)

there is precident for separate umask and dmask options in other filesystems (see for example fat: https://github.com/torvalds/linux/tree/master/fs/fat)

this addition should not affect backward-compatibility; the original umask option retains the same meaning, but non-zero fmask or dmask will override it.

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -298,6 +298,14 @@ struct fuse_config {
 	int show_help;
 	char *modules;
 	int debug;
+
+	/**
+	 * `fmask` and `dmask` function the same way as `umask`, but apply
+	 * to files and directories separately. If non-zero, `fmask` and
+	 * `dmask` take precedence over the `umask` setting.
+	 */
+	unsigned int fmask;
+	unsigned int dmask;
 };
 
 
--- a/doc/mount.fuse3.8
+++ b/doc/mount.fuse3.8
@@ -181,8 +181,8 @@ This option is an alternative to
 cached data is invalidated on \fBopen\fP(2) if the modification
 time or the size of the file has changed since it was last opened.
 .TP
-\fBumask=M\fP
-Override the permission bits in \fIst_mode\fP set by the filesystem. The resulting permission bits are the ones missing from the given umask value.  The value is given in octal representation.
+\fBumask=M fmask=M dmask=M\fP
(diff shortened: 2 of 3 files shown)


include/fuse.h
```diff
@@ -113,9 +113,13 @@ struct fuse_config {
 	/**
 	 * If `set_mode` is non-zero, the any permissions bits set in
 	 * `umask` are unset in the st_mode attribute of each file.
+	 * `fmask` and `dmask` may be used instead of `umask` to affect
+	 * files and directories separately, respectively.
 	 */
 	int set_mode;
 	unsigned int umask;
+	unsigned int fmask;
+	unsigned int dmask;
```
bsbernd, line 122: I think this is going to break the ABI (we currently have no test for that, but I'm working on that). Please see this PR https://github.com/libfuse/libfuse/pull/978

bsbernd (changes requested):
Except of the ABI breakage looks good to me. Could you please also add something like "high-level: " in your commit message?

include/fuse.h
gandalfs-cat: i interpreted "The remaining options are used by libfuse internally and should not be touched" to imply "not part of the abi" and accordingly placed fmask and dmask before those fields as opposed to at the literal end of the structure; is that correct? or shall i move them to the literal end?

bsbernd: Yeah it is confusing, I'm changing that in my PR. And also asked for review from other maintainers for PR#978.
From my point of view your additions have to go below as otherwise values are shifted by 64 bit.

bsbernd:
Thanks, updated patch looks better, let's see if @ashleypittman or @trapexit respond to my review request for  PR #978 
Could you please also remove "Fixup" from the commit message - this is really a new feature and not a bug fix.

bsbernd:
Ah, now I see, you prefer "Squash and merge", while I prefer force pushes to update previous commits.

gandalfs-cat:
ah i thought it might be tidier/easier to follow the conversation that way but, reviewers preference wins :]

bsbernd:
No further reviews or annotations - merged.

PR #981, opened 2024-06-26
Use a single bit for "no_interrupt"

bsbernd:
This is an addition to commit cef8c8b24
"Add support for no_interrupt"
and uses a single bit - the variable is used as boolean.

This should allow us to add more booleans into the "unsigned" in the future. Though I'm not sure yet how to handle "$arch" that does not have unsigned as 32bit value.

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -642,7 +642,7 @@ struct fuse_conn_info {
 	 * 2) Return ENOSYS for the reply of FUSE_INTERRUPT request to
 	 * inform the kernel not to send the FUSE_INTERRUPT request.
 	 */
-	unsigned no_interrupt;
+	unsigned no_interrupt : 1;
 
 	/**
 	 * For future use.


bsbernd:
@yangyun50 Could you please review?

include/fuse_common.h
```diff
@@ -642,7 +642,7 @@ struct fuse_conn_info {
 	 * 2) Return ENOSYS for the reply of FUSE_INTERRUPT request to
 	 * inform the kernel not to send the FUSE_INTERRUPT request.
 	 */
-	unsigned no_interrupt;
+	unsigned no_interrupt : 1;
```
ashleypittman, line 645: Does this want, or need padding here?  There's no need to keep the struct small from a memory usage standpoint but it's part of the API and having a bit-field allow expansion whilst maintaining backwards compatibility would be useful.

```suggestion
	unsigned no_interrupt : 1;
	
	/* For future use */
	unsigned reserved_b : 31;
```

yangyun50: Thank you for your advice. I think it‘s better to pad here for backward compatibility.  But as @bsbernd  said, some "$arch" may not have unsigned type as 32bit. So I think this may be more general:
```suggestion
	unsigned no_interrupt : 1;
	
	/* For future use */
	unsigned reserved_b : sizeof(unsigned) * 8 - 1;
```

yangyun50:
> @yangyun50 Could you please review?

Sure. Thanks for your request! I have commented on @ashleypittman's review. And you can give some advice.

include/fuse_common.h
bsbernd: Sorry for late reply, I was busy with other things.

We can pad, but there is an intrinsic padding here, as an unsigned is 'consumed', with this change just one bit of it.
I really don't like that this struct is full of 'unsigned' - we don't know the exact number of bits.
I wonder if we should just break ABI for weird archs and change all of these unsigned values to uint32_t (in all structs then).

yangyun50: I have found that some fields (type unsigned) in `struct fuse_conn_info` is assigned to an uint32_t type value (`se->conn.proto_major = arg->major` in function `do_init() : line 1986` for exampe). The  `struct fuse_init_in` and  `struct fuse_init_out`, which have similar members  with `struct fuse_conn_info`, but use uint32_t type.

Besides, I think it can guarantee the compatibility functionally for archs which have 16-bit，64-bit，or any-other bit unsigned type for now. Because we can set the `FUSE_KERNEL_MINOR_VERSION` to make the 16-bit unsigned type to work and we don't use the higher bits (i.e., 33th, 34th and 64th bit)  of 64-bit unsigned  for now.

bsbernd:
Closing this one here, part of this patch series, which also changes to int32_t/uint32_t:
https://github.com/libfuse/libfuse/pull/1081

PR #983, opened 2024-07-02
Allow to have page aligned writes

bsbernd:
Read/writes IOs should be page aligned as fuse server
might need to copy data to another buffer otherwise in
order to fulfill network or device storage requirements.

Simple reproducer is example/passthrough*
and opening a file with O_DIRECT - without this change
writing to that file failed with -EINVAL if the underlying
file system was using ext4 (for passthrough_hp the
'passthrough' feature has to be disabled).

### DIFF
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -188,6 +188,8 @@ void cuse_lowlevel_init(fuse_req_t req, fuse_ino_t nodeide, const void *inarg);
 
 int fuse_start_thread(pthread_t *thread_id, void *(*func)(void *), void *arg);
 
+void fuse_buf_free(struct fuse_buf *buf);
+
 int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
 				 struct fuse_chan *ch);
 void fuse_session_process_buf_int(struct fuse_session *se,
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -190,10 +190,12 @@ int fuse_start_thread(pthread_t *thread_id, void *(*func)(void *), void *arg);
 
 void fuse_buf_free(struct fuse_buf *buf);
 
-int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
-				 struct fuse_chan *ch);
-void fuse_session_process_buf_int(struct fuse_session *se,
-				  const struct fuse_buf *buf, struct fuse_chan *ch);
+int fuse_session_receive_buf_internal(struct fuse_session *se,
+				      struct fuse_buf *buf,
+				      struct fuse_chan *ch);
+void fuse_session_process_buf_internal(struct fuse_session *se,
+				       const struct fuse_buf *buf,
+				       struct fuse_chan *ch);
 
 struct fuse *fuse_new_31(struct fuse_args *args, const struct fuse_operations *op,
 		      size_t op_size, void *private_data);
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2740,10 +2740,11 @@ static int fuse_ll_copy_from_pipe(struct fuse_bufvec *dst,
 void fuse_session_process_buf(struct fuse_session *se,
 			      const struct fuse_buf *buf)
 {
-	fuse_session_process_buf_int(se, buf, NULL);
+	fuse_session_process_buf_internal(se, buf, NULL);
 }
 
-void fuse_session_process_buf_int(struct fuse_session *se,
+/* libfuse internal handler */
+void fuse_session_process_buf_internal(struct fuse_session *se,
 				  const struct fuse_buf *buf, struct fuse_chan *ch)
 {
 	const size_t write_header_size = sizeof(struct fuse_in_header) +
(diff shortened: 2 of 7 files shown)


bsbernd:
Note: This is a draft until the kernel feature is merged.

trapexit:
Happy to see this get kernel support but I solved this some time ago (for my purposes anyway) but allocating aligned buffers that were large enough such that I could offset the reads from fuse such that the header would be page 1 alignment minus hdr size so that the data was written starting page aligned.

If I understand the patch, the kernel will put the header in the first page and data in second+?

bsbernd:
Hmm yeah, that is an interesting alternative you are using. 

Yep, kernel right now just puts data after the headers, with this patch is it moves forward data to the next page.

ashleypittman:
> I solved this some time ago (for my purposes anyway) but allocating aligned buffers that were large enough such that I could offset the reads from fuse such that the header would be page 1 alignment minus hdr size so that the data was written starting page aligned.

Nice.  I'd like to see details of that if you are able to publish anything.

trapexit:
> Hmm yeah, that is an interesting alternative you are using.

It was the only one that could work given the situation :) I kinda just figured it wasn't an issue for others since it was originally designed as it was and hadn't seen much conversation on it so didn't look into a kernel solution.

@ashleypittman The code is in my mergerfs project but it is really quite simple. I always allocate fuse recv buffers of max_pages + 1 and then add pagesize and subtract sizeof(struct fuse_in_header) + sizeof(struct fuse_write_in). For cleaning up you need to keep track of the original pointer returned by posix_memalign or whatnot or calculate it. In my case I want to keep the size of usable space so I have a simple struct { char *mem; u64 size; }; that is actually stored at the beginning of the allocation.

bsbernd:
@trapexit  yeah, if libfuse wouldn't support splice we could take that approach, but with splice it doesn't work

bsbernd:
@trapexit What do think about these changes? Should be basically what you do in mergerfs. Miklos didn't want to merge the patch that adds alignment in kernel, as one can do the alignment trick in userspace. In the io-uring branch there is a header/payload separation that I will probably also add to legacy /dev/fuse io, but that is for the future.
@joannekoong would you mind to review as well?

bsbernd:
@szmi if you have time, please review as well.

bsbernd:
Btw, the ABI test fails here, but I think we can ignore it, as the extended part is only used by libfuse internal functions.

szmi (approved):
ACK. I don't see any issues with offsetting the start of the buffer with the write header.

bsbernd:
A bit more complex now, as `fuse_session_receive_buf` is an exported function and there is no guarantee callers would free the right pointer.

bsbernd:
@trapexit as I wrote, header/payload separation is going to come. Patches in my uring branch are mess, because I don't want to spend time on cleaning up before we have a fixed kernel interface (although it should be now), but anyway take a look here:

https://github.com/bsbernd/libfuse/commit/0b9ee3e3d8486eb5a33682eb117eb5579f0700d0

The same can be done for plain /dev/fuse, libfuse just needs to send an iovec instead of a single buffer. I don't make promises yet when I find time to work on it, though.

bsbernd:
Personally I prefer to keep the extra pointer, but majority prefers, I can also do the calculation on free.

bsbernd:
Actually cleaner without the extra pointer and a new internal function `fuse_buf_free` - caught an issue with that in my previous patch.

lib/fuse_lowlevel.c
```diff
@@ -3137,6 +3183,18 @@ int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
 	return res;
 }
 
+int fuse_session_receive_buf(struct fuse_session *se, struct fuse_buf *buf)
+{
+	return _fuse_session_receive_buf(se, buf, NULL, false);
+}
+
+/* libfuse internal handler */
+int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
```
trapexit, line 3192: `int` as in `internal`? perhaps verbose but I would suggest the full word for clarity.

lib/fuse_i.h
```diff
@@ -188,8 +188,11 @@ void cuse_lowlevel_init(fuse_req_t req, fuse_ino_t nodeide, const void *inarg);
...
 int fuse_start_thread(pthread_t *thread_id, void *(*func)(void *), void *arg);
 
-int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
-				 struct fuse_chan *ch);
+void fuse_buf_free(struct fuse_buf *buf);
+
+int fuse_session_receive_buf_internal(struct fuse_session *se,
+				      struct fuse_buf *buf,
+				      struct fuse_chan *ch);
 void fuse_session_process_buf_int(struct fuse_session *se,
```
trapexit, line 196: Sorry, I missed that process_buf was also `int`. You were being consistent. In that case probably both need to be `int` or `internal`. My preference is the latter though I guess... is there a non-internal variant?

bsbernd: @joannekoong had recently changed the internal callers to the non-internal version. Initially looked ok as it simplified the code a bit, but in this PR we see that we need to use the internal functions for libfuse internal calls. The non `_internal` functions are defined the public API (`include/`) and are also exported in `lib/fuse_versionscript`- I'm personally not so happy to have `fuse_session_process_buf`` and `fuse_session_receive_buf` in the public API - no idea what is using that.

I created a new patch for the rename of _int to _internal, instead of having that in the main part. I'm a bit used for fn_int(), my colleagues tend to do the same, but I personally find it also confusing and typically use `_fn` and `__fn`

trapexit (approved):

bsbernd:
@joannekoong I already merged, please let me know if you have concerns.

bsbernd:
Thanks @trapexit and @szmi  for your review.

PR #984, opened 2024-07-02
Support max writes larger than the default limit

joannekoong:
Currently, the max write size for a request is limited in the kernel by FUSE_MAX_MAX_PAGES. There is a patch upstream [1] for enabling this limit to be dynamically configured through a sysctl. The motivating factor behind this is for improving performance for large workloads (some benchmarks can be found in the discussion in [2]).

In libfuse, users can request the max write size through setting "conn->max_write" in their init callback. However, this will fail for sizes larger than the default max write size (FUSE_MAX_MAX_PAGES * PAGESIZE) because the buffer allocated for receiving requests (in fuse_session_receive_buf_int) is allocated with size se->bufsize which is set to the default max write size (since at the time of buffer allocation, the init callback has not been called, so we do not know what max write size the user requests).

To support max write sizes larger than the default size, this commit adds an option -o "extended_max_write=" which allocates the request buffer at the requested max write size. This option can only be used for writes exceeding the default max write size (for writes less than the default max write size, users should set "conn->max_write" in their init callback). If for some reason the user sets both conn->max_write and uses extended_max_write, the extended_max_write value will be used and passed to the kernel as the requested max write size. The user is responsible for ensuring that the kernel they are running on has a fuse max pages limit that supports their requested max write size.

Another alternative that was explored is transparently reallocating the buffer for the user within libfuse if in the init callback the max_write size is larger than the default. This eliminates having to pass the write size through an option and users can just set it through "conn->max_write". Unfortunately, this has three issues: 1) max write sizes larger than the default would only be supported when users do not pass in a user-allocated fuse_buf 2) there is a race condition where if the user calls fuse_session_loop_mt, then fuse_session_receive_buf, and then fuse_session_process_buf, the fuse_session_loop_mt can fail if the init request is picked up by the call to fuse_session_receive_buf, before the fuse_session_loop_mt thread and 3) reallocating the buffer violates the "const"-ness of the fuse_session_process_buf public API.

[1] https://lore.kernel.org/linux-fsdevel/20240702014627.4068146-1-joannelkoong@gmail.com/T/#u
[2] https://lore.kernel.org/linux-fsdevel/20240124070512.52207-1-jefflexu@linux.alibaba.com/T/#u

Signed-off by: Joanne Koong <joannelkoong@gmail.com>

### DIFF
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -70,6 +70,11 @@ struct fuse_session {
 	 * a later version, to 'fix' it at run time.
 	 */
 	struct libfuse_version version;
+
+	/* Used for max writes larger than the default limit of
+	 * FUSE_MAX_MAX_PAGES * PAGE_SIZE.
+	 */
+	unsigned int extended_max_write;
 };
 
 struct fuse_chan {
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2125,6 +2125,9 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 	if (se->op.init)
 		se->op.init(se->userdata, &se->conn);
 
+	if (se->extended_max_write)
+		se->conn.max_write = se->extended_max_write;
+
 	if (se->conn.max_write > bufsize - FUSE_BUFFER_HEADER_SIZE)
 		se->conn.max_write = bufsize - FUSE_BUFFER_HEADER_SIZE;
 
@@ -2858,6 +2861,7 @@ static const struct fuse_opt fuse_ll_opts[] = {
 	LL_OPTION("-d", debug, 1),
 	LL_OPTION("--debug", debug, 1),
 	LL_OPTION("allow_root", deny_others, 1),
+	LL_OPTION("extended_max_write=%u", extended_max_write, 0),
 	FUSE_OPT_END
 };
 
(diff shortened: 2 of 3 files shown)


bsbernd:
Thanks,  I had seen your kernel patch. Let's first see Mikos' opinion about your kernel patch, so far Miklos always resisted against increasing max-pages. I don't have a strong opinion, as I don't see a perf improvement with my uring patches beyond 128KB IO - it basically avoids kernel/userspace transitions due to the ring. But then yeah, if you want to send large IOs over network or hardware device, one needs to sort (ring) requests - not impossible, but a bit cumbersome - your patch would allow to avoid that.
Let's first wait for Miklos and if the kernel patch gets approved we can think about libfuse solution. One way I see would be to handle the init request differently from other request and threads.

joannekoong:
Sounds great, I will wait and see what Miklos says. Thanks.

ashleypittman:
I support larger buffer sizes and we do see increased bandwidth with them so this functionality is great.

I also think getting the kernel support right first is important though and then we know the interface that userspace needs to work to and this has the potential to complicate the libfuse -> application interface significantly.  It may be useful to have a upper build as well we a current max size so that an application would know what size to allocate, however this then leads to extra memory usage potentially so it maybe not ideal.  Applications should ideally be able to make use of changes in kernel configuration without having to remount filesystems.

Realloc of buffers in libfuse is a bad option in my opinion, in fact I'd like to have the option of moving all buffer management to the application so it can manage locality and network addressability outside of libfuse.

bsbernd:
@ashleypittman The issue (without my ring patches) is that the buffer belongs to the thread and not the request. With the ring patches the application can access the queue buffer at startup and register it for networking or other rdma operations. We could fix it for legacy /dev/fuse io, but I don't have the time to work on code that is so much slower than what I already have...

ashleypittman:
@bsbernd  I'm not asking you to do any extra here, I'd like to change the buffer management so it `can` be handed by the application rather than the thread so if we're making changes to support larger buffers I'd like to fix both issues at the same time - once we have clarity on what the kernel interface will look like.

joannekoong:
Deprecating this in favor of a cleaner solution here https://github.com/libfuse/libfuse/pull/1041

PR #985, opened 2024-07-04
example/notify_inval_entry.c: minor fixes in comments

desertwitch:
- Notifications should be enabled in one example command, but present command would disable them.
- Arguments passed to example command are better readable when following the same format (subjective).

I came across these irregularities reading up on `fuse_lowlevel_notify_inval_entry()` , at least the way I understood.
Feel free to modify or not merge in case the misunderstanding is on my side, thanks for your efforts.


bsbernd (approved):

bsbernd:
Thank you @desertwitch. Merged.

PR #987, opened 2024-07-05
passthrough_hp: include . and .. in readdir

joannekoong:
generic/401 fails currently because it checks that "." and ".." are listed as directory entries.

Include "." and ".." as listed directory entries in passthrough_hp's readdir implementation.

Signed-off by: Joanne Koong <joannelkoong@gmail.com>

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -747,15 +747,20 @@ static void do_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
             break; // End of stream
         }
         d->offset = entry->d_off;
-        if (is_dot_or_dotdot(entry->d_name))
-            continue;
 
         fuse_entry_param e{};
         size_t entsize;
         if (plus) {
-            err = do_lookup(ino, entry->d_name, &e);
-            if (err)
-                goto error;
+            if (is_dot_or_dotdot(entry->d_name)) {
+		/* fuse kernel ignores attributes for these and also does
+		 * not increase lookup count (see fuse_direntplus_link) */
+		e.attr.st_ino = entry->d_ino;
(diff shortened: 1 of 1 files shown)


example/passthrough_hp.cc
```diff
@@ -747,15 +747,18 @@ static void do_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
...
 
         fuse_entry_param e{};
         size_t entsize;
         if (plus) {
-            err = do_lookup(ino, entry->d_name, &e);
-            if (err)
-                goto error;
+            if (is_dot_or_dotdot(entry->d_name)) {
+		e.attr.st_ino = entry->d_ino;
+		e.attr.st_mode = entry->d_type << 12;
```
bsbernd, line 756: Would you mind to add a comment that fuse kernel does not refresh attributes and with that also does not increase nlookup? Without that comment I wasn't sure and had to look it up. Something like
/* fuse kernel ignores attributes for these and also does not increase lookup count */

bsbernd:
Thank you, another issue fixed! Would be great if you could add a comment, so that one knows why there is an exception.

example/passthrough_hp.cc
joannekoong: Absolutely, I will add that in right now.

bsbernd:
Thank you @joannekoong

joannekoong:
Thanks for the review @bsbernd. 

I just reran the fuse fstests and this fixes generic/471 as well. Right now, I'm seeing about 25 failures (running with passthrough_hp). I have an upstream fix for generic/434 and generic/184, which I will submit this week and cc you on, Bernd.

PR #988, opened 2024-07-09
xfstests: Add '-odev' to passthrough mount options

bsbernd:
generic/434 and generic/184 are testing device files and fail because fuse sets 'nodev' by default.


bsbernd:
@joannekoong Could you please test this?

joannekoong:
Oh nice, I didn't know the option '-odev' existed. This fix is better than what I was going to submit upstream. 

I just tested your changes and verified that it fixes generic/434 and generic/184.

Issue #982, opened 2024-07-02
overwrite the whitelist


staltux:
how i can ignore or overwrite the whitelist of filesystem types when using fusermount?
appimage stop working on my debian testing with the error fusermount: mounting over filesystem type 0x858458f6 is forbidden
but work in debian stable
i am the only user of the machine and can take the risk

probonopd:
This is the code in question:

https://github.com/libfuse/libfuse/blob/f88e08f34d2d4f398f23797707e1c50cd306e405/util/fusermount.c#L1103-L1151

bsbernd:
0x858458f6 is RAMFS, right? I guess you could create a patch and just add it in.

probonopd:
For it to help the OP, we'd also need it for libfuse2 since that is what most existing AppImages out there are using.

bsbernd:
Oh dear, another patch that wants to get back ported to 2.9 :( 
What is 'AppImages'?

probonopd:
https://en.wikipedia.org/wiki/AppImage

staltux:
> 0x858458f6 is RAMFS, right? I guess you could create a patch and just add it in.

yes, but i dont have ramfs mounted outside the "normal" /dev/shm and  /run/lock ish
maybe my compressed zram? but disabling and unloading the module do nothing



> For it to help the OP, we'd also need it for libfuse2 since that is what most existing AppImages out there are using.
unfortunately the two i am trying is using the old libfuse2




> This is the code in question:
> 
> https://github.com/libfuse/libfuse/blob/f88e08f34d2d4f398f23797707e1c50cd306e405/util/fusermount.c#L1103-L1151

i can recompile the lib and install it,but the moment my debian testing update things can start to get weird

staltux:
i opened a issue in the appimagekit and a maintener give me a workaround, the appimage use the /tmp and mounting with tmpfs instead of ramfs "fixed it"
is not ideal because tmpfs don't have limit on ram usage, but is a temporary tradeoff

julian-payne:
Seems this was solved with a patch for @staltux but is there actually a way to "ignore or overwrite the whitelist of filesystem types when using fusermount?" ?

Somewhat unrelated, but I'm trying to use s3 mountpoint inside of a docker compose where the mount destination is a Apple File System (Mac OSX) and I'm getting the same error as this post.

```
my-image      | fusermount: mounting over filesystem type 0x6a656a63 is forbidden
[...]

PR #992, opened 2024-07-14
Add RAMFS to fusermount permitted mount targets

bsbernd:
Addresses issue https://github.com/libfuse/libfuse/issues/982


Rondom:
Nitpick: Rest of list was sorted

PR #995, opened 2024-07-23
Gate #include <execinfo.h> behind #ifdef HAVE_BACKTRACE

NilsIrl:
#include <execinfo.h> is unnecessary if HAVE_BACKTRACE is not defined.

This increases portability for systems without execinfo.h


bsbernd:
Thank you, merged.

PR #1001, opened 2024-07-29
example/passthrough_hp: Fix . and .. readdir lookup count

bsbernd:
Commit 170edc6a8ef0 added dot and dotdot (. and ..) to readdir results, but introduced an issue when max number of entries was reached - lookup count must not be decreased without doing the lookup.
With ext4 as underlying file system readir seems to return . and .. at random offsets and randomly failed xfstests for me.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -735,6 +735,7 @@ static void do_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
     }
 
     while (1) {
+        bool did_lookup = false;
         struct dirent *entry;
         errno = 0;
         entry = readdir(d->dp);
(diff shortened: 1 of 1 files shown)


bsbernd:
@joannekoong Could you please review?

bsbernd:
I'm merging it, in order to simplify handling of my different branches. @joannekoong please still review, if you want to have something changed we can do it post merge.

example/passthrough_hp.cc
```diff
@@ -772,7 +774,7 @@ static void do_readdir(fuse_req_t req, fuse_ino_t ino, size_t size,
         if (entsize > rem) {
             if (fs.debug)
                 cerr << "DEBUG: readdir(): buffer full, returning data. " << endl;
-            if (plus)
+            if (did_lookup)
```
joannekoong, line 777: Thanks for fixing this. An alternative check could be `if (e.ino)` instead of needing the did_lookup variable.

joannekoong:
This looks good to me. Thanks for the fix.

PR #1006, opened 2024-08-06
doc/libfuse-operations: Fix FUSE_STATX in_args[1] description

joannekoong:
FUSE_STATX only has 1 in_args value (struct fuse_statx_in), which is for in_args[0].

in_args[1 ] is not used.


bsbernd:
Thank you @joannekoong ! As I wrote at the top of the file, this list was generated with the help of chatgpt, I had fixed a few things it got wrong, but I didn't verify all op codes.

Issue #1009, opened 2024-08-12
fusermount -u vs NFS; patch included


oclsc:
If a FUSE file system is mounted on an NFS file system with the root_squash option in effect, and any directory in the path leading to the mount point is protected against searching by others, fusermount -u refuses to unmount it.

In the example, arith is a toy FUSE file system: cat X/Y should print X+Y.

This works:
DlS$ ls -ld mnt
drwxr-xr-x 2 norman cdfstaff 4096 Aug 12 11:53 mnt
DlS$ ./arith mnt
DlS$ cat mnt/3/4
7
DlS$ fusermount -u mnt

But this doesn't:
DlS$ ls -ld private private/mnt
drwx------ 3 norman cdfstaff 4096 Aug 12 11:55 private
drwxr-xr-x 2 norman cdfstaff 4096 Aug 12 11:55 private/mnt
DlS$ ./arith private/mnt
DlS$ cat private/mnt/4/5
9
DlS$ fusermount -u private/mnt
fusermount: failed to unmount /h/u1/norman/hack/fuse/private/mnt: Permission denied
DlS$ 

The trouble is in the privilege checks in util/fusermount.c unmount_fuse_locked.  The code drops privs (drop_privs()) before calling chdir_to_parent, so permissions are checked with the correct credentials; then restores them (restore_privs()) before calling umount2, presumably expecting that privs are required for unmounting.  But the full pathname is passed to umount2, so permissions are, in effect, checked again as super-user.  The trouble is that if NFS squash_root is in effect, and some directory in the path forbids access to others, it will then forbid access to the super-user (who for permission purposes is treated as nobody thanks to squash_root), and umount2 fails.

On Linux, drop_privs and restore_privs manipulate not the effective ids but the file-system ids.  If privs are dropped, the effective userid is still zero, so umount2 is still allowed; but the file-system ids are the real user's, so permission checks walking the path are done as that user and will succeed.  (We know they will succeed because chdir_to_parent succeeded.)  So the problem NFS exposes can be mended just by moving restore_privs to come after, rather than before, the umount2 call.

I'm guessing the existing code was written before there were separate fs ids; if all you have to manipulate is the effective ids, running umount2 with privs up is the only option.  And then nobody noticed that using fs ids allowed the NFS-permission problem (which was there all along) to be worked around.

[...]

PR #1011, opened 2024-08-12
fusermount -u vs NFS root_squash

oclsc:
Rearrange util/fusermount.c umount_fuse_locked() so that umount2 is called with privs dropped, not raised.  This works around a clash with NFS permissions: if FUSE mounted on NFS client directory with root_squash in effect, and some directory in the path leading to the mount point denies permissions to others, umount2 will fail because userid 0 cannot search it.  Since drop_privs now merely sets fsuid/fsgid, and effective userid is still 0, umount2 works fine.

See also issue #1009 .

### DIFF
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -523,11 +523,13 @@ static int unmount_fuse_locked(const char *mnt, int quiet, int lazy)
 
 	drop_privs();
 	res = chdir_to_parent(copy, &last);
-	restore_privs();
-	if (res == -1)
+	if (res == -1) {
+		restore_privs();
 		goto out;
+	}
 
 	res = umount2(last, umount_flags);
+	restore_privs();
 	if (res == -1 && !quiet) {
 		fprintf(stderr, "%s: failed to unmount %s: %s\n",
 			progname, mnt, strerror(errno));


bsbernd:
So I think why this works is that `setfsuid/setfsgid` is not dropping capabilities, see https://elixir.bootlin.com/linux/v6.10/source/kernel/sys.c#L872
I.e. the umount syscall calls into  

```
ksys_umount
    path_umount
         can_umount
             may_mount
                 ns_capable(current->nsproxy->mnt_ns->user_ns, CAP_SYS_ADMIN)
```

I.e. it doesn't check the userid at all, but only for CAP_SYS_ADMIN. Without looking this up, the s-uid binary will have that cap, then drop privs. which does not unset CAP_SYS_ADMIN - umount still works. So I don't think related to egid. 

(I was looking this up, to be sure we don't break anything.)

@oclsc would you mind to update the commit message a bit? And maybe also add a comment in the code?

oclsc:
As a certified old fart, I knew about capabilities but hadn't thought about how they interact with traditional set-userid.  So TIL!

Will make sure I understand it and update.

Procedural trivial: should I just submit a fresh pull request or is there some other preferred technique?

bsbernd:
Personally I prefer patch update using a force push - needs to update the patch locally (`git amend` or rebase, or personally I use `stg refresh` - new way of old style `quilt refresh`).

bsbernd:
Btw, recent `generic_permission`, which is used by the path walk/lookup is a pain to read, at least for group IDs. I switched back to linux-3.10 to refresh my memory for path based permissions.

bsbernd:
@oclsc Did you have a chance to look at kernel code? As written before, for me your change looks good, just the commit message should be updated a bit.

oclsc:
Partly this got shoved to the back of the table by other things, but mostly I fell down a rabbit hole trying to spot the actual place in the kernel where set-user-ID-root programs get their capabilities changed.  I found the routine that probably does it but can't find my way through all the indirection infrastructure to see how it gets called.

Not that it matters a lot.  As an old-school (pre-Linux) kernel hacker it was just too much fun to resist digging in.  I did find where it's clearly documented, in the middle of capabilities(7).

Will try to get to it tomorrow.

Cheers!

oclsc:
Pushed in a new commit statement.  I started by writing a really long one, but it seems too damn long (almost 50 lines).  If you think I'm being too terse now let me know and I'll find a happy middle.

bsbernd (approved):
Thanks for the comment update, looks good to me.

PR #1017, opened 2024-08-19
Fix FUSE_USE_VERSION in example/notify_store_retrieve.c

yangyun50:
This is an addition to commit e75d2c54a347. This example sets FUSE_USE_VERSION = 34 but uses fuse_loop_cfg_* APIs, which is not allowed since these APIs are not introduced in version 34.

### DIFF
--- a/example/notify_store_retrieve.c
+++ b/example/notify_store_retrieve.c
@@ -58,7 +58,7 @@
  */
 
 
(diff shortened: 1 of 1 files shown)


bsbernd:
Thanks for fixing it @yangyun50

PR #1018, opened 2024-08-19
fuse_loop_config_v1 conversion: Check for v2

bsbernd:
It is possible that v2 'struct fuse_loop_config' gets passed to fuse_session_loop_mt_32, when the application uses an old FUSE_USE_VERSION, but creates the config with the 3.12 API.

With this change fuse_loop_cfg_convert() will detect FUSE_LOOP_MT_V2_IDENTIFIER, copy over values (up to to struct fuse_loop_config_v1 size to avoid out of bounds read) and fill in the rest with defaults. A warning message is also shown now.


bsbernd:
@yangyun50 Could you review?

bsbernd:
handled by https://github.com/libfuse/libfuse/pull/1022

PR #1020, opened 2024-08-19
Add FUSE_OPEN_GETATTR support

bsbernd:
Additional notes:
fuse_reply_open_getattr takes 'struct timespec *' to avoid the conversion from double to sec/nsec. Especially with timeouts in the µsec and nsec area converting to double is awkward. If needed we can add a generic function to convert from struct timespec to double.

FUSE_OPENDIR_GETATTR is not implemented yet.


bsbernd:
Going to be upstream as compounds

PR #1022, opened 2024-08-20
add version check for the fuse_loop_cfg* operations

yangyun50:
fuse_loop_cfg* operations are introduced in fuse version 312 and can not be used in an early version.
If using these operations in an earlier version. The compile can not report errors before this patch and 
can lead to unexpected behaviors.
Also fix some typo.

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -121,11 +121,15 @@ struct fuse_file_info {
 /**
  * Configuration parameters passed to fuse_session_loop_mt() and
  * fuse_loop_mt().
+ * For FUSE API versions less than 312, use a public struct
+ * fuse_loop_config in applications and struct fuse_loop_config_v1
+ * is used in library (i.e., libfuse.so). These two structs are binary
+ * compatible in earlier API versions and can be linked.
  * Deprecated and replaced by a newer private struct in FUSE API
- * version 312 (FUSE_MAKE_VERSION(3, 12)
+ * version 312 (FUSE_MAKE_VERSION(3, 12)).
  */
 #if FUSE_USE_VERSION < FUSE_MAKE_VERSION(3, 12)
-struct fuse_loop_config_v1; /* forward declarition */
+struct fuse_loop_config_v1; /* forward declaration */
 struct fuse_loop_config {
 #else
 struct fuse_loop_config_v1 {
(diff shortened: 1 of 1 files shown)


include/fuse_common.h
```diff
@@ -984,6 +988,12 @@ int fuse_set_fail_signal_handlers(struct fuse_session *se);
  */
 void fuse_remove_signal_handlers(struct fuse_session *se);
 
+/**
+ * Config operations.
+ * These APIs are introduced in version 312 (FUSE_MAKE_VERSION(3, 12)).
+ * Using them in earlier versions will result in errors.
+ */
+#if FUSE_USE_VERSION >= FUSE_MAKE_VERSION(3, 12)
```
bsbernd, line 996: Yeah, I had thought about it and should have done it from the beginning in libfuse-3.12. So let's assume we take this change and a distribution now updates libfuse with this change. Some applications linked against libfuse have might have a wrong FUSE_USE_VERSION. So now distribution updates libfuse, but does not rebuild applications - everything works fine. Later on distribution triggers a rebuild of the application (for whatever reason, like new compiler/flags/libstdc++) - out of the sudden the build fails.
I.e. I should have done it from the beginning, but I fear that it is too late right now :(

bsbernd: I know that there is an issue, but we also cannot randomly fail distributions, as explained. Can you check with the alternative PR https://github.com/libfuse/libfuse/pull/1018 ? That will print a warning (i.e. in ideal case will trigger a bug report for the distribution package) and will take fields as much as possible?

bsbernd: Let's keep it open for now, until the other patch is merged. In principle I would like to hear the opinion of distribution maintainers now...

bsbernd:
@szmi What do you think about this change? Would it be possible to get in contact with the RHEL and Fedora maintainer to see if this would be an issue? See my comment above.

szmi:
> @szmi What do you think about this change? Would it be possible to get in contact with the RHEL and Fedora maintainer to see if this would be an issue? See my comment above.

As for RHEL, updating version should be rare so it's not a big issue.

Not sure about community distros where version updates are much more frequent, but I guess they can also handle breakages of dependencies and if happens once in a while it shouldn't be a big problem.

bsbernd:
Thanks @szmi Merging it then.

PR #1023, opened 2024-08-20
remove unused headers in example/poll_client

yangyun50:
fuse_config.h is used in libfuse internally. This may confuse developers that fuse_config.h is used in the examples.


bsbernd:
Thank you, merged!

PR #1026, opened 2024-08-23
statx support

joannekoong:
This commit adds libfuse support for FUSE_STATX requests on linux distributions.

Currently, statx is only supported on linux. To make the interface as ergonomic as possible (eg using native 'struct statx' vs 'struct fuse_statx'), this implementation gates the 'struct statx' changes by #ifdef linux.


bsbernd:
@joannekoong Thank you very much for your work on this. Would it be possible to add a test into `test/test_syscalls.c`? 
I also think we need a meson test if statx is supported by the distribution, there are people who compile recent libfuse with RHEL7 and I think that would fail. Or we would need add `struct statx` as compat and do the syscall ourselves - probably more work than just adding a meson check if the distribution has statx support.

example/passthrough.c
```diff
@@ -521,6 +521,24 @@ static off_t xmp_lseek(const char *path, off_t off, int whence, struct fuse_file
...
+static int xmp_statx(const char *path, int flags, int mask, struct statx *stxbuf,
+		     struct fuse_file_info *fi)
+{
+	int fd = 0;
+	int res;
+
+	if (fi)
+	    fd = fi->fh;
+
+	res = statx(fd, path, flags | AT_SYMLINK_NOFOLLOW, mask, stxbuf);
```
bsbernd, line 534: Kernel code has
```
	if (file && S_ISREG(inode->i_mode)) {
		struct fuse_file *ff = file->private_data;

		inarg.getattr_flags |= FUSE_GETATTR_FH;
		inarg.fh = ff->fh;
	}
```

So shouldn't be an empty path used when fi is set, with the AT_EMPTY_PATH flag?

from `man 2 statx`
       By file descriptor
              If pathname is an empty string and the AT_EMPTY_PATH flag is specified in flags  (see  be‐
              low), then the target file is the one referred to by the file descriptor dirfd.

Please correct me if I'm wrong, but right now it would treat `fd` as dirfd?

bsbernd: Also initialization `fd = 0` is a bit weird - 0 is a valid random value.

bsbernd: Ok, fd is ignored when pathname begins with a slash, but could you still initialize it to -1?

joannekoong:
Thanks for taking a look at this PR, @bsbernd! 

> Would it be possible to add a test into test/test_syscalls.c?

Absolutely, great idea! I will add this in.

> I also think we need a meson test if statx is supported by the distribution, there are people who compile recent libfuse with RHEL7 and I think that would fail. Or we would need add struct statx as compat and do the syscall ourselves - probably more work than just adding a meson check if the distribution has statx support.

Good point. I think statx is supported so long as glibc is version 2.28+, so i think we can just add this to the meson.build:

```
private_cfg.set('HAVE_STATX',
        cc.has_function('statx',  prefix : '#define _GNU_SOURCE\n#include <sys/stat.h>'))
```

Speaking of which, instead of gating the internal implementations by #ifdef linux, I'll change this to gating by #ifdef HAVE_STATX, so that this won't break the build if the user is on an old version of glibc.

lib/fuse_lowlevel.c
```diff
@@ -1131,6 +1131,22 @@ int fuse_reply_lseek(fuse_req_t req, off_t off)
 	return send_reply_ok(req, &arg, sizeof(arg));
 }
 
+#ifdef linux
+int fuse_reply_statx(fuse_req_t req, int flags, struct statx *statx,
+		     double attr_timeout)
```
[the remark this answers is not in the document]
joannekoong: I really don't like passing in double either and I try to generally avoid using floating point values whenever i can, but `fuse_reply_attr`'s API uses double for the attr_timeout (eg `int fuse_reply_attr(fuse_req_t req, const struct stat *attr, double attr_timeout);`) and unfortunately I think it's too late to change this for `fuse_reply_attr`. It feels like a cleaner interface to have these two APIs be consistent with `attr_timeout` but I don't feel too intensely about this. I'm happy to change this to `struct timespec` instead if you still prefer that

example/passthrough.c
joannekoong: I believe lib/fuse.c (in line 4475 below) will automatically set AT_EMPTY_PATH in flags if the path is NULL, so that the fuse server (if it's not using the low-level interface) doesn't have to worry about it. In lib/fuse.c 's statx implementation, if fi is set it calls get_path_nullok() (in line 4467) which can either return a NULL path or return a valid path.

joannekoong:
Hi @bsbernd, I recall you mentioning at LPC that you'd prefer this to wait on a future change you'll be making for using 'struct timespec'. Is there an eta for this change or can I help in some way to unblock this? I'm going to be upgrading Meta's internal libfuse library and would love to have this change in it as well. Thanks!

bsbernd:
@joannekoong Sorry and thanks for the reminder. Is Friday ok for this https://github.com/libfuse/libfuse/pull/1032 ? I don't think I will have time tomorrow, but Friday should be possible.

In principle you could already change your patch to use struct timespec, as it is independent.  I'm going to update PR1032 to use 'struct timespec entry_timeout' instead of double based on a feature flag. But I don't think you need to do that for your patch - everything that is new should discourage double as timeout, imho.

joannekoong:
No worries at all and thanks for the quick reply! It's totally okay if this takes a few weeks too if you're swamped with other stuff this week - there's no urgent rush on this particular issue since I'm also hoping to get the request timeout stuff in as well before I update the library, which hasn't landed upstream yet, but my aim is to get all of this in by the end of November. Thanks!

bsbernd:
Ok, the alternative is that prepare the patch, but I don't merge it yet. And then eventually finish the 1.17 release over the weekend (there is a long private discussion about the so bump and the other changes in that PR).

bsbernd:
Another issue we just noticed with Horst for TMPFILE (https://github.com/libfuse/libfuse/pull/1069) appending to ll_ops is still an ABI problem. I going to submit a different patch for that

joannekoong:
Hi @bsbernd, circling back on this - is there anything that's blocking this from being able to be merged in?

bsbernd:
Hi @joannekoong, I think I give up on changing timeouts from double, so no, nothing blocking. Would you have time to rebase it? I think there is a merge conflict with tmpfile in the mean time (sorry about that).

joannekoong:
Sounds great, I will rebase it right now. Thanks!

bsbernd:
@joannekoong we can ignore the ABI check - I thought I had already fixed that it does not complain about new functions, but could you please have a look at checkpatch output - looks like indentation is not right?
Also no idea why CodeSpell complains about unchanged files - it never did before - can be ignored.

joannekoong:
There are some warnings about ENOSYS checkpatch complains about:

`WARNING: ENOSYS means 'invalid syscall nr' and nothing else`

but I think these are ok. GNU [defines ENOSYS]( https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html#index-ENOSYS) as:

Macro: int ENOSYS
“Function not implemented.” This indicates that the function called is not implemented at all, either in the C library itself or in the operating system.

ChangeLog.rst
```diff
@@ -1,3 +1,9 @@
+libfuse 3.17.1 (unreleased)
+===========================
+
+* Add statx for distributions with stax support
```
bsbernd, line 4: Can you add that under 3.17? I created a branch, but didn't release it yet - will merge master into the 3.17 over the weekend?

example/memfs_ll.cc
```diff
@@ -1087,6 +1087,7 @@ static const struct fuse_lowlevel_ops memfs_oper = {
 	.copy_file_range = nullptr,
 	.lseek = nullptr,
 	.tmpfile = nullptr,
+	.statx = nullptr,
```
bsbernd, line 1090: #ifdef HAVE_STATX ?

bsbernd: Btw, interesting that the compiler does not complain for passthrough_hp.cc that it is not set (although I find missing C99 initialization silly in c++, maybe I should update standard for these files to c++20, which has that.

bsbernd:
Thanks for the update @joannekoong. I had tried to ignore -ENOSYS (https://github.com/libfuse/libfuse/commit/48d12f86e73e456b53defc4a8a98ed4db8a54b39), but looks like it didn't work, still complains. We are just going to ignore it.

joannekoong:
Thanks for the review @bsbernd!

include/fuse_lowlevel.h
```diff
@@ -167,9 +170,7 @@ enum fuse_notify_entry_flags {
 #define FUSE_SET_ATTR_KILL_PRIV	(1 << 14)
 #define FUSE_SET_ATTR_OPEN	(1 << 15)
 #define FUSE_SET_ATTR_TIMES_SET	(1 << 16)
-#define FUSE_SET_ATTR_TOUCH	(1 << 17)
-
-/* ----------------------------------------------------------- *
+#define FUSE_SET_ATTR_TOUCH	(1 << 17) /* ----------------------------------------------------------- *
```
bsbernd, line 173: I think something went wrong here.

include/fuse.h
```diff
@@ -850,6 +853,19 @@ struct fuse_operations {
 	 * Find next data or hole after the specified offset
 	 */
 	off_t (*lseek) (const char *, off_t off, int whence, struct fuse_file_info *);
+
+#ifdef HAVE_STATX
```
bsbernd, line 857: That is a public header, but HAVE_STATX is defined in `private_cfg`.  Either it needs to be in `public_cfg` in `meson.build` or without the ifdef in this header.

include/fuse_lowlevel.h
```diff
@@ -1325,6 +1325,21 @@ struct fuse_lowlevel_ops {
...
+	 *   fuse_reply_err
+	 *
+	 * @param req request handle
+	 * @param ino the inode number
+	 * @param flags bitmask of requested flags
+	 * @param mask bitmask of requested fields
+	 * @param fi file information (may be NULL)
+	 */
+	void (*statx)(fuse_req_t req, fuse_ino_t ino, int flags, int mask,
+		      struct fuse_file_info *fi);
```
bsbernd, line 1342: Here without #ifdef STATX, below for the reply with :)

bsbernd:
Thanks @joannekoong ! And I'm sorry, I had missed before that HAVE_STATX is in a private config only. If we remove HAVE_STATX from headers applications could implement their `->statx() ` functionality, and would not need care if libfuse was compiled with/without statx support? Hmm, although it might be useful if applictions could just use a define to know about that, than add configuration tests.
Maybe in private and public config, but headers still without the ifdef? 

(And with that I notice that we should also have a define if libfuse supports ->tmpfile())

include/fuse.h
joannekoong: Thanks, I think I need to add HAVE_STATX to the `public_cfg` too then. I think we need to have this ifdef guard in the header or else the "struct statx *stxbuf" arg will lead to compiler warnings for systems that don't have "struct statx" defined.

joannekoong:
Thanks for the review @bsbernd! I will add HAVE_STATX to the public config as well so that it's both in the public and the private config. I think we have to have the #ifdef HAVE_STATX guard in the headers where we explicitly use "struct statx" or else we'll have compilation issues. We don't need the #ifdef guard in the include/fuse_lowlevel.h file though because "struct statx" isn't used, but I can add an #ifdef guard around it anyways just for consistency.

include/fuse_lowlevel.h
joannekoong: i think if we add an #ifdef to this statx in `struct fuse_lowlevel_ops` here, then we'll need to add more #ifdefs to `do_statx()` in fuse_lowlevel.c. We don't strictly need an ifdef here because it doesn't reference "struct statx", so I thought it seemed cleaner to have this defined without ifdef, but I can add that in if you prefer.

include/fuse.h
bsbernd: There is a forward declaration of statx - that is maybe debatable, but should avoid compiler warnings?

bsbernd:
Thanks for the update @joannekoong There are BSD warnings now, you can see them by downloading the raw logs
```
2025-07-15T16:32:20.1425258Z ld: error: version script assignment of 'FUSE_3.17' to symbol 'fuse_fs_statx' failed: symbol not defined
2025-07-15T16:32:20.1426625Z ld: error: version script assignment of 'FUSE_3.17' to symbol 'fuse_reply_statx' failed: symbol not defined
2025-07-15T16:32:20.1427600Z cc: error: linker command failed with exit code 1 (use -v to see invocation)
2025-07-15T16:32:20.1611322Z [44/71] cc -Itest/test_syscalls.p -Itest -I../test -Iinclude -I../include -Ilib -I../lib -I. -I.. -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -std=gnu11 -O2 -g -D_REENTRANT -DHAVE_LIBFUSE_PRIVATE_CONFIG_H -Wno-sign-compare -D_FILE_OFFSET_BITS=64 -Wstrict-prototypes -Wmissing-declarations -Wwrite-strings -fno-strict-aliasing -MD -MQ test/test_syscalls.p/test_syscalls.c.o -MF test/test_syscalls.p/test_syscalls.c.o.d -o test/test_syscalls.p/test_syscalls.c.o -c ../test/test_syscalls.c
2025-07-15T16:32:20.1614798Z ../test/test_syscalls.c:966:12: warning: unused function 'test_statx' [-Wunused-function]
2025-07-15T16:32:20.1615510Z   966 | static int test_statx(void)
2025-07-15T16:32:20.1615923Z       |            ^~~~~~~~~~
2025-07-15T16:32:20.1616282Z 1 warning generated.
2025-07-15T16:32:21.9591809Z [45/71] c++ -Iexample/memfs_ll.p -Iexample -I../example -Iinclude -I../include -Ilib -I../lib -I. -I.. -fdiagnostics-color=always -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_FAST -Wall -Winvalid-pch -Wextra -std=c++17 -O2 -g -D_REENTRANT -DHAVE_LIBFUSE_PRIVATE_CONFIG_H -D_GNU_SOURCE -D_FILE_OFFSET_BITS=64 -Wno-sign-compare -Wmissing-declarations -Wwrite-strings -fno-strict-aliasing -pthread -std=c++17 -MD -MQ example/memfs_ll.p/memfs_ll.cc.o -MF example/memfs_ll.p/memfs_ll.cc.o.d -o example/memfs_ll.p/memfs_ll.cc.o -c ../example/memfs_ll.cc
2025-07-15T16:32:21.9595443Z ../example/memfs_ll.cc:1100:1: warning: missing field 'statx' initializer [-Wmissing-designated-field-initializers]
2025-07-15T16:32:21.9596168Z  1100 | };
2025-07-15T16:32:21.9596329Z       | ^
```


I just pushed a PR to increase C++ 20 std=c++20, that should avoid the memfs_ll warning.

example/memfs_ll.cc
```diff
@@ -1094,6 +1094,7 @@ static const struct fuse_lowlevel_ops memfs_oper = {
...
```
bsbernd, line 1097: #ifdef HAVE_STATX

ChangeLog.rst
```diff
@@ -45,6 +45,8 @@ libfuse 3.17 (2024-01-01, not officially releaesed)
 
 * Added support for FUSE_CAP_HANDLE_KILLPRIV_V2
 
+* Add statx for distributions with stax support
+
```
bsbernd, line 49: Oh, I always forget about that file - we need a new 3.18 section. Forgot to put in io-uring to it.

example/memfs_ll.cc
joannekoong: Thanks for taking a look at this. I was seeing a build failure when I had this gated with #ifdef HAVE_STATX earlier :( in fuse_lowlevel.h, the statx callback isn't ifdef-ed in the fuse_lowlevel_ops struct, so maybe we should just ifdef that in fuse_lowlevel_ops?

ChangeLog.rst
joannekoong: I'll remove this line from the Changelog for now. AFIACT from the https://github.com/libfuse/libfuse/releases page, fuse-3.18 hasn't been released yet

example/memfs_ll.cc
bsbernd: Yeah I think we should ifdef in fuse_lowlevel_ops and I had added a comment there back in January, but you had a counter argument :)

Anyway, once we merge https://github.com/libfuse/libfuse/pull/1285 this `nullptr` assignment shouldn't be needed anymore.

bsbernd: Btw, I merged the c++20 change. There is a FreeBSD test in the mean time (acativated a few weeks ago) - we only need to make this test happy.

joannekoong: After rebasing to on top of the c++20 change, i'm still seeing the failure 

```
../../../home/runner/work/libfuse/libfuse/example/memfs_ll.cc:1097:1: error: missing field 'statx' initializer [-Werror,-Wmissing-field-initializers]
 1097 | };
```

bsbernd: Hmm, I really thought that c++20 has the c99 struct initialization with zeroing of missing members :/

joannekoong: No worries, i'll just add an ifdef for it then in memfs_ll.cc

PR #1028, opened 2024-08-26
getattr: Make use of FUSE_GETATTR_FH in lowlevel examples

bsbernd:
High level examples were already using it, but not lowlevel. Also update the documentation.


joannekoong (approved):
This LGTM!

bsbernd:
Thanks for your review @joannekoong . Eventually took some time to test (manually) if this actually works.

Issue #1027, opened 2024-08-25
passthrough_hp with CAP_SYS_ADMIN does not handle O_DIRECT


hanwen:
The passthrough_hp example does not work correctly when using the passthrough backing FD feature with O_DIRECT. Repro:

```
$ rm -rf /tmp/mnt /tmp/orig && mkdir /tmp/mnt /tmp/orig && dd if=/dev/zero of=/tmp/orig/file bs=1024 count=1024 && sudo ./example/passthrough_hp --foreground /tmp/orig /tmp/mnt & (sleep 1; sudo dd if=/tmp/mnt/file of=/dev/null count=1024 bs=1024 iflag=direct; echo "result: $?"; sudo umount -f  /tmp/mnt)
[1] 55958
1024+0 records in
1024+0 records out
1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.00682801 s, 154 MB/s
ERROR: fuse_passthrough read failed.
dd: error reading '/tmp/mnt/file': Input/output error
0+0 records in
0+0 records out
0 bytes copied, 0.000720046 s, 0.0 kB/s
result: 1
```

This is because `sfs_read` assumes that any call when `passthrough` is set is some sort of error, however, it is expected when using `O_DIRECT` ? As a fix,  I tried

```
-    if (fs.passthrough && !fs.direct_io) {
-        cerr << "ERROR: fuse_passthrough read failed." << endl;
+    if (fs.passthrough && !fs.direct_io && !fi->direct_io) {
+        cerr << "ERROR: fuse_passthrough read failed. " <<  endl;
```

to fix this, but somehow the `fi->direct_io` flag gets stomped between the open and read call. I am not familiar enough with libfuse to further debug this.

joannekoong:
Hi @hanwen. I think this is because you need to set the `--direct-io` flag when you run passthrough_hp. Eg

```
[vmuser@myvm build]$  rm -rf /tmp/mnt /tmp/orig && mkdir /tmp/mnt /tmp/orig && dd if=/dev/zero of=/tmp/orig/file bs=1024 count=1024 && sudo ./example/passthrough_hp --foreground --direct-io /tmp/orig /tmp/mnt & (sleep 1; sudo dd if=/tmp/mnt/file of=/dev/null count=1024 bs=1024 iflag=direct; echo "result: $?"; sudo umount -f  /tmp/mnt)
[3] 2786
1024+0 records in
1024+0 records out
[...]

PR #1031, opened 2024-09-12
example/passthrough_hp: No auto FOPEN_DIRECT_IO in passthrough mode

bsbernd:
sfs_open and sfs_create set fi->direct_io (FOPEN_DIRECT_IO) when O_DIRECT is given, in order to benefit from a shared inode lock in kernel, i.e. to get parallel DIO writes. However, kernel side disabled passthrough when FOPEN_DIRECT_IO is set. Reads/writes had been totally failing in this case for O_DIRECT as sfs_write_buf() and sfs_read() have a sanity check. That sanity check could be modified, but for performance passthrough is better than parallel DIO, hence, we only want automatic FOPEN_DIRECT_IO for O_DIRECT when passthrough is not enabled.

Fixes: https://github.com/libfuse/libfuse/issues/1027

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -871,6 +871,9 @@ static void sfs_create_open_flags(fuse_file_info *fi)
     To make parallel_direct_writes valid, need set fi->direct_io
     in current function. */
     fi->parallel_direct_writes = 1;
+
+    fi->keep_cache = (fs.timeout != 0);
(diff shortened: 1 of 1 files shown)


bsbernd:
@yangyun50 Thanks for your review. Maybe we should just add a comment for the sanity check for now? I'm going to merge this first and will think about a comment next week.

Issue #727, opened 2023-01-30
Unnecessary use of floating point operations when using fuse_reply_entry or fuse_add_direntry_plus


davidgaleano:
The structure `fuse_entry_param` requires storing the timeout values as `double`s which internally are converted to integers when filling the `fuse_entry_out` structure that is sent back to the kernel.
This conversion from floating point values back to integer values is unfortunate when the application using libfuse can already provide the timeouts as integer seconds and nanoseconds, which is our case, we go from integers to floating point and back to integers.
It would be great if there was an alternative way of passing those timeouts using integer values, perhaps even to be able to fill the internal structure `fuse_entry_out` directly instead of going to the transient representation of a `stat` structure, a `fuse_reply_entry_raw` perhaps?

Nikratio:
I didn't check the code, but if you're right then pull requests are welcome.

bsbernd:
Entirely agreed, we are setting 200ns and that gets rather close to resolution of double. The other part that I don't like at all is that struct fuse_entry_param has 'fuse_ino_t ino', while fuse_entry_out and kernel actually treat this as nodeid (inode identifier) - totally confusing to call this 'ino', as one could assume that is the same as st_ino in struct stat.

bsbernd:
@davidgaleano could you review https://github.com/libfuse/libfuse/pull/1032 ?

davidgaleano:
That change is an improvement but, to be honest, it is rather messy and I think it actually complicates the library unnecessarily. Having to pass `timeout_as_double` around to know how to interpret the incoming entry parameter is unfortunate, at a minimum I would suggest to get rid of it and to assume that **all** internal APIs will use `timespec`, the external API can still use doubles or timespec according to the `#define` but when using doubles they will always be converted to call the `timespec` internal functions.

Also, I am concerned about existing programs dynamically linking to libfuse, this new version will break the ABI, right? `fuse_entry_param` is bigger because of the union with the timespec on it, and I am not sure if the new static inline functions will appear as part of the external symbols. I guess you were planning to increase the library version?

In any case, my hope was for a new set of functions and structures that totally avoided the need for `fill_entry` or `convert_stat`, basically let the application fill the data already in the format that the kernel requires, without the need for intermediate copies. That new set of functions would be leaner and simpler than the current one and it will not be hampered by having to support older versions of the API.

davidgaleano:
[...]

PR #1032, opened 2024-09-12
Allow to use struct timespec for attribute timeout

bsbernd:
This is rather ugly and mostly copy-and-paste. Main change for the new v2 functions/structs is that the attribute timeout is given as struct timespec instead of double.

### DIFF
--- a/.github/workflows/abicheck.yml
+++ b/.github/workflows/abicheck.yml
@@ -57,6 +57,7 @@ jobs:
 
       - name: Run abidiff
         run: abidiff 
+          --no-added-syms
           --headers-dir1 previous/include/ 
           --headers-dir2 current/include/ 
           previous/build/lib/libfuse3.so 
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -43,7 +43,10 @@
  * \include passthrough_hp.cc
  */
 
+#include "fuse_util.h"
 #define FUSE_USE_VERSION FUSE_MAKE_VERSION(3, 12)
+
+#define FUSE_ATTR_TIMEOUT_AS_TIMESPEC
 
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
@@ -144,7 +147,7 @@ struct Fs {
     std::mutex mutex;
     InodeMap inodes; // protected by mutex
     Inode root;
-    double timeout;
+    struct timespec timeout;
     bool debug;
     bool debug_fuse;
     bool foreground;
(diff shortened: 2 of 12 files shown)


bsbernd:
@joannekoong This is something I want to discuss at Cern next week.

joannekoong:
I'm interested in hearing how the discussion goes!

bsbernd:
I think the ABI check failure is a false positive, as ABI compatibility should be guaranteed by the version in compat.c

joannekoong:
Thanks for updating this to unblock statx! Do we know yet the performance gains from using timespec and avoiding the double to uint32_t conversion?

bsbernd:
I won't look into performance gains here, I think you could several more rounds between type conversions before you actually notice a difference, but then that creates OS noise, slightly increases power usage and also leads to weird timeouts like this (from our code)
```
/* Timeout for files should be less or equal than min network latency.
 * Using 200ns here as this is a bit less than best IB latency we've obeserved. 
 */
#define FUSE_ENTRY_TIMEOUT  0.0000002  // timeout for attributes (200ns)
```

And it is not only me reporting this: https://github.com/libfuse/libfuse/issues/727

joannekoong:
I'm not sure I wholly see the benefit of having this. I personally think the additional complexity this adds in the codebase isn't worth the benefit of supporting "struct timespec" but I'm also happy to defer to whatever you think would be best.

joannekoong:
I don't think I'm fully understanding what the main benefit is of replacing double. Is it so that the API is more ergonomic and we have to do less casting in the library?

If we're okay with replacing double with an equivalent-sized int, maybe we could have that as a union where depending on if the client opts into it, we can use the int instead of the double?

bsbernd:
Well, issue is that it is a bit weird that:
- application has timeouts as integer, needs to convert it to double
- libfuse converts is back to integers

That gets especially weird, if you want to set ns timeouts.

attr_valid has uint64_t, but I guess if we define something like

```
struct fuse_entry_param {
   union {
        double timeout;
        struct {
            uint32_t timeout_sec;
            uint32_t timeout_nsec;
        };
   }; 
};
```
struct fuse_entry_out::entry_valid and ::attr_valid are 64-bit types, but I really do not see why one would need more than 32 bit for these - kernel uses them as relative values...

I need to think a bit about if there is a way to simplify the patch a bit, I don't like the additional complexity either.

ashleypittman:
I agree it would certainly be good to remove the use of double but not at the cost of code complexity so I wonder if there would be a better way.

bsbernd:
I'm closing it, no time to work on it.

PR #1033, opened 2024-09-13
fusermount: close inherited fds

bsbernd:
When using the auto_unmount option, the fusermount3 watchdog process retains any inherited file descriptors. This PR ensures they are closed.

Reason is that FDs that are kept open for a long time might cause issues for applications using libfuse, for example if these expect a pipe close, but the pipe is kept open through the inherited file descriptor.
See for example: https://github.com/cvmfs/cvmfs/issues/3645


bsbernd:
@mharvey-jt Please also see the 2nd commit and its commit message.

PR #1034, opened 2024-09-14
passthrough_hp: Use Use fuse_loop_cfg_set_max_threads()

bsbernd:
fuse_loop_cfg_set_idle_threads() was by accident and
setting it might cause a performance issue.


joannekoong (approved):

joannekoong:
Nice catch!

PR #1036, opened 2024-09-19
Use NO_EXPORT_SUPPORT for passthrough_hp

bsbernd:


bsbernd:
@joannekoong

bsbernd:
@szmi Do you see a good reason for the checks the first patch removes? As the commit message says, kernel is just going to ignore unsupported flags? @Nikratio do you remember why you had added that check?

bsbernd:
@Nikratio The file system might just run an older kernel - i.e. with the additional check the file system ends up in

```
if (conn->capable & FEATURE)
    conn->want & FEATURE;
```

instead of just

```
conn->want & FEATURE;
```

I don't see a good reason why the file system should *need* to know which features the kernel has, when it just wants a feature. If it cannot run at all when  the kernel does not have a certain feature it can still check conn->capable.

bsbernd:
What we can do is to keep the test, but not to test for kernel flags, but for flags known by libfuse - using a bit that libfuse does not know about would be indeed a bug.

Nikratio:
> @Nikratio The file system might just run an older kernel - i.e. with the additional check the file system ends up in
> 
> ```
> if (conn->capable & FEATURE)
>     conn->want & FEATURE;
> ```
> 
> instead of just
> 
> ```
> conn->want & FEATURE;
> ```
> 
> I don't see a good reason why the file system should _need_ to know which features the kernel has, when it just wants a feature. If it cannot run at all when the kernel does not have a certain feature it can still check conn->capable.


You are completely right. 

If `conn->want` is set but not in conn->capable,  this can have two reasons:

1. The filesystem does not need to do anything special if FEATURE is not supported, and therefore sets it unconditionally. This is fine.
2. The filesystem author has assumed that this feature is always available (perhaps not even aware that different kernels support different features), or has made a mistake in coding the check, and things will go wrong. This is not good.

The point of the libfuse check is therefore purely to catch inattentive filesystem developers, at the expense of requiring a bit more coding from all filesystem developers. I think this is a reasonable price to pay, and personally I would always write

```
if (conn->capable & FEATURE) {
     conn->want & FEATURE;
} else {
   <Do what is needed to function correctly in absence of FEATURE> 
}
```
...or add a comment in the else branch explaining why nothing special needs to be done.

bsbernd:
@Nikratio I added a helper function to set these flags (based on a suggestion from @joannekoong) and kept the check. I just moved it to separate function. Personally I would actually like to be strict, if libfuse knows about set feature flags (if it does not, those bits are for sure an error). And be less strict for features the current kernel version does not support - not for now.

PR #1038, opened 2024-09-23
Increase the .so version and more ABI breakage

bsbernd:
There had been a slight ABI breakage in 3.11 (wrong position of `struct fuse_config::no_rofd_flush`), with a similar issue in 3.14.1 (wrong position of `struct fuse_config::parallel_direct_writes`). Much worse in 3.14. is `struct fuse_file_info::parallel_direct_writes` which has the risk of causing unintended behavior (which could lead in worst cast to data corruption).
  `abi-compliance-checker` also annotates in master/3.17 several issue when comparing to 3.16.2, although most of these are false positives. 
Some of these reported issues are in `struct fuse_config`  due to extension of that struct. Typically that struct is created internally, but we still better add reserved fields and also convert from int to int32_t (similar for uint32_t) (and re-order some fields) - when we do an ABI version bump anyway.
The main ABI breakage in  3.14.1 could be handled with heuristics, but it would be not beautiful and we wouldn't know when we could disable it. Only the upcoming 3.17 has encoded the version into the lib to be prepared to handle such situations. For now we better go with an so version bump to enforce recompilation.

### DIFF
--- a/include/cuse_lowlevel.h
+++ b/include/cuse_lowlevel.h
@@ -31,11 +31,14 @@ extern "C" {
 struct fuse_session;
 
 struct cuse_info {
-	unsigned	dev_major;
-	unsigned	dev_minor;
-	unsigned	dev_info_argc;
+	uint32_t	flags;
+	uint32_t	dev_major;
+	uint32_t	dev_minor;
+	uint32_t	dev_info_argc;
 	const char	**dev_info_argv;
-	unsigned	flags;
+
+	/* for future additions and ABI compatibility */
+	uint64_t	reserved[16];
 };
 
 /*
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -24,6 +24,7 @@
 #include <sys/stat.h>
 #include <sys/statvfs.h>
 #include <sys/uio.h>
+#include <assert.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -97,28 +98,39 @@ typedef int (*fuse_fill_dir_t) (void *buf, const char *name,
  *
  * Note: this data structure is ABI sensitive, new options have to be
  * appended at the end of the structure
+ *
+ * Note: This structure is ABI sensitive, only add new fields in reserved
+ *       bits.
  */
 struct fuse_config {
 	/**
 	 * If `set_gid` is non-zero, the st_gid attribute of each file
 	 * is overwritten with the value of `gid`.
 	 */
-	int set_gid;
-	unsigned int gid;
+	int32_t set_gid;
+	uint32_t gid;
 
 	/**
 	 * If `set_uid` is non-zero, the st_uid attribute of each file
 	 * is overwritten with the value of `uid`.
 	 */
-	int set_uid;
-	unsigned int uid;
+	int32_t set_uid;
+	uint32_t uid;
 
 	/**
 	 * If `set_mode` is non-zero, the any permissions bits set in
 	 * `umask` are unset in the st_mode attribute of each file.
 	 */
-	int set_mode;
-	unsigned int umask;
+	int32_t set_mode;
+	uint32_t umask;
+
+	/**
+	 * `fmask` and `dmask` function the same way as `umask`, but apply
+	 * to files and directories separately. If non-zero, `fmask` and
+	 * `dmask` take precedence over the `umask` setting.
+	 */
+	uint32_t fmask;
+	uint32_t dmask;
 
 	/**
 	 * The timeout in seconds for which name lookups will be
@@ -145,14 +157,14 @@ struct fuse_config {
 	/**
 	 * Allow requests to be interrupted
 	 */
-	int intr;
+	int32_t intr;
 
 	/**
 	 * Specify which signal number to send to the filesystem when
 	 * a request is interrupted.  The default is hardcoded to
 	 * USR1.
 	 */
-	int intr_signal;
+	int32_t intr_signal;
 
 	/**
 	 * Normally, FUSE assigns inodes to paths only for as long as
@@ -164,7 +176,7 @@ struct fuse_config {
 	 * A number of -1 means that inodes will be remembered for the
 	 * entire life-time of the file-system process.
 	 */
-	int remember;
+	int32_t remember;
 
 	/**
 	 * The default behavior is that if an open file is deleted,
@@ -182,7 +194,7 @@ struct fuse_config {
 	 * ENOENT): read(2), write(2), fsync(2), close(2), f*xattr(2),
 	 * ftruncate(2), fstat(2), fchmod(2), fchown(2)
 	 */
-	int hard_remove;
+	uint8_t hard_remove;
 
 	/**
 	 * Honor the st_ino field in the functions getattr() and
@@ -195,7 +207,7 @@ struct fuse_config {
 	 * Note that this does *not* affect the inode that libfuse 
 	 * and the kernel use internally (also called the "nodeid").
 	 */
-	int use_ino;
+	uint8_t use_ino;
 
 	/**
 	 * If use_ino option is not given, still try to fill in the
@@ -204,7 +216,7 @@ struct fuse_config {
 	 * found there will be used.  Otherwise it will be set to -1.
 	 * If use_ino option is given, this option is ignored.
 	 */
-	int readdir_ino;
+	uint8_t readdir_ino;
 
 	/**
 	 * This option disables the use of page cache (file content cache)
@@ -223,7 +235,7 @@ struct fuse_config {
 	 * `direct_io` field of `struct fuse_file_info` - overwriting
 	 * any value that was put there by the file system.
 	 */
-	int direct_io;
+	uint8_t direct_io;
 
 	/**
 	 * This option disables flushing the cache of the file
@@ -242,7 +254,7 @@ struct fuse_config {
 	 * `keep_cache` field of `struct fuse_file_info` - overwriting
 	 * any value that was put there by the file system.
 	 */
-	int kernel_cache;
+	uint8_t kernel_cache;
 
 	/**
 	 * This option is an alternative to `kernel_cache`. Instead of
@@ -250,7 +262,7 @@ struct fuse_config {
 	 * invalidated on open(2) if if the modification time or the
 	 * size of the file has changed since it was last opened.
 	 */
-	int auto_cache;
+	uint8_t auto_cache;
 
 	/**
 	 * By default, fuse waits for all pending writes to complete
@@ -258,15 +270,7 @@ struct fuse_config {
 	 * With this option, wait and FLUSH are not done for read-only
 	 * fuse fd, similar to the behavior of NFS/SMB clients.
 	 */
-	int no_rofd_flush;
-
-	/**
-	 * The timeout in seconds for which file attributes are cached
-	 * for the purpose of checking if auto_cache should flush the
-	 * file data on open.
-	 */
-	int ac_attr_timeout_set;
-	double ac_attr_timeout;
+	uint8_t no_rofd_flush;
 
 	/**
 	 * If this option is given the file-system handlers for the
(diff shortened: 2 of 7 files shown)


include/fuse_common.h
```diff
@@ -114,6 +114,9 @@ struct fuse_file_info {
 	 * create and open.  It is used to create a passthrough connection
 	 * between FUSE file and backing file. */
 	int32_t backing_id;
+
+	/* For future additions and ABI compatibility */
+	uint64_t reserved[32];
 };
```
mxmlnkn, line 120:
```
#define STATIC_ASSERT(COND,MSG) typedef char static_assertion_##MSG[(COND)?1:-1]

STATIC_ASSERT(sizeof(A) == 123, struct_size_must_not_change); 
```

Something like this might be a possibility to avoid struct size changes with something stronger than comments.

include/fuse.h
```diff
@@ -293,23 +296,28 @@ struct fuse_config {
...
 	 */
-	int parallel_direct_writes;
+	uint32_t parallel_direct_writes;
+
+	/**
+	 * The timeout in seconds for which file attributes are cached
+	 * for the purpose of checking if auto_cache should flush the
+	 * file data on open.
+	 */
+	int32_t ac_attr_timeout_set;
```
mxmlnkn, line 306: If changing the order of these two `ac_attr_*` members was intended, then I think that should be clearly documented by doing this in a separate commit. Currently, this is done in the commit labeled `Convert to uint32_t/int32_t in struct fuse_config`.

include/fuse.h
```diff
@@ -195,7 +206,7 @@ struct fuse_config {
 	 * Note that this does *not* affect the inode that libfuse 
 	 * and the kernel use internally (also called the "nodeid").
 	 */
-	int use_ino;
+	uint32_t use_ino;
```
mxmlnkn, line 209: You are changing signed int to unsigned int. Based on the commit message, which states `int -> int32_t`, this does not look intentional? Other members are also affected.

bsbernd: @mxmlnkn Yeah, was intended, but I was not done yet with the change. Initially I wanted change to `uint32_t field : 1;`, but I had forgotten that some values are exposed as parameters in fuse.c - resulted in removing `:1`. I updated these to uint8_t now. Also updated the commit message. 
Thanks for your review! Will review the entire series myself in the morning and then add other reviewers. I hope we get 3.17 out this week.

include/fuse.h
bsbernd: Thanks, as explained below, I updated the commit message.

include/fuse_common.h
bsbernd: I won't work for single bits - the issue in `struct fuse_config` was not beautiful, but mostly not critical. Shifted bits in `struct fuse_file_info` had been really bad. I prefer to go with external ABI verification tools, at least that is going to make me to look carefully if it reports something, even if it is a false positive.

bsbernd:
I just merged an a new ABI check that uses abidiff and rebased this PR - as I hoped for, it reports ABI breakage - expected here, but should prevent more issues in the future.

bsbernd:
@szmi @HereThereBeDragons  @ashleypittman  @trapexit  @amir73il  @joannekoong If you should have time could you please review? 

@amir73il I had created a patch to ABI heuristics (https://github.com/bsbernd/libfuse/tree/abi-heuristics), but I don't feel comfortable with that commit and there had been other issues. I prefer to bump the ABI version and to prevent such issues in the future - we have a need ABI check for that.

include/fuse_common.h
```diff
@@ -55,48 +56,47 @@ struct fuse_file_info {
...
+	uint32_t  cache_readdir : 1;
 
 	/** Can be filled in by open, to indicate that flush is not needed
 	    on close. */
-	unsigned int noflush : 1;
+	uint32_t  noflush : 1;
 
 	/** Padding.  Reserved for future use*/
-	unsigned int padding : 23;
-	unsigned int padding2 : 32;
```
amir73il, line 99: why did u remove the extra padding2? don't you think we will need it in the future?

bsbernd: Thanks Amir, good catch! I should have mentioned in the commit message - issue is that it created a hole according to `pahole`. Maybe I better do the other way around and add a `uint32_t padding3` to make use of that hole.

include/fuse.h
```diff
@@ -182,7 +194,7 @@ struct fuse_config {
 	 * ENOENT): read(2), write(2), fsync(2), close(2), f*xattr(2),
 	 * ftruncate(2), fstat(2), fchmod(2), fchown(2)
 	 */
-	int hard_remove;
+	uint8_t hard_remove;
```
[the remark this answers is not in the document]
bsbernd: Right, thanks, I updated the commit message to The C 'bool' type is not used, as the exact width is not defined by the C standard.`

szmi:
This is going to be tricky on the distro side, since some 3rd party software might rely on the old soversion, where there would be no way to force a recompile.   Which means that now two different libfuse3 versions would need to be distributed to cater for this case. 

And since distros will recompile dependent packages against the newest version of the library, the backward compatibility issues shouldn't affect those, leaving only 3rd party software with possible breakage.  AFAICS this isn't any worse, than having a new soversion, but avoids a lot of pain.

But I'm no expert, so possibly I'm missing something fundamental.

Thanks,
Miklos

bsbernd:
Hi Miklos, 

thanks for looking into it. I'm assuming that distributions do not recompile packages against a new library version, unless absolutely needed - if I'm right library ABI changes cause silent issues. I also assume that distributions typically do not care too much about 3rd party software. I.e. my assumption is that bumping the so version will only enforce recompilation of dependency packages, but I'm not an expert for that either. Going to ask on debian-devel (I assume Ubuntu works the same, but I have no clue about Fedora/RHEL and Suse/SLES and whom to contact there). 

I also think that it is better to notify 3rd party software, than just silently causing issues. Main issue for those might be that we already have an ABI issue and users might install a wrong version (to /usr/local).

Best,
Bernd

include/cuse_lowlevel.h
```diff
@@ -31,11 +31,14 @@ extern "C" {
...
 struct cuse_info {
-	unsigned	dev_major;
-	unsigned	dev_minor;
-	unsigned	dev_info_argc;
+	uint32_t 	dev_major;
+	uint32_t 	dev_minor;
+	uint32_t 	dev_info_argc;
 	const char	**dev_info_argv;
-	unsigned	flags;
+	uint32_t 	flags;
```
joannekoong, line 38: does it make sense here to move `uint32_t flags` to the top of this struct so that it's better aligned?

lib/fuse_i.h
```diff
@@ -43,28 +43,28 @@ struct fuse_notify_req {
...
 	char *mountpoint;
-	volatile int exited;
-	int fd;
+	volatile int32_t exited;
+	int32_t fd;
 	struct fuse_custom_io *io;
 	struct mount_opts *mo;
-	int debug;
-	int deny_others;
+	int32_t debug;
```
joannekoong, line 50: if we're forcing recompilation anyways, does it make sense to rework some of these fields like for example having debug, got_init, got_destroy, deny_others, etc. to be a bool/uint8_t or a bit instead of a int32_t?

smcv:
With distribution maintainer hat on:

The worst thing for distributions is an ABI break that is *not* accompanied by a SONAME change, which leaves the distro with no good choices available, only a series of bad choices (diverging from upstream, or doing a flag-day transition with no possibility of backwards compatibility, or ignoring the problem and hoping it resolves itself by mistake).

Doing a transition to a new SONAME with a new ABI is considerably less painful than that. At a scale of individual packages, it's usually a rare event (the packages that bump SONAME very frequently tend to be annoying to package in a distribution), but if you consider the distribution as a whole, there will usually be *something* undergoing a SONAME bump at any given time. If the ABI changed but the API has stayed (mostly) the same, so that the same source code can (usually) be recompiled against the new library without modifications, then it's quite a routine thing to do. Debian calls these events "transitions" and has procedures for coordinating them, and I would hope that other high-quality distributions do the same. https://release.debian.org/transitions/ lists some transitions that are going on at the moment or queued for a future time: not all of them are SONAME-related, but most of them are.

Obviously distributions prefer it if libraries can avoid breaking ABI for an extended period of time - a SONAME bump every few years is completely fine, a SONAME bump every few weeks is more difficult. So, if you can put mechanisms in place to make it less likely that another SONAME bump will be needed in the near future, that's good too.

If significant source code changes are required to update from the "old" version to the "new" version, then distributions prefer them to be parallel-installable (like FUSE 2 → 3, and similar major-version changes in libraries like GTK, Qt and SDL). If the same source code is still valid and a simple recompile will usually be all that is necessary, then that isn't needed.

> I'm assuming that distributions do not recompile packages against a new library version, unless absolutely needed - if I'm right library ABI changes cause silent issues

That is correct, in most cases (certainly true for the Debian, Fedora and Arch families). There are a few distros like NixOS that do cascading rebuilds (every new libfuse upload causes every libfuse-dependent package to be rebuilt), but that's rather expensive in terms of compile time and amount of attention required from developers, and is only really feasible when building on fast architectures like x86 - any distro that has a port available for slower architectures like riscv64 is unlikely to be doing this.

However, any sufficiently useful distribution will have a process for recompiling all dependent packages as a one-time action when the SONAME changes.

> distros will recompile dependent packages against the newest version of the library

As above: not routinely, only if there's a reason to need to do so (like a SONAME change). For example, [fuse-overlayfs in Debian](https://tracker.debian.org/pkg/fuse-overlayfs) was most recently updated nearly a year ago, compiled against whatever version of libfuse was current at the time, and has not been rebuilt since.

> some 3rd party software might rely on the old soversion, where there would be no way to force a recompile. Which means that now two different libfuse3 versions would need to be distributed to cater for this case

If the ABI of `libfuse3.so.3` has already had several incompatible versions, then the damage to third party software has already been done - if it was compiled against the "old" ABI and is now run against the "new" ABI, then there's already no guarantee that it still works as intended. So I think the best thing to do to accommodate third party software is *also* to bump the SONAME, so that at least third party distributors can know that if their binaries are linked to `libfuse3.so.4`, then they have avoided the ambiguous ABI of `libfuse3.so.3`.

In distributions that name their packages according to the SONAME (like Debian and openSUSE), users can keep `libfuse3.so.3` installed for as long as they need it, even after the OS packages have all switched to `libfuse3.so.4` or whatever is the SONAME of the new version. For example in Debian-world, you'd keep both the `libfuse3-3` and `libfuse3-4` packages installed.

Not all distributions do that - some call their packages something like `fuse` or `fuse3-libs`, without the SONAME being reflected in the packaging. I personally think that's a design issue in those distros, while maintainers of those distros probably think of it as a positive simplification. Either way, those distros have made their choice and now they'll have to live with the consequences.

> I also think that it is better to notify 3rd party software, than just silently causing issues

Yes, I agree.

lib/fuse_i.h
bsbernd: Thank you! fuse_i.h was actually accidental and not intended. We should be able to update it anytime, but I took your suggestion and reworked it a bit. Also made fields cache line aligned now.

joannekoong: Should we also roll in your "struct timespec" changes from https://github.com/libfuse/libfuse/pull/1032 while we're breaking ABI?

bsbernd:
@szmi Another things that we can fix during ABI breakage is capability bits. Firstly they are 32 bit variables right now - easy to increase to 64-bit. Secondly, their values got out of sync with their underlying kernel bits. Do you have objections to include `fuse_kernel.h` into `fuse_common`? And with that, fuse_kernel.h needs to be get installed? Ideally I would like to have

```
enum fuse_cap_bits {

/**
 * Indicates that the filesystem supports asynchronous read requests.
 *
 * If this capability is not requested/available, the kernel will
 * ensure that there is at most one pending read request per
 * file-handle at any time, and will attempt to order read requests by
 * increasing offset.
 *
 * This feature is enabled by default when supported by the kernel.
 */
#define FUSE_CAP_ASYNC_READ		FUSE_ASYNC_READ
```

Thanks,
Bernd

bsbernd:
@smcv Thanks a lot for your verbose reply!

bsbernd:
We probably need to take this https://github.com/libfuse/libfuse/pull/1081 PR in order to help RHEL.

bsbernd:
Closing as PR #1081 has been merged.

PR #1041, opened 2024-09-26
Initialize fuse session buffer size to consider max_pages_limit set in /proc/sys/fs/fuse

joannekoong:
Currently in libfuse, the buffer size for a fuse session is capped at 1 MiB on a 4k page system. A recent [patch](https://lore.kernel.org/linux-fsdevel/20240923171311.1561917-1-joannelkoong@gmail.com/T/#t) upstream was merged that allows the max number of pages per fuse request to be dynamically configurable through the /proc/sys interface (/proc/sys/fs/fuse/max_pages_limit).

This commit adds support for this on the libfuse side to set the fuse session buffer to take into account the max pages limit set in /proc/sys/fs/fuse/max_pages_limit. If this sysctl does not exist (eg older kernels), it will default to old behavior (using FUSE_MAX_MAX_PAGES (256) as the max pages limit). This allows for things like bigger write buffers per request.

Testing:
* install kernel with [patch](https://lore.kernel.org/linux-fsdevel/20240923171311.1561917-1-joannelkoong@gmail.com/T/#t) applied
* echo 512 | sudo tee /proc/sys/fs/fuse/max_pages_limit
* mount fuse filesystem
* dd if=/dev/urandom of=~/fuse_mount/file.txt bs=4M count=4
* write sizes per request are 2 MiB

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2907,6 +2907,29 @@ static void fuse_ll_pipe_destructor(void *data)
 	fuse_ll_pipe_free(llp);
 }
 
+static unsigned int get_max_pages(void)
+{
+	char buf[32];
+	int res;
+	int fd;
+
+	fd = open("/proc/sys/fs/fuse/max_pages_limit", O_RDONLY);
+	if (fd < 0)
+		return FUSE_MAX_MAX_PAGES;
+
+	res = read(fd, buf, sizeof(buf) - 1);
+
+	close(fd);
+
+	if (res < 0)
+		return FUSE_MAX_MAX_PAGES;
+
+	buf[res] = '\0';
+
+	res = strtol(buf, NULL, 10);
+	return res < 0 ? FUSE_MAX_MAX_PAGES : res;
+}
+
 int fuse_session_receive_buf(struct fuse_session *se, struct fuse_buf *buf)
 {
 	return fuse_session_receive_buf_int(se, buf, NULL);
@@ -3142,7 +3165,7 @@ struct fuse_session *_fuse_session_new_317(struct fuse_args *args,
 	if (se->debug)
 		fuse_log(FUSE_LOG_DEBUG, "FUSE library version: %s\n", PACKAGE_VERSION);
 
-	se->bufsize = FUSE_MAX_MAX_PAGES * getpagesize() +
+	se->bufsize = get_max_pages() * getpagesize() +
 		FUSE_BUFFER_HEADER_SIZE;
 
 	list_init_req(&se->list);
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2915,19 +2915,19 @@ static unsigned int get_max_pages(void)
 
 	fd = open("/proc/sys/fs/fuse/max_pages_limit", O_RDONLY);
 	if (fd < 0)
-		return FUSE_MAX_MAX_PAGES;
+		return FUSE_DEFAULT_MAX_PAGES_LIMIT;
 
 	res = read(fd, buf, sizeof(buf) - 1);
 
 	close(fd);
 
 	if (res < 0)
-		return FUSE_MAX_MAX_PAGES;
+		return FUSE_DEFAULT_MAX_PAGES_LIMIT;
 
 	buf[res] = '\0';
 
 	res = strtol(buf, NULL, 10);
-	return res < 0 ? FUSE_MAX_MAX_PAGES : res;
+	return res < 0 ? FUSE_DEFAULT_MAX_PAGES_LIMIT : res;
 }
 
 int fuse_session_receive_buf(struct fuse_session *se, struct fuse_buf *buf)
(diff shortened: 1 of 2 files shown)


lib/fuse_lowlevel.c
```diff
@@ -2907,6 +2907,28 @@ static void fuse_ll_pipe_destructor(void *data)
...
+		return FUSE_MAX_MAX_PAGES;
+
+	res = read(fd, buf, sizeof(buf) - 1);
+
+	close(fd);
+
+	if (res < 0)
+		return FUSE_MAX_MAX_PAGES;
+
+	buf[res] = '\0';
```
bsbernd, line 2927: I know it is badly inconvenient to use, but would you mind to switch to `strtol` as that provides error checking?

lib/fuse_lowlevel.c
```diff
@@ -2907,6 +2907,28 @@ static void fuse_ll_pipe_destructor(void *data)
...
+	fd = open("/proc/sys/fs/fuse/max_pages_limit", O_RDONLY);
+	if (fd < 0)
+		return FUSE_MAX_MAX_PAGES;
+
+	res = read(fd, buf, sizeof(buf) - 1);
+
+	close(fd);
+
+	if (res < 0)
+		return FUSE_MAX_MAX_PAGES;
```
bsbernd, line 2925: While you are there, would you have time to change this name? I think your kernel patch changed to FUSE_MAX_PAGES_LIMIT?

joannekoong: Will do!

lib/fuse_lowlevel.c
joannekoong: Will do!

lib/fuse_lowlevel.c
```diff
@@ -2907,6 +2907,29 @@ static void fuse_ll_pipe_destructor(void *data)
...
+
+	buf[res] = '\0';
+
+	res = strtol(buf, NULL, 10);
```
bsbernd, line 2929: That is what I mean with 'inconvenient to use'. Problem is that some architectures have sizeof(int) == 8 - result of LONG_MAX would be positive. I had created a wrapper for strtol in our code that does this check

```
    errno = 0;
    char *endptr;
    long val = strtol(str, &endptr, 10);

    if ((errno == ERANGE && (val == LONG_MAX || val == LONG_MIN)) ||
        (errno == EINVAL) || (errno != 0 && val == 0) || (endptr == str) ||
        (errno == 0 && *endptr != 0))
``

Comment (bsbernd) [reply lib/fuse_lowlevel.c:2929]: Btw, we are testing a new AI tool (since yesterday), it gives me this code suggestion
```
errno = 0;
char *endptr;
long res = strtol(str, &endptr, 10);

if ((errno == ERANGE && (res == LONG_MAX || res == LONG_MIN)) ||
    (errno != 0 && res == 0) ||
    (endptr == str) ||
    (*endptr != '\0' && !isspace((unsigned char)*endptr)))

```

Comment (bsbernd) [reply lib/fuse_lowlevel.c:2929]: (And the people who designed the result of strtol() should be punished by typing this 100 x per day, at least for a month ;) )

Comment (joannekoong) [reply lib/fuse_lowlevel.c:2929]: The highest value that can be set for the `/proc/sys/fs/fuse/max_pages_limit` sysctl value is 65535 and the kernel automatically enforces that the system can't set a value less than or equal to 0 or a non-numerical value. i verified this with
```
$ echo 5a132 | sudo tee /proc/sys/fs/fuse/max_pages_limit
5a132
tee: /proc/sys/fs/fuse/max_pages_limit: Invalid argument
$ echo 0 | sudo tee /proc/sys/fs/fuse/max_pages_limit
0
tee: /proc/sys/fs/fuse/max_pages_limit: Invalid argument
$ echo -12 | sudo tee /proc/sys/fs/fuse/max_pages_limit
-12
tee: /proc/sys/fs/fuse/max_pages_limit: Invalid argument
$ echo 765535 | sudo tee /proc/sys/fs/fuse/max_pages_limit
765535
tee: /proc/sys/fs/fuse/max_pages_limit: Invalid argument
```
Do you think we still need these checks for ERANGE and endptr != '\0'?

Btw I looked at the [glibc implementation of atoi ](https://codebrowser.dev/glibc/glibc/stdlib/atoi.c.html#atoi) and it calls strtol underneath haha 
```
/* Convert a string to an int.  */
int
atoi (const char *nptr)
{
  return (int) strtol (nptr, (char **) NULL, 10);
}
```

bsbernd: Yeah, atoi() is not checking for errors.

bsbernd: Ok, fine, let's merge it, but I might change it, when we need another char-to-int conversion - from my point of view it makes sense to just add a wrapper for strtol to make it easier to use.

joannekoong: Thanks, I'll submit a PR that adds this wrapper and fixes strtol / atoi usages to use the wrapper.

joannekoong:
I'm going to submit an amendment to this PR. I think it's best if the server explicitly opts into extended max write sizes. Otherwise, if there are lots of fuse servers running simultaneously (eg on a production machine) where only a certain subset of servers wish to use large max writes, it'll be much to have libfuse allocate extremely large buffers for every single fuse server.

PR #1043, opened 2024-09-26
fuse_lowlevel FUSE_INIT: group bufsize and max_write adjustments

bsbernd:
The further these lines are separated from each other the harder it is to read the code.

There shouldn't be any code change behavior here.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2146,11 +2146,8 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 		bufsize = FUSE_MIN_READ_BUFFER;
 	}
 
-	if (se->conn.max_write > bufsize - FUSE_BUFFER_HEADER_SIZE)
-		se->conn.max_write = bufsize - FUSE_BUFFER_HEADER_SIZE;
-	if (se->conn.max_write < bufsize - FUSE_BUFFER_HEADER_SIZE)
-		bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
-	se->bufsize = bufsize;
+	se->conn.max_write = MIN(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE);
+	se->bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
 
 	if (arg->flags & FUSE_MAX_PAGES) {
 		outarg.flags |= FUSE_MAX_PAGES;
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2114,19 +2114,9 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 
 	se->conn.time_gran = 1;
 
-	if (bufsize < FUSE_MIN_READ_BUFFER) {
-		fuse_log(FUSE_LOG_ERR, "fuse: warning: buffer size too small: %zu\n",
-			bufsize);
-		bufsize = FUSE_MIN_READ_BUFFER;
-	}
-	se->bufsize = bufsize;
-
 	se->got_init = 1;
 	if (se->op.init)
 		se->op.init(se->userdata, &se->conn);
-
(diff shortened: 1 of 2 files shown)


bsbernd:
@joannekoong  @legezywzh If you have time, could you please review?

lib/fuse_lowlevel.c
```diff
@@ -2149,9 +2139,19 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
...
+	if (bufsize < FUSE_MIN_READ_BUFFER) {
+		fuse_log(FUSE_LOG_ERR,
+			 "fuse: warning: buffer size too small: %zu\n",
+			 bufsize);
+		bufsize = FUSE_MIN_READ_BUFFER;
 	}
+
+	if (se->conn.max_write > bufsize - FUSE_BUFFER_HEADER_SIZE)
+		se->conn.max_write = bufsize - FUSE_BUFFER_HEADER_SIZE;
+	if (se->conn.max_write < bufsize - FUSE_BUFFER_HEADER_SIZE)
```
joannekoong, line 2151: Maybe this could be `else if`?

Or, another way of expressing this block of logic that might be simpler:
```
se->bufsize = min(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE);
se->conn.max_write = se->bufsize;
```

joannekoong (approved):
LGTM. Your change makes the code more readable, thanks!

lib/fuse_lowlevel.c
bsbernd: Good idea to further simplify it, I'm just more tended to 

```
	se->conn.max_write = MIN(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE);

        /* the code might want to do larger allocations for the buffer than needed, 
         * to avoid fragmentation and let the system use transparent huge pages.
         */
	se->bufsize = MAX(bufsize, se->conn.max_write + FUSE_BUFFER_HEADER_SIZE);
```

What do you think?

bsbernd: Btw, I think your above setting is not correct, as max_write is bufsize - FUSE_BUFFER_HEADER_SIZE;

We could also do

```
se->conn.max_write = MIN(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE);
se->bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
```

And then let the actual allocation code round it up to something more THP usable, if it wants to.

joannekoong: Oh yeah, se->bufsize needs to include the FUSE_BUFFER_HEADER, nice catch.

```
se->conn.max_write = MIN(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE); se->bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
```
looks great to me

lib/fuse_lowlevel.c
```diff
@@ -2146,13 +2146,11 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
...
-		se->conn.max_write = bufsize - FUSE_BUFFER_HEADER_SIZE;
-	if (se->conn.max_write < bufsize - FUSE_BUFFER_HEADER_SIZE)
-		bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
-	se->bufsize = bufsize;
+	se->conn.max_write = MIN(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE);
+	se->bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
 
-	if (arg->flags & FUSE_MAX_PAGES) {
+	if (arg->flags & FUSE_MAX_PAGES)
+	{
```
joannekoong, line 2153: I think you accidentally included this line break for { as part of your commit

joannekoong (approved):
looks great!

lib/fuse_lowlevel.c
bsbernd: Ah thank you!

PR #1045, opened 2024-10-07
Add libfuse util strtol wrapper

joannekoong:
Add a wrapper around strtol for more rigorous error checking and convert uses of atoi and strtol to use this instead.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -1,0 +1,27 @@
+#include <stdlib.h>
+#include <errno.h>
+
+#include "util.h"
+
+int libfuse_strtol(const char *str, long *res)
+{
+	char *endptr;
+	int base = 10;
+	long val;
+
+	errno = 0;
+
+	if (!str)
+		return -EINVAL;
+
+	val = strtol(str, &endptr, base);
+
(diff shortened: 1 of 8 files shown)


lib/util.c
```diff
@@ -0,0 +1,26 @@
+#include <stdlib.h>
+#include <errno.h>
+
+#include "util.h"
+
+long libfuse_strtol(const char *str)
```
trapexit, line 6: Is this not confusing? Forcing a signed type to be effectively unsigned?

bsbernd: For me as well, the wrapper I had in my mind is more like

int libfuse_strtol(const char *str, long *res);

Return is then 0/error-code and res provides the result.

@joannekoong Sorry for my late reply, was concentrating on fuse-uring. Do want to make the update above, or shall I do that for you?

joannekoong: Ah I didn't realize "strtol" also takes in negative numbered strings. I'll fix this up and resubmit

bsbernd:
Thank you very much @joannekoong !

PR #1049, opened 2024-10-10
Ignore mtab on android

drosen-google:
Updating the mtab on Android fails due to differences in toybox's mount command. Setting IGNORE_MTAB avoids that issue.


bsbernd:
Thanks @drosen-google , merged.

PR #1050, opened 2024-10-11
fuse_buf_size fix for SIZE_MAX

rockeet:
When `.size == SIZE_MAX` is not the last element of array, it should be a bug.

### DIFF
--- a/lib/buffer.c
+++ b/lib/buffer.c
@@ -25,10 +25,10 @@ size_t fuse_buf_size(const struct fuse_bufvec *bufv)
 	size_t size = 0;
 
 	for (i = 0; i < bufv->count; i++) {
-		if (bufv->buf[i].size == SIZE_MAX)
-			size = SIZE_MAX;
-		else
-			size += bufv->buf[i].size;
+		if (bufv->buf[i].size >= SIZE_MAX - size)
+			return SIZE_MAX;
+
+		size += bufv->buf[i].size;
 	}
 
 	return size;


bsbernd:
Obviously correct. Thouh, I see more issues if there are elements before and their sum exceeds SIZE_MAX... While you are updating this code anyway, could you do like this?

```
size_t fuse_buf_size(const struct fuse_bufvec *bufv)
{
	size_t i;
	size_t size = 0;

        /* sum of bufv element sizes, with overflow handling  */
	for (i = 0; i < bufv->count; i++) {
		if (bufv->buf[i].size >= SIZE_MAX - size )
			return SIZE_MAX;

                size += bufv->buf[i].size;
	}

	return size;
}

```

rockeet:
Done!

lib/buffer.c
```diff
@@ -25,8 +25,8 @@ size_t fuse_buf_size(const struct fuse_bufvec *bufv)
 	size_t size = 0;
 
 	for (i = 0; i < bufv->count; i++) {
-		if (bufv->buf[i].size == SIZE_MAX)
-			size = SIZE_MAX;
+		if (bufv->buf[i].size >= SIZE_MAX - size)
+			return SIZE_MAX;
 		else
```
bsbernd, line 30: Thanks, objections to remove this else? I don't think we need it?

rockeet: @bsbernd  Yes, the else is not needed, done as your advice!

bsbernd:
@yangyun50 You mean `struct fuse_buf buf[1]`? Isn't that just an outdated way to describe a flexible array member? We should change it to `[]`.

PR #1063, opened 2024-11-11
msbuild 2.0.0

bysnet:


bsbernd:
Sorry, I don't understand what this is about. libfuse is a linux and bsd specific library.

PR #1065, opened 2024-11-15
Reallocate fuse_session buffer transparently for extended max writes 

joannekoong:
A previous [PR](https://github.com/libfuse/libfuse/pull/1041) supported extended max writes (eg write requests larger than 1 MB) by initializing the fuse session buffer size to use the max_pages_limit set in /proc/sys/fs/fuse. However, this is a huge problem for machines where multiple fuse servers may be running but only one server needs large writes. In this case, a lot of memory will be wasted and will lead to OOM issues.
 
This PR does a reallocation of the session buffer transparently if the server set "se->conn.max_write" to a value larger than 1 MiB. This is only for buffers that are "owned" by libfuse - if the server wishes to provide its own allocated buffer for receiving/processing requests, then it should ensure that buffer is allocated to the proper size from the start.

Local testing showed:
echo 65535 | sudo tee /proc/sys/fs/fuse/max_pages_limit
dd if=/dev/urandom of=hello_file bs=6M count=2

write requests: 
write request size is 5242880
write request size is 1048576
write request size is 5242880
write request size is 1048576

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -859,6 +859,14 @@ struct fuse_buf {
 	 * Used if FUSE_BUF_FD_SEEK flag is set.
 	 */
 	off_t pos;
+
+	/**
+	 * Size of memory pointer
+	 *
+	 * Used only if mem was internally allocated.
+	 * Not used if mem was user-provided.
+	 */
+	size_t mem_size;
 };
 
 /**
@@ -915,6 +923,7 @@ struct libfuse_version
 			/* .mem =   */ NULL,			\
 			/* .fd =    */ -1,			\
 			/* .pos =   */ 0,			\
+			/* .mem_size = */ 0,                    \
 		} }						\
 	} )
 
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -1977,6 +1977,7 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 	size_t outargsize = sizeof(outarg);
 	uint64_t inargflags = 0;
 	uint64_t outargflags = 0;
+	bool buf_reallocable = se->buf_reallocable;
 	(void) nodeid;
 	if (se->debug) {
 		fuse_log(FUSE_LOG_DEBUG, "INIT: %u.%u\n", arg->major, arg->minor);
@@ -2061,6 +2062,7 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 			if (bufsize > max_bufsize) {
 				bufsize = max_bufsize;
 			}
+			buf_reallocable = false;
 		}
 		if (inargflags & FUSE_DIRECT_IO_ALLOW_MMAP)
 			se->conn.capable |= FUSE_CAP_DIRECT_IO_ALLOW_MMAP;
@@ -2149,6 +2151,8 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 		bufsize = FUSE_MIN_READ_BUFFER;
 	}
 
+	if (buf_reallocable)
+	    bufsize = UINT_MAX;
 	se->conn.max_write = MIN(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE);
 	se->bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
 
@@ -2907,30 +2911,6 @@ static void fuse_ll_pipe_destructor(void *data)
 	fuse_ll_pipe_free(llp);
 }
 
-static unsigned int get_max_pages(void)
-{
-	char buf[32];
-	long res;
-	int fd;
-	int err;
-
-	fd = open("/proc/sys/fs/fuse/max_pages_limit", O_RDONLY);
-	if (fd < 0)
-		return FUSE_DEFAULT_MAX_PAGES_LIMIT;
-
-	res = read(fd, buf, sizeof(buf) - 1);
-
-	close(fd);
-
-	if (res < 0)
-		return FUSE_DEFAULT_MAX_PAGES_LIMIT;
-
-	buf[res] = '\0';
-
-	err = libfuse_strtol(buf, &res);
-	return err ? FUSE_DEFAULT_MAX_PAGES_LIMIT : res;
-}
-
 int fuse_session_receive_buf(struct fuse_session *se, struct fuse_buf *buf)
 {
 	return fuse_session_receive_buf_int(se, buf, NULL);
@@ -2941,8 +2921,8 @@ int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
 {
 	int err;
 	ssize_t res;
-#ifdef HAVE_SPLICE
 	size_t bufsize = se->bufsize;
+#ifdef HAVE_SPLICE
 	struct fuse_ll_pipe *llp;
 	struct fuse_buf tmpbuf;
 
@@ -3022,6 +3002,8 @@ int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
 					"fuse: failed to allocate read buffer\n");
 				return -ENOMEM;
 			}
+			buf->mem_size = se->bufsize;
+			se->buf_reallocable = true;
 		}
 		buf->size = se->bufsize;
 		buf->flags = 0;
@@ -3059,22 +3041,40 @@ fallback:
 				"fuse: failed to allocate read buffer\n");
 			return -ENOMEM;
 		}
+		buf->mem_size = se->bufsize;
+		se->buf_reallocable = true;
 	}
 
 restart:
+	if (se->buf_reallocable)
+	    bufsize = buf->mem_size;
 	if (se->io != NULL) {
 		/* se->io->read is never NULL if se->io is not NULL as
 		specified by fuse_session_custom_io()*/
-		res = se->io->read(ch ? ch->fd : se->fd, buf->mem, se->bufsize,
+		res = se->io->read(ch ? ch->fd : se->fd, buf->mem, bufsize,
 					 se->userdata);
 	} else {
-		res = read(ch ? ch->fd : se->fd, buf->mem, se->bufsize);
+		res = read(ch ? ch->fd : se->fd, buf->mem, bufsize);
 	}
 	err = errno;
 
 	if (fuse_session_exited(se))
 		return 0;
 	if (res == -1) {
+		if (err == EINVAL && se->buf_reallocable && se->bufsize > buf->mem_size)  {
+		    void *newbuf  = malloc(se->bufsize);
+		    if (!newbuf) {
+			fuse_log(FUSE_LOG_ERR,
+				"fuse: failed to (re)allocate read buffer\n");
+			return -ENOMEM;
+		    }
+		    free(buf->mem);
+		    buf->mem = newbuf;
+		    buf->mem_size = se->bufsize;
+		    se->buf_reallocable = true;
+		    goto restart;
+		}
+
 		/* ENOENT means the operation was interrupted, it's safe
 		   to restart */
 		if (err == ENOENT)
@@ -3130,7 +3130,8 @@ struct fuse_session *_fuse_session_new_317(struct fuse_args *args,
 		goto out1;
 	}
 	se->fd = -1;
-	se->conn.max_write = UINT_MAX;
+	se->conn.max_write = FUSE_DEFAULT_MAX_PAGES_LIMIT * getpagesize();
+	se->bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
 	se->conn.max_readahead = UINT_MAX;
 
 	/* Parse options */
@@ -3165,9 +3166,6 @@ struct fuse_session *_fuse_session_new_317(struct fuse_args *args,
 
 	if (se->debug)
 		fuse_log(FUSE_LOG_DEBUG, "FUSE library version: %s\n", PACKAGE_VERSION);
-
-	se->bufsize = get_max_pages() * getpagesize() +
-		FUSE_BUFFER_HEADER_SIZE;
 
 	list_init_req(&se->list);
 	list_init_req(&se->interrupts);
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -9,6 +9,8 @@
 #include "fuse.h"
 #include "fuse_lowlevel.h"
 
+#include <stdbool.h>
+
 #define MIN(a, b) \
 ({									\
 	typeof(a) _a = (a);						\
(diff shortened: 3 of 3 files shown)


include/fuse_common.h
```diff
@@ -846,6 +847,14 @@ struct fuse_buf {
 	 */
 	void *mem;
 
+	/**
```
joannekoong, line 850: Would it still break ABI if this was moved to the end of this struct? 

Otherwise, the only other option is to have the server pass in the extended max write size as a command-line arg (see https://github.com/libfuse/libfuse/pull/984), which I'd very much like to avoid if possible.

bsbernd: I will take a look tomorrow in the evening, the 4th weekend occupied with something else :(
My guess is that moving to the end would also have some issues, as we don't know the size of the passed struct - don't know if it is initialized, but the good part is that there is `enum fuse_buf_flags flags` - I guess we need to enforce that this flag gets set for the extended struct size. But I really first need to check the code tomorrow.

lib/fuse_lowlevel.c
```diff
@@ -2835,6 +2838,8 @@ void fuse_session_process_buf_int(struct fuse_session *se,
 		do_write_buf(req, in->nodeid, inarg, buf);
 	else if (in->opcode == FUSE_NOTIFY_REPLY)
 		do_notify_reply(req, in->nodeid, inarg, buf);
+	else if (in->opcode == FUSE_INIT)
+		do_init(req, in->nodeid, inarg, buf->mem_size);
```
bsbernd, line 2842: You can't trust if this isn't called from an older libfuse version - at a minimum this needs to check a flag.
And with header/payload separation we can probably get entirely rid of the if conditions here - would be great if we could avoid another exception.

The issue I have in general is that with fuse-io-uring buffers are registered with FUSE_URING_REQ_FETCH and then pages are for privileged users persistently mapped. 

And then, what is if allocation at run time fail? Already bad enough that we (so far) dynamic threads, but now it also would do re-allocation.

Personally, I would really prefer to the decision about IO sizes at initialization time - which is already possible. We also might just set 1MB (or lower buffer sizes by default).

joannekoong: Thanks for taking a look at this PR.

> You can't trust if this isn't called from an older libfuse version - at a minimum this needs to check a flag.

In the do_init() call, if the older libfuse version doesn't support `FUSE_MAX_PAGES`, then no reallocation for big write buffers happens.

> The issue I have in general is that with fuse-io-uring buffers are registered with FUSE_URING_REQ_FETCH and then pages are for privileged users persistently mapped.

For fuse-io-uring buffers, my understanding is that at the time buffers get registered with FUSE_URING_REQ_FETCH, the needed buf size would already be known (from the server's reply to the init call, which would go through through /dev/fuse before uring initialization is set up by the server).

 > And then, what is if allocation at run time fail? 

I think this would only fail for the out of memory case, which seems rare and could happen for any other allocation path at runtime too. 

> Already bad enough that we (so far) dynamic threads, but now it also would do re-allocation.

The re-allocation would only need to happen once. The main problem is that we need to allocate a buffer in order to set up / initialize the connection, but we only know if the server opts into bigger write buffers through their init callback. Unless instead, they mandate that they provide this information at mount time.

> Personally, I would really prefer to the decision about IO sizes at initialization time - which is already possible. We also might just set 1MB (or lower buffer sizes by default).

By 'initialization time', do you mean 'mount time'? I'm okay with this too - I just didn't like how clunky this interface gets ergonomically with having to opt into this at mount time. This is the approach I first submitted in https://github.com/libfuse/libfuse/pull/984, I'm happy to resurrect this and go with this approach if you prefer this approach instead.

lib/fuse_lowlevel.c
```diff
@@ -3059,22 +3041,35 @@ int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
...
 	if (se->io != NULL) {
 		/* se->io->read is never NULL if se->io is not NULL as
 		specified by fuse_session_custom_io()*/
-		res = se->io->read(ch ? ch->fd : se->fd, buf->mem, se->bufsize,
+		res = se->io->read(ch ? ch->fd : se->fd, buf->mem, buf->mem_size ?: se->bufsize,
 					 se->userdata);
 	} else {
-		res = read(ch ? ch->fd : se->fd, buf->mem, se->bufsize);
+		res = read(ch ? ch->fd : se->fd, buf->mem, buf->mem_size ?: se->bufsize);
 	}
```
bsbernd, line 3055: I mean here - another use of buf->mem_size.

lib/fuse_lowlevel.c
bsbernd: > Thanks for taking a look at this PR.
> 
> > You can't trust if this isn't called from an older libfuse version - at a minimum this needs to check a flag.
> 
> In the do_init() call, if the older libfuse version doesn't support `FUSE_MAX_PAGES`, then no reallocation for big write buffers happens.

There are also users in `fuse_session_receive_buf_int`, I just annotated it. And it is reading a field beyond old struct size (assuming you move the field at the end of the struct)  - I really think a flag would be needed if we go this way. 

> 
> > The issue I have in general is that with fuse-io-uring buffers are registered with FUSE_URING_REQ_FETCH and then pages are for privileged users persistently mapped.
> 
> For fuse-io-uring buffers, my understanding is that at the time buffers get registered with FUSE_URING_REQ_FETCH, the needed buf size would already be known (from the server's reply to the init call, which would go through through /dev/fuse before uring initialization is set up by the server).

Right now not, these /dev/fuse threads act on their own, we would basically need to wait for the reply to set things up. But then the reply might not come for a long time - if the /dev/fuse reading thread is waiting in fuse_dev_do_read(), because the provided buffer size is correct.  Also, fuse-io-uring is set up before libfuse even starts to read /dev/fuse, just the async nature of io-uring then typically cause FUSE_INIT to be complete before fuse-io-uring setup is complete.

With the current patch that uses UINT_MAX, I wouldn't know at all what to use as  fuse-io-uring buffer size.



> 
> > And then, what is if allocation at run time fail?
> 
> I think this would only fail for the out of memory case, which seems rare and could happen for any other allocation path at runtime too.

Other _large_ run time allocations are these dynamics threads. I won't look into any of that as I see it deprecated with io-uring.

> 
> > Already bad enough that we (so far) dynamic threads, but now it also would do re-allocation.
> 
> The re-allocation would only need to happen once. The main problem is that we need to allocate a buffer in order to set up / initialize the connection, but we only know if the server opts into bigger write buffers through their init callback. Unless instead, they mandate that they provide this information at mount time.
> 
> > Personally, I would really prefer to the decision about IO sizes at initialization time - which is already possible. We also might just set 1MB (or lower buffer sizes by default).
> 
> By 'initialization time', do you mean 'mount time'? I'm okay with this too - I just didn't like how clunky this interface gets ergonomically with having to opt into this at mount time. This is the approach I first submitted in #984, I'm happy to resurrect this and go with this approach if you prefer this approach instead.

That introduces se->extended_max_write`. why can't we use 'se->conn.max_write? 

Well, maybe I just move the initialization of io-uring post FUSE_INIT reply and we double ensure that new threads are not created as long as FUSE_INIT reply is not send. The FUSE_INIT reply would then still need to use the max size, that is currently read from /proc. I would also like to avoid the extra parameter for the do_init() function, if you don't want to store 'reallocable' into fuse_session, a local file variable should also do?

And alternative solution to reading /proc would involve kernel change, and fuse-client would send something like FUSE_INIT_PARAMS (or whatever it is called), with the parameters kernel has eventually chosen.

include/fuse_common.h
```diff
@@ -22,6 +22,7 @@
 
 #include "fuse_opt.h"
 #include "fuse_log.h"
+#include <stdbool.h>
```
ashleypittman, line 25: This is an installed header and it doesn't require this include to parse, the extra include should be in the c code where it's required rather than where it has visibility outside the project.

lib/fuse_lowlevel.c
```diff
@@ -3059,22 +3041,35 @@ int fuse_session_receive_buf_int(struct fuse_session *se, struct fuse_buf *buf,
...
 	err = errno;
 
 	if (fuse_session_exited(se))
 		return 0;
 	if (res == -1) {
+		if (err == EINVAL && buf->mem_size && se->bufsize > buf->mem_size)  {
+		    free(buf->mem);
+		    buf->mem = malloc(se->bufsize);
```
ashleypittman, line 3063: realloc would be preferable here I think, not because of the need to copy the buffer contents across but because on failure the old buffer is still allocated.

I know the changeable buffer size is new but is there any documentation or link what describes the kernel behaviour here?  At least a comment about why this is needed and when this code-path might be taken.

Finally, what's the value of bufsize here, hopefully not `UINT_MAX`?

ashleypittman (commented):
I agree that it's important to have this and it should be simple/transparent for the easy case but I'm also concerned about buffer management, in my case about ensuring network locality as well as on the URING side.  I was hoping that we could export the buffer management to the client for complex cases, perhaps via the custom_io code?

lib/fuse_lowlevel.c
```diff
@@ -1968,7 +1968,7 @@ static void do_lseek(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 /* Prevent bogus data races (bogus since "init" is called before
  * multi-threading becomes relevant */
 static __attribute__((no_sanitize("thread")))
-void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
+void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg, bool buf_reallocable)
```
ashleypittman, line 1971: I just don't see what's happening here with the code, it looks like you're passing a `size_t` to a `bool` parameter?

bsbernd: Yeah, conversion from size to bool, I had seen it, but I would like to avoid that parameter altogether.

ashleypittman: Agreed, I'd like to see a different solution here but I'm surprised this compiled.

bsbernd:
> I agree that it's important to have this and it should be simple/transparent for the easy case but I'm also concerned about buffer management, in my case about ensuring network locality as well as on the URING side. I was hoping that we could export the buffer management to the client for complex cases, perhaps via the custom_io code?

I have not done it yet in my uring branch, but the underlying server will be able to provide its own buffer - my eventual goal is zero-copy (for erase coding, checksum, compression that will probably need bpf).

lib/fuse_lowlevel.c
bsbernd: kernel side
```
static ssize_t fuse_dev_do_read(struct fuse_dev *fud, struct file *file,
				struct fuse_copy_state *cs, size_t nbytes)
{
	ssize_t err;
	struct fuse_conn *fc = fud->fc;
	struct fuse_iqueue *fiq = &fc->iq;
	struct fuse_pqueue *fpq = &fud->pq;
	struct fuse_req *req;
	struct fuse_args *args;
	unsigned reqsize;
	unsigned int hash;

	/*
	 * Require sane minimum read buffer - that has capacity for fixed part
	 * of any request header + negotiated max_write room for data.
	 *
	 * Historically libfuse reserves 4K for fixed header room, but e.g.
	 * GlusterFS reserves only 80 bytes
	 *
	 *	= `sizeof(fuse_in_header) + sizeof(fuse_write_in)`
	 *
	 * which is the absolute minimum any sane filesystem should be using
	 * for header room.
	 */
			   fc->max_write))
		return -EINVAL;


```

lib/fuse_lowlevel.c
joannekoong: > That introduces se->extended_max_write`. why can't we use 'se->conn.max_write?

We could use 'se->conn.max_write' to store the size, but the mount option will still need to be 'extended_max_write' since this would only need to be passed in for writes exceeding the default 1 MiB limit.

> I would also like to avoid the extra parameter for the do_init() function, if you don't want to store 'reallocable' into fuse_session, a local file variable should also do?

I'm fine with storing 'reallocable' as a field in 'struct fuse_session' instead of directly passing 'reallocable' as a parameter.

lib/fuse_lowlevel.c
joannekoong: The kernel max_write value will stay constant per connection (even if the sysctl value changes). The libfuse buffer must be large enough to accommodate max_write or the read will error out with -EINVAL.

ashleypittman: Is there anyway to read the value of max_write from userspace or have I missed that code somewhere?

joannekoong: The sysctl value is for max_pages, whihc you can read from userspace through `/proc/sys/fs/fuse/max_pages_limit`, and the kernel internally uses max_pages to determine max_write.

joannekoong: > Finally, what's the value of bufsize here, hopefully not UINT_MAX?

The value of bufsize here is the max write value that the user set in their init callback through "se->conn.max_write = x"

joannekoong:
Updates from v1:
* store "buf_reallocable" in fuse session struct instead of passing as parameter to do_init
* if reallocation fails due to out of memory issues, still preserve original buf->mem, since the caller is responsible for freeing it
*  move "mem_size" field to end of struct

bsbernd:
Thanks for the update @joannekoong ! On quick glance looks to good to me. Just getting late for me and I would like to go over it again in the morning.

bsbernd:
@joannekoong So the part I'm still not sure about is how I would know for over-io-uring which buffer size to use. Can we we keep the get_max_pages() function and use that for the buffer size in do_init(), instead of UINT_MAX? Well, maybe that needs to be discussed with @szmi differently, how libfuse shall figure out about the actually used buffer size.

joannekoong:
Thanks for taking a look @bsbernd! 

>  Can we we keep the get_max_pages() function and use that for the buffer size in do_init(), instead of UINT_MAX?

I think io-uring would run into the same issue where a machine might have the max pages set to a large number but only want large writes for one server (and not for the other 10 fuse servers running on the system). 

In do_init, the buffer size will never be UINT_MAX. It will be either a) the initial 256 MiB (eg USE_DEFAULT_MAX_PAGES_LIMIT * getpagesize()) max default value it was set to in fuse_session_new() or b) if the server sets a conn->max_write value in its init callback, then this conn->max_write value. 

My understanding is that it'd work the same way for the buffer size for io-uring (eg either 256 MiB or whatever max_write value the server set in its init callback). But maybe I'm missing something here that makes the io uring case different - please let me know if my understanding isn't right.

bsbernd:
@joannekoong Sorry for late replies, just about to send out a new fuse-uring patch version. 

Ah right, buffer size if limited to what fuse client sends as max_write, I can use that as upper limit for fuse-io-uring then. I don't want to do a ping pong of different buffer sizes there, issue is that with rings and many core we need even more memory - we do not want to have *all* ring entries using max_write. And with that we cannot bounce back ring entries with smaller buffers. Initially I wanted to configure different mem sizes with ioctls and had started work on that, until Miklos reviews came in and I had to give up on that. So new way (not part of the upstream patches) will be to send different buffer sizes on registration (FUSE_URING_REQ_FETCH) - that requires that we do not bounce back entries with smaller buffers.

PR #1070, opened 2024-11-20
Fix alignment of allocation in fuse_reply_create

phcoder:
When allocating as an array of char only alignment of 1 is guaranteed but the structure needs an alignment of 8. Allocate as an array of uint64 to ensure the correct alignment.

Without this fix I get a sporadical Bus error on arm.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -23,6 +23,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <stddef.h>
+#include <stdalign.h>
 #include <string.h>
 #include <unistd.h>
 #include <limits.h>
@@ -446,7 +447,7 @@ int fuse_reply_entry(fuse_req_t req, const struct fuse_entry_param *e)
 int fuse_reply_create(fuse_req_t req, const struct fuse_entry_param *e,
 		      const struct fuse_file_info *f)
 {
-	char buf[sizeof(struct fuse_entry_out) + sizeof(struct fuse_open_out)];
+	alignas(uint64_t) char buf[sizeof(struct fuse_entry_out) + sizeof(struct fuse_open_out)];
 	size_t entrysize = req->se->conn.proto_minor < 9 ?
 		FUSE_COMPAT_ENTRY_OUT_SIZE : sizeof(struct fuse_entry_out);
 	struct fuse_entry_out *earg = (struct fuse_entry_out *) buf;


lib/fuse_lowlevel.c
```diff
@@ -446,7 +446,7 @@ int fuse_reply_entry(fuse_req_t req, const struct fuse_entry_param *e)
 int fuse_reply_create(fuse_req_t req, const struct fuse_entry_param *e,
 		      const struct fuse_file_info *f)
 {
-	char buf[sizeof(struct fuse_entry_out) + sizeof(struct fuse_open_out)];
+	uint64_t buf[(sizeof(struct fuse_entry_out) + sizeof(struct fuse_open_out) + 7) / 8];
```
bsbernd, line 449: I would prefer to avoid magic numbers and I guess an ROUND_UP macro would do the job? From my uring branch, where I need it to allocate buffers:

```
#define ROUND_UP(val, round_to) \
	(((val) + (round_to - 1)) & ~(round_to - 1))
``

Comment (phcoder) [reply lib/fuse_lowlevel.c]: Done

File: lib/util.h [hunk 2]
```diff
@@ -1 +1,3 @@
 int libfuse_strtol(const char *str, long *res);
+
+#define DIV_ROUND_UP(a, b) (((a) + (b) - 1) / (b))
```
bsbernd, line 3: Why not my version, which is optimized for power of 2?

phcoder: Because I need to divide in addition to rounding up. As for optimization: it's precomputed by compiler, so optimization doesn't matter.

bsbernd: Sorry, I don't get it, this doesn't work?

```
uint8_t buf[DIV_ROUND(sizeof(struct fuse_entry_out) + sizeof(struct fuse_open_out), sizeof(uint64_t))];
```

(and we could also keep the char type)

phcoder: No it doesn't. The problem is not the size but the alignment of resulting buffer. Using uint64_t type is the usual way to ensure correct alignment of the buffer. Different expression for the size is only a side effect of using uint64_t type

bsbernd: Ah yes, sorry, I'm totally focused on something else.  We are using C11 - the more natural way is to use 

```
alignas(sizeof(uint64_t)) char buf[sizeof(struct fuse_entry_out) + sizeof(struct fuse_open_out)];
```

?

bsbernd: Well, maybe not "more natural", but better to read and harder to break with whatever future changes.

phcoder: alignas needs either C++11 or C23. C11 is not good enough. I tried with __attribute__((__align__(8))) but it's ignored on variables

phcoder: Alternatively I could rewrite it using structs and unions

bsbernd:
```
index d00abf6efef6..cbbdbcc4de6b 100644
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -29,6 +29,7 @@
 #include <assert.h>
 #include <sys/file.h>
 #include <sys/ioctl.h>
+#include <stdalign.h>
 
 #ifndef F_LINUX_SPECIFIC_BASE
 #define F_LINUX_SPECIFIC_BASE       1024
@@ -471,7 +472,8 @@ int fuse_reply_entry(fuse_req_t req, const struct fuse_entry_param *e)
 int fuse_reply_create(fuse_req_t req, const struct fuse_entry_param *e,
                      const struct fuse_file_info *f)
 {
-       char buf[sizeof(struct fuse_entry_out) + sizeof(struct fuse_open_out)];
+       alignas(sizeof(uint8_t)) char buf[sizeof(struct fuse_entry_out) +
+                                         sizeof(struct fuse_open_out)];
        size_t entrysize = req->se->conn.proto_minor < 9 ?
                FUSE_COMPAT_ENTRY_OUT_SIZE : sizeof(struct fuse_entry_out);
        struct fuse_entry_out *earg = (struct fuse_entry_out *) buf;
```

Are you sure? My clang stdalign.h has
```
/*===---- stdalign.h - Standard header for alignment ------------------------===
 *
 * Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 * See https://llvm.org/LICENSE.txt for license information.
 * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 *
 *===-----------------------------------------------------------------------===
 */

#ifndef __STDALIGN_H
#define __STDALIGN_H

#if defined(__cplusplus) ||                                                    \
    (defined(__STDC_VERSION__) && __STDC_VERSION__ < 202311L)
#ifndef __cplusplus
#define alignas _Alignas
#define alignof _Alignof
#endif

#define __alignas_is_defined 1
#define __alignof_is_defined 1
#endif /* __STDC_VERSION__ */

#endif /* __STDALIGN_H */

```

Dunno when llvm/clang got it, gcc has it since 4.8
https://gcc.gnu.org/gcc-4.8/changes.html

bsbernd: Just need to leave keyboard for about 2.5 hours, I can test when I'm back if it works

ashleypittman: Does `_Alignas` work?

bsbernd: _Alignas should work, I just don't like to use keywords are already deprecated. clang and gcc should already have the wrappers for alignas

phcoder: With stdalign it seems to work. Will upload new version

bsbernd:
Thank you @phcoder !

PR #1071, opened 2024-11-21
Support O_TMPFILE

hbirth:
Support O_TMPFILE in FUSE low level API
- Note that only support for the low level API is added since the lookup and name hashes as well as the API itself makes files without names very hard to handle. 
-  add tests for tmpfile creation

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -914,6 +914,86 @@ static void sfs_create(fuse_req_t req, fuse_ino_t parent, const char *name,
     fuse_reply_create(req, &e, fi);
 }
 
+static Inode *create_new_inode(int fd, fuse_entry_param *e)
+{
+    memset(e, 0, sizeof(*e));
+    e->attr_timeout = fs.timeout;
+    e->entry_timeout = fs.timeout;
+
+    auto res = fstatat(fd, "", &e->attr, AT_EMPTY_PATH | AT_SYMLINK_NOFOLLOW);
+    if (res == -1) {
+        if (fs.debug)
+            cerr << "DEBUG: lookup(): fstatat failed" << endl;
+        return NULL;
+    }
+
+    SrcId id {e->attr.st_ino, e->attr.st_dev};
+    unique_lock<mutex> fs_lock {fs.mutex};
+    Inode* p_inode;
+    try {
+        p_inode = &fs.inodes[id];
+    } catch (std::bad_alloc&) {
+        return NULL;
+    }
+
+    e->ino = reinterpret_cast<fuse_ino_t>(p_inode);
+    e->generation = p_inode->generation;
+
+    lock_guard<mutex> g {p_inode->m};
+    p_inode->src_ino = e->attr.st_ino;
+    p_inode->src_dev = e->attr.st_dev;
+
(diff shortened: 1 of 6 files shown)


example/passthrough_hp.cc
```diff
@@ -914,6 +914,86 @@ static void sfs_create(fuse_req_t req, fuse_ino_t parent, const char *name,
...
+    lock_guard<mutex> g {inode->m};
+
+    sfs_create_open_flags(fi);
+
+    if (fs.passthrough)
+        do_passthrough_open(req, e.ino, fd, fi);
+ 
+    fuse_reply_create(req, &e, fi);
+}
+
```
bsbernd, line 996: I think it would be possible to avoid code dup with do_lookup() - do_lookup could call into this new function.  Though some parts probably have to be moved into sfs_tmpfile(), like sfs_create_open_flags..fuse_reply_create

test/test_syscalls.c
```diff
@@ -1957,6 +1957,90 @@ static int do_test_create_ro_dir(int flags, const char *flags_str)
...
+
+	int res = mkdir(testdir, 0777);
+	if (res)
+		return -1;
+
+	start_test("create and link tmpfile");
+
+	int fd = open(testdir, O_TMPFILE | O_RDWR | O_EXCL, S_IRUSR | S_IWUSR);
+	if(fd == -1) {
+		PERROR("open");
```
bsbernd, line 2009: Could you add a bit more descriptive error message, like `open(, O_TMPFILE`)?

test/test_syscalls.c
```diff
@@ -1957,6 +1957,90 @@ static int do_test_create_ro_dir(int flags, const char *flags_str)
...
+	}
+	close(fd);
+
+	fd = open(testdir, O_TMPFILE | O_WRONLY | O_EXCL, S_IRUSR | S_IWUSR);
+	if(fd == -1){
+		PERROR("open");
+		return -1;
+	};
+	close(fd);
+
```
bsbernd, line 1985: Would be good if we could differentiate the two open calls - add in the flags into the message?

test/test_syscalls.c
hbirth: done

example/passthrough_hp.cc
hbirth: done

test/test_syscalls.c
hbirth: done

example/passthrough_hp.cc
bsbernd: Ok, let's leave keep it as it is for now.

test/test_syscalls.c
```diff
@@ -1971,6 +1971,12 @@ static int test_create_tmpfile(void)
 
 	int fd = open(testdir, O_TMPFILE | O_RDWR, S_IRUSR | S_IWUSR);
 	if(fd == -1) {
+		if (errno == ENOTSUP) {
+			/* don't bother if we're working on an old FUSE 
+					or on the high level API */
```
bsbernd, line 1976: Could you update to "an old kernel"?

PR #1081, opened 2024-12-28
Increase the .so version, but also change back to previous ABI

bsbernd:
There had been a slight ABI breakage in 3.11 (wrong position of struct fuse_config::no_rofd_flush), with a similar issue in 3.14.1 (wrong position of struct fuse_config::parallel_direct_writes). Much worse in 3.14. is struct fuse_file_info::parallel_direct_writes which has the risk of causing unintended behavior (which could lead in worst cast to data corruption).
abi-compliance-checker also annotates in master/3.17 several issue when comparing to 3.16.2, although most of these are false positives.
The main ABI breakage in 3.14.1 could be handled with heuristics, but it would be not beautiful and we wouldn't know when we could disable it. Only the upcoming 3.17 has encoded the version into the lib to be prepared to handle such situations. For now we better go with an so version bump to enforce recompilation.

This PR also tries to keep binary compatibility with old libfuse versions (before 3.11), mainly to help RHEL, which seems to have support for binaries that cannot be easily recompiled.

### DIFF
--- a/example/passthrough_ll.c
+++ b/example/passthrough_ll.c
@@ -168,18 +168,20 @@ static bool lo_debug(fuse_req_t req)
 static void lo_init(void *userdata,
 		    struct fuse_conn_info *conn)
 {
-	struct lo_data *lo = (struct lo_data*) userdata;
+	struct lo_data *lo = (struct lo_data *)userdata;
+	bool has_flag;
 
-	if (lo->writeback &&
-	    conn->capable & FUSE_CAP_WRITEBACK_CACHE) {
-		if (lo->debug)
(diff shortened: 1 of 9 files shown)


bsbernd:
@szmi This is the alternative PR with binary compatibility to https://github.com/libfuse/libfuse/pull/1038

bsbernd:
ABI failure is expected, as it compares to the previous commit in master branch and not to 3.10.

bsbernd:
@ashleypittman @trapexit I'm tempted to merge this ASAP and let that then be 3.17-rc1

trapexit (approved):

PR #1082, opened 2024-12-30
Update the version to 4.0

bsbernd:
We might just follow linux kernel scheme and avoid high versions, by arbitrary increasing libfuse versionsn.
Additioanlly we have an so version bump - increasing to libfuse4 is even more justified.

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -887,10 +887,10 @@ int fuse_main_real(int argc, char *argv[], const struct fuse_operations *op,
 		   size_t op_size, struct libfuse_version *version,
 		   void *user_data);
 #else
-int fuse_main_real_317(int argc, char *argv[], const struct fuse_operations *op,
+int fuse_main_real_400(int argc, char *argv[], const struct fuse_operations *op,
 		   size_t op_size, struct libfuse_version *version, void *user_data);
 #define fuse_main_real(argc, argv, op, op_size, version, user_data) \
-	fuse_main_real_317(argc, argv, op, op_size, version, user_data);
+	fuse_main_real_400(argc, argv, op, op_size, version, user_data);
 #endif
 
 /**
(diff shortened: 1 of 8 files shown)


bsbernd:
@trapexit @ashleypittman @szmi @Nikratio What do you think about a version bump to 4.0 in addition to the so bump?

trapexit:
If the library is still compatible with v3.x I'm not sure it's a great idea to bump the major version. That said I think the best would be to get feedback from the distro packagers. I can imaging changing the package name to libfuse4 will cause issues. People are still having problems with libfuse2 -> 3. Don't want to annoy maintainers further.

bsbernd:
I think the issue with libfuse2 -> 3 is due to incompatible API changes  - we should really avoid that. 
Bumping the major version should be similar to other version upgrades. I'm actually thinking a bit that it might help package maintainers and users that still have binary dependencies on previous libfuse versions.

trapexit:
Yes, it was the API change... but I wasn't referring to that. I am referring to the general trouble it was for the package maintainers. That lots of projects still haven't moved and parallel installs of the libraries was not smooth. I don't want any other troubles for maintainers. 

Are you saying that libfuse 4.x would be installed instead of libfuse 3.x but be compatible with 3.x? From a packaging perspective these packages are named "libfuse3" which would mean they need to move to a package "libfuse4" which would be confusing for those packages expecting "libfuse3" or they leave the package named "libfuse3" and that would be confusing too.

bsbernd:
I don't think renaming to libfuse4 is needed, as is it still API compatible. Something like libfuse-4.0 would be sufficient.

Nikratio:
I would bump the major version only if backwards compatibility is broken, i.e. programs that compiled fine against libfuse3 do not compile against libfuse4 without changes.

(This is assuming that any ABI changes can be handled transparently. If we're in a situation where the API is the same, but programs compiled against libfuse3 will fail at runtime if linked against current libfuse, then I'm not sure. Maybe in that case I would deliberately break the API as well to make the transition explicit).

trapexit:
> I don't think renaming to libfuse4 is needed, as is it still API compatible. Something like libfuse-4.0 would be sufficient.

https://packages.debian.org/sid/libfuse3-3

They are already named libfuse2 and libfuse3.

DrDaveD:
> > I don't think renaming to libfuse4 is needed, as is it still API compatible. Something like libfuse-4.0 would be sufficient.
> 
> https://packages.debian.org/sid/libfuse3-3
> 
> They are already named libfuse2 and libfuse3.

On RHEL it is fuse3-libs-3.3.0, so it would need to be fuse3-libs-4.0.0, which would be extremely confusing.  Please do not change the major version.

bsbernd:
Ok, I drop the major version bump. Thanks everyone for comments!

@Nikratio I increased the .so version, as 3.11 and 3.14.2 had changes that accidentally broke the ABI. I reverted to the previous ABI as part of PR #1081, but there are versions in between. 
3.17 has a patch to encode the linked version into the libfuse user and there is also an ABI test now - I hope ABI issues will never come up again.

PR #1083, opened 2025-01-01
passthrough_ll: set correct keep_cache flag in opendir

amir73il:
In cache=always mode, set keep_cache flag in opendir(), same as done in open() and same as passthrough_hp does in opendir().

In the default cache=auto mode, use readdir cache, but do not set keep_cache, same as regular files use page cache for an open file, but do not keep_cache for a new open.

Note that passthrough_hp by default behaves the same as passthrough_ll cache=always mode and supports the cache=never mode with --nocache, but it does not support the equivalent of cache=auto mode.


bsbernd (approved):
Thanks Amir!

PR #896, opened 2024-02-25
fix readdirplus when filler is called with zero offset

farlongsignal:
fixes #235 
In fill_dir_plus(), there's a lookup for caching dirent attributes. However, when offset = 0 the cache metadata from the lookup is lost as only the entry attributes are passed when added to the list. Kernel probably ignores the attributes since .ino = 0.
With the following PR i can finally witness file attributes being cached and getattr() not being called when running `ls -l`.

PR #1088, opened 2025-01-02
Fix readdirplus regression

amir73il:


amir73il:
@farlongsignal please review fix of regression caused by #896

lib/fuse.c
```diff
@@ -3691,11 +3697,12 @@ static int readdir_fill_from_list(fuse_req_t req, struct fuse_dh *dh,
 
 		if (flags & FUSE_READDIR_PLUS) {
 			struct fuse_entry_param e = {
-				.ino = 0,
+				.ino = de->ino,
```
bsbernd, line 3700: I will look into this patch immediately in the morning, but I'm confused about setting .ino here. `do_lookup`  will overwrite it with `e->ino = node->nodeid;`?

bsbernd: Ah sorry, in the lines below.

bsbernd:
Hmm, I expect a bit junk results when FUSE_FILL_DIR_PLUS is used with invalid stat info, but it actually also happens without that flag, which is actually also what your command line is doing.

lib/fuse.c
```diff
@@ -163,6 +163,7 @@ struct node_lru {
 
 struct fuse_direntry {
 	struct stat stat;
+	fuse_ino_t ino;
```
bsbernd, line 166: The part I don't like is

"FUSE_UNKNOWN_INO" == "enable-readdir-plus" and "0" == "disable-readdir-plus"


But I don't have a better idea yet either.

bsbernd: Maybe 
```
#define FUSE_INO_NEEDS_LOOKUP FUSE_UNKNOWN_INO
```

and
```
/* disables readdir-plus in fuse-kernel for these entries */
#define FUSE_INO_NONE 0
```

?

amir73il: We don't really need to store ino in the list. We can store instead just bool plus instead so it won't look like a hack.

bsbernd:
Thanks a lot @amir73il! Looks really great now.

PR #1089, opened 2025-01-02
Fix libfuse build with FUSE_USE_VERSION 30

amir73il:
Builds fail for high level filesystem that defines FUSE_USE_VERSION 30


bsbernd (approved):
Thanks a lot Amir!

PR #1091, opened 2025-01-03
Fix build of memfs_ll without manual meson reconfigure

amir73il:
After pulling latest code, memfs_ll build would fail because it builds with C++11.

Changing the default of cpp_std in meson.build is not enough to fix this problem even if user runs 'meson setup --reconfigure'.

I had to run 'meson setup -Dcpp_std= --reconfigure' to fix the build as mentioned in this meson issue:
https://github.com/mesonbuild/meson/issues/8062#issuecomment-1568249672


bsbernd:
Ah thanks @amir73il . I admit that I had noticed it, but it was just sufficient to entire delete and then to recreate the build dir - after that worked I didn't further looked into it.

Issue #1092, opened 2025-01-04
Changing signatures of exported symbols


billziss-gh:
PR #942 / commit 125b86b7be62744efc953f8bd73089328dcdcce1 changed the signatures of exported symbols like `fuse_main_real`.

Previously the signature of `fuse_main_real` was:

```C
int fuse_main_real(int argc, char *argv[], const struct fuse_operations *op,
		   size_t op_size, void *private_data);
```

After that PR/commit the signature of `fuse_main_real` became:

```C
int fuse_main_real(int argc, char *argv[], const struct fuse_operations *op,
		   size_t op_size, struct libfuse_version *version,
//                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ added
		   void *user_data);
```

This can create problems for software that needs to use libfuse dynamically (i.e. via `dlopen` / `dlsym`). Since there is no exported `fuse_main` symbol, such software is obligated to use the exported symbol `fuse_main_real`. If such software picks up an exported symbol with a changed signature, the passed arguments will not match which will lead to a crash, corruption, etc.

**I ask that exported symbol signatures are treated as ABI and do not experience any further changes.**

Copying the original project maintainer and people involved in the PR/commit:

@Nikratio @bsbernd @ashleypittman

bsbernd:
It actually does not

```
bschubert2@imesrv6 build-ubuntu>nm lib/libfuse3.so | grep fuse_main_real
00000000000203f0 t fuse_main_real_30
0000000000020120 T fuse_main_real_317
00000000000203f0 T fuse_main_real@FUSE_3.0
0000000000020120 T fuse_main_real@@FUSE_3.17
```

The `fuse_main_real@FUSE_3.0` function is actually compatible and does not take the additional version argument. I'm not an expert with versioned dlopen/dsym and versioned symbols, but AFAIK, you have to use `dlvsym`, i.e.
`dlvsym(handle, "fuse_main_real", "FUSE_3.0");`


That being said, there is a bug in the commit you mentioned and it forgot to export  `fuse_main_real` for libc/systems that have libfuse without versioned symbols. Good that 3.17 is not finally released yet (the branch exists already, but will get recent commits from master shortly).

billziss-gh:
I am no expert on versioned symbols either, but my understanding is that `@@` designates the default symbol:

https://sourceware.org/binutils/docs/ld/VERSION.html

[...]

PR #1093, opened 2025-01-04
Fixes for commit 58f85bfa9b7d "Add in the libfuse version a program was compiled with"

bsbernd:
1) Some compat functions for libfuse without versioned symbols were missing

2) Related to issue #1092 we should avoid exporting any function that does not absolutely need to be exported - some random users might pick it up and use it. Same goes for function declarations - internal functions should not be declared globally. Too late for fuse_main_real and other as used in issue #1092  , but at least for new functions we can avoid that.


bsbernd:
The ABI failure is expected, as we remove two functions from exports - these are not part of a releases ABI yet.

Issue #432, opened 2019-07-08
fuse "-o auto_unmount" doesn't close fds[1] upon exiting the fuse loop


DavidButterfield:
fuse "-o auto_unmount" doesn't close fds[1] (from the socketpair() call in
fuse_mount_fusermount) upon exiting the fuse loop.  So valgrind reports it as an
unclosed file descriptor on exit.

Perhaps more importantly, the fusermount3 process remains running with the fuse
filesystem mounted even though the fuse thread has shut down.  That situation
could remain indefinitely -- the program is not necessarily terminating just
because it ends the fuse thread. 

==26642== Open AF_UNIX socket 6: <unknown>
==26642==    at 0x4A5314A: socketpair (syscall-template.S:78)
==26642==    by 0x48E7F6E: ??? (in /lib/x86_64-linux-gnu/libfuse.so.2.9.9)
==26642==    by 0x48E8BE3: ??? (in /lib/x86_64-linux-gnu/libfuse.so.2.9.9)
==26642==    by 0x48E47D7: ??? (in /lib/x86_64-linux-gnu/libfuse.so.2.9.9)
==26642==    by 0x48E48B6: ??? (in /lib/x86_64-linux-gnu/libfuse.so.2.9.9)
==26642==    by 0x48E49EA: ??? (in /lib/x86_64-linux-gnu/libfuse.so.2.9.9)
==26642==    by 0x310262: UMC_fuse_run (UMC_fuse.c:465)
==26642==    by 0x486DE7D: sys_thread_fn (mte_service.c:145)
==26642==    by 0x491C181: start_thread (pthread_create.c:486)
==26642==    by 0x4A51B1E: clone (clone.S:95)

Nikratio:
Thanks for the report! Do you think you could work on this (or do you know someone who could)?

libfuse is a fully volunteer driven project. This means that new features and bugfixes are implemented only when someone has a personal interest in them and therefore also does the necessary
work. So if someone (that probably means you) is interested in working on this, then I'd be happy to give guidance and review patches. If not, I will close this issue in a few days (since the purpose of the issue tracker is to coordinate ongoing and planned work, rather than enumerate all the work that could possibly be done).

bsbernd:
I think this is already fixes by PR 1033.

PR #1094, opened 2025-01-08
Identify the FD hold by parent process

zegang:
### Issue:
There is no directly way to get the FD hold by parent process which asked do fuse mount.
Known Issue on the FD: [fuse "-o auto_unmount" doesn't close fds[1] upon exiting the fuse loop](https://github.com/libfuse/libfuse/issues/432#top)

### Use Case:
For auto_unmount case, identify the FD can easy to close the FD and make automatically unmount manually and explicitly.
In auto-unmount case, parent process can close this FD explicitly to do unmount.
The FD[1] can be got via `getenv(FUSE_COMMFD2_ENV)`.
One potential use case is to satisfy FD-Leak checks.

### Solution:
Add an extra env _FUSE_COMMFD2 to store the FD. This will provide a easy way to get the FD via FUSE_COMMFD2_ENV.

### DIFF
--- a/lib/mount.c
+++ b/lib/mount.c
@@ -48,6 +48,7 @@
 
 #define FUSERMOUNT_PROG		"fusermount3"
 #define FUSE_COMMFD_ENV		"_FUSE_COMMFD"
+#define FUSE_COMMFD2_ENV	"_FUSE_COMMFD2"
 
 #ifndef MS_DIRSYNC
 #define MS_DIRSYNC 128
@@ -367,6 +368,14 @@ static int setup_auto_unmount(const char *mountpoint, int quiet)
 	char arg_fd_entry[30];
 	snprintf(arg_fd_entry, sizeof(arg_fd_entry), "%i", fds[0]);
 	setenv(FUSE_COMMFD_ENV, arg_fd_entry, 1);
+	/*
+	 * This helps to identify the FD hold by parent process.
+	 * In auto-unmount case, parent process can close this FD explicitly to do unmount.
+	 * The FD[1] can be got via getenv(FUSE_COMMFD2_ENV).
+	 * One potential use case is to satisfy FD-Leak checks.
+	 */
+	snprintf(arg_fd_entry, sizeof(arg_fd_entry), "%i", fds[1]);
+	setenv(FUSE_COMMFD2_ENV, arg_fd_entry, 1);
 
 	char const *const argv[] = {
 		FUSERMOUNT_PROG,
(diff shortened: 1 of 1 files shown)


bsbernd:
I'm missing the user for this new env. Also note that I would like to depcrecate the envs, fusermount.c already got the option 
```
		case 'c':
			commfd = optarg;
			break;
```

We are just not using it for now as old versions of libfuse might be installed. For new variables that issue does not exist - the parameter can easily be passed.

Also see PR #1033 - all inherited FDs are closed.

zegang:
> I'm missing the user for this new env. Also note that I would like to depcrecate the envs, fusermount.c already got the option
> 
> ```
> 		case 'c':
> 			commfd = optarg;
> 			break;
> ```
> 
> We are just not using it for now as old versions of libfuse might be installed. For new variables that issue does not exist - the parameter can easily be passed.
> 
> Also see PR #1033 - all inherited FDs are closed.

The `commfd` is to refer to the first FD created by `socketpair()`.

In common code path to do fuse-mount utilizing libfuse, it is like this:

1. Parent-Process: call `fuse_mount()`;
2. libfuse `fuse_mount()`: create 2 FDs (`FD[0] 151, FD[1] 152`) by `socketpair()` and `fork()` a Child-Process to do real fuse-mount:
- 2.1 Child-Process closes `FD[1] 152` and keeps `FD[0] 151` for communications;
- 2.2 Parent-Process closes  `FD[0] 151` and keeps `FD[1] 152` as open in `auto_unmout` case;
- 2.3 The FD array is a function internal on-stack variable;
3. Parent-Process does not know the `FD[1] 152`. This is okay for most case as it will be closed when the Parent-Process exits.

However, in some cases, Parent-Process wants to know the `FD[1] 152` and to do close **explicitly** to make the Child-Process do unmount automatically.

My understanding is the option `c` and `commfd` is designed to store the `FD[0] 151`.

bsbernd:
Who is going to read '/proc/pid/environ' and check for FUSE_COMMFD2_ENV? 

>My understanding is the option c and commfd is designed to store the FD[0]
Yes, but we could add an option for FD[1]. I just don't understand which process is going to use the FUSE_COMMFD2_ENV

zegang:
> Who is going to read '/proc/pid/environ' and check for FUSE_COMMFD2_ENV?
> 
> > My understanding is the option c and commfd is designed to store the FD[0]
> > Yes, but we could add an option for FD[1]. I just don't understand which process is going to use the FUSE_COMMFD2_ENV

Any one wants to know the FD hold by the Parent-Process. The Parent-Process is the one who can close the FD close explicitly to make the Child-Process do unmount automatically. Then Parent-Process can do its customized resource usage or resource leak checks, e.g. all FDs are closed and no FD leak.

bsbernd:
I would like to see in the commit message what the FD is being used for - you seem to have a use case, but somehow are not revealing it.

zegang:
> I would like to see in the commit message what the FD is being used for - you seem to have a use case, but somehow are not revealing it.

Thanks @bsbernd. Add the Use Case in Codes and Commit comments.

bsbernd:
@zegang Thank you, could please fix checkpatch warnings (signed-off and commit message length)?

Also, I have to admit that I'm personally suspicious about the libfuse internal umount - assuming someone did "mount --bind <new-dir> &&  umount <old-dir>"  or "mount --move" and then the libfuse dir would another mount, libfuse would unmount the wrong mount....

zegang:
> Also, I have to admit that I'm personally suspicious about the libfuse internal umount - assuming someone did "mount --bind && umount " or "mount --move" and then the libfuse dir would another mount, libfuse would unmount the wrong mount....

Any Tests in libfuse to test and verify on your "suspicious"?

bsbernd:
> > Also, I have to admit that I'm personally suspicious about the libfuse internal umount - assuming someone did "mount --bind && umount " or "mount --move" and then the libfuse dir would another mount, libfuse would unmount the wrong mount....
> 
> Any Tests in libfuse to test and verify on your "suspicious"?

No, there are no tests for that, but given it works on the path - broken by design. If I would have been maintainer by the time patches for that feature came in, I wouldn't have let through these patches... Too late now... At best we can make it an LSFMM topic, unless the new mount API already handles it.

bsbernd:
Thanks for the update @zegang . We ignore the current checkpatch message about mail mismatch - we need to disable that check.

PR #1101, opened 2025-01-23
Update include/fuse_kernel.h to version 7.43

joannekoong:
Sync include/fuse_kernel.h with the most up to date fuse uapi headers in the kernel.

PR #1102, opened 2025-01-23
Support request timeouts

joannekoong:
This adds the libfuse changes needed to support request timeouts.
A timeout may be set by the server in its init call. If a request is not
completed by the timeout, the connection will be aborted by the kernel.

Signed-off-by: Joanne Koong <joannelkoong@gmail.com>

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -709,9 +709,15 @@ struct fuse_conn_info {
 	uint64_t want_ext;
 
 	/**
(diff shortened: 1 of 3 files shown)


joannekoong:
This is rebased on top of [PR 1101]( https://github.com/libfuse/libfuse/pull/1101)

bsbernd:
Thanks @joannekoong. The abi warning looks like a false-positive - ignored.

PR #1108, opened 2025-01-31
Fix possible 64 bits value truncation

luis-henrix:
The first patch fixes a 64 bit value truncation.

The second patch is just a trivial change which removes an incorrect comment.


bsbernd:
Thank you @luis-henrix !

PR #1112, opened 2025-02-09
Add an ABI check to also compare to 3.10 as baseline

bsbernd:


bsbernd:
Pity, seems to be impossible to ignore some parts - all attempts do not work, giving up.

PR #1114, opened 2025-02-10
Update version to 3.17.1-rc0

bsbernd:
I should have used 3.17.0-rc0 previously - 3.17.0 will never be released.


bsbernd:
Actually does not work for `make dist` - not suitable

PR #1122, opened 2025-02-12
passthrough_hp: enable splice move by default

joannekoong:
This commit enables splice moves by default, unless the caller opts into nosplice.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -165,7 +165,7 @@ static Fs fs{};
 #define FUSE_BUF_COPY_FLAGS                      \
         (fs.nosplice ?                           \
             FUSE_BUF_NO_SPLICE :                 \
-            static_cast<fuse_buf_copy_flags>(0))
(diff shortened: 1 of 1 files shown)


example/passthrough_hp.cc
```diff
@@ -165,7 +165,7 @@ static Fs fs{};
 #define FUSE_BUF_COPY_FLAGS                      \
         (fs.nosplice ?                           \
             FUSE_BUF_NO_SPLICE :                 \
-            static_cast<fuse_buf_copy_flags>(0))
+            static_cast<fuse_buf_copy_flags>(FUSE_BUF_SPLICE_MOVE))
```
bsbernd, line 168: Hmm, I wonder if you make this more generic, if fuse_buf_copy() should automatically pick it up from CAP features. At least something to think about.

bsbernd (approved):
Thank you Joanne! I had seen your findings on fsdevel.

PR #1126, opened 2025-02-14
lib/fuse.c: fix release, opendir, and open callbacks if op is not implemented by server

joannekoong:
For release, opendir, and open, if the server has not implemented the callback, this should return back -ENOSYS.


joannekoong:
Will investigate the test failure

joannekoong:
Marking this as closed. Looking closer, doesn't seem these should return -ENOSYS if not defined.

PR #1130, opened 2025-02-16
Fix build of example/memfs_ll.cc on 32 bit architectures

gcsideal:
The code uses std::min() which expects its arguments to be size_t. Two times it uses an offset declared as off_t. While both size_t and off_t are 32-bit integers, the latter are signed. On 64 bit architectures the conversation of off_t -> size_t performed automatically. On 32 bit architectures it needs a type cast.

### DIFF
--- a/example/memfs_ll.cc
+++ b/example/memfs_ll.cc
@@ -150,7 +150,7 @@ class Inode {
 
 	void read_content(char *buf, size_t size, off_t offset) const
 	{
-		size_t bytes_to_read = std::min(size, content.size() - offset);
+		size_t bytes_to_read = std::min(size, content.size() - (size_t)offset);
 		std::copy(content.begin() + offset,
 			  content.begin() + offset + bytes_to_read, buf);
 	}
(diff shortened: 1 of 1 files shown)


bsbernd:
Ooops, looks like we need a 32bit compilation test. Thanks for fixing it. Would you mind to add your Signed-off-by?

bsbernd:
Interesting that so far other issues didn't come up up, see here https://github.com/libfuse/libfuse/pull/1132/commits

gcsideal:
> Ooops, looks like we need a 32bit compilation test. Thanks for fixing it. Would you mind to add your Signed-off-by?

Updated the commit message as requested. Meanwhile this is filed separately as issue 1129 by an other user.
The results of the [build try of 3.17.1~rc0 is here](https://buildd.debian.org/status/package.php?p=fuse3&suite=experimental). If you need [the i386 log directly](https://buildd.debian.org/status/fetch.php?pkg=fuse3&arch=i386&ver=3.17.1%7Erc0-1&stamp=1739643155&raw=0).

PR #1133, opened 2025-02-17
FreeBSD stuff

vassilit:
some comments:

#### tests: Re-enable mknod and mkfifo tests on FreeBSD
On FreeBSD, device nodes may be created in file systems with mknod() but such nodes cannot be used to access devices.

#### mount_bsd: Fix usage of libfuse_strtol
Note that on platforms where long and int are of different sizes, fd could be truncated when passed to the OS. That is true for util/fusermount.c as well.

Before checking for fd < 0:
```
root@fbsd14-ufs:/home/vt/dev/fuse-archive # FUSE_DEV_FD=-1 out/fuse-archive -o debug test/data/archive.tar
fuse-archive 53467 - - Created mount point 'archive'
FUSE library version: 3.17.1-rc0
nullpath_ok: 0
mount_fusefs: invalid option -- 1
usage:
mount_fusefs [-A|-S|-v|-V|-h|-D daemon|-O args|-s special|-m node|-o option...] special node [daemon args...]
[...]
```

#### FreeBSD: Remove useless options
See https://cgit.freebsd.org/src/commit/?id=4abf87666ad69ba377a7d0aa71e87bc8824759fa
Since FreeBSD 13-RELEASE, `sysctlbyname("vfs.fuse.init_backgrounded")` returned ERR#2 'No such file or directory', so `init_backgrounded()` always returned 0.

#### mount_bsd: Remove the double-fork
This is the one I'm the less confident with.

After my naive unanswered [question](https://github.com/libfuse/libfuse/discussions/1100), I've tested by myself with [fuse-test-ready.c](https://gist.github.com/vassilit/e00d59135e3204aabcc4cca8c333c140).

It appears that with the double-fork, the FS is ready between 150µs and 1ms after returning from `fuse_main()`, and between 20µs and 100µs without.

However, I might be missing something. I'm running a patched version of libfuse2 without the double-fork for more than a month and I never got dead-locked. I've also tested successfully a libfuse3 version (without double-fork).
In particular I've run the test suite of fuse-archive many times and it never failed. 

What worries me a bit is that the code from the time `mount_bsd.c` was written has not changed much:
- Fuse for FreeBSD before inclusion into the mainline, as it was when mount_bsd.c was written: [here](https://github.com/glk/fuse-freebsd/blob/5c9d7f4881d72c7b2229989f3e59ab5266ac0b09/fuse_module/fuse_vfsops.c#L512).
- `mount_fusefs` call the `nmount` syscall, it then goes [here](https://github.com/freebsd/freebsd-src/blob/408394185fe9be0c158189b7671bde316ac14ac9/sys/fs/fuse/fuse_vfsops.c#L447) to send FUSE_INIT.

So the dead-lock, if existed, could possibly still exist, so maybe I am overlooking something. But I've never met the dead-lock in practice. And people are reporting here and there that the FS is not ready after returning from `fuse_main()`, so avoiding the double-fork and hence waiting for mount_fusefs, while not resolving the problem entirely, reduce it.

On [OSX](https://github.com/zargony/fuse-rs/issues/9#issuecomment-32167809), their analysis is « It then synchronously waits for the FUSE_INIT ». Is it specific for OSX ? It seems very similar to FreeBSD.
I've check with `truss`/`dtrace`/`dwatch`,  indeed, there is a lot of \*stat\* calls during the mount, but removing the double-fork never blocked them.

If unsure, feel free to remove this commit while waiting from more inputs.

Ideally, we could call `(n)mount()` directly in the future, but that would not solve the double-fork problem.

### DIFF
--- a/lib/mount_bsd.c
+++ b/lib/mount_bsd.c
@@ -19,7 +19,6 @@
 
 #include <sys/stat.h>
 #include <sys/wait.h>
-#include <sys/sysctl.h>
 #include <sys/user.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -86,7 +85,6 @@ static const struct fuse_opt fuse_mount_opts[] = {
 	FUSE_DUAL_OPT_KEY("private",		KEY_KERN),
 	FUSE_DUAL_OPT_KEY("neglect_shares",	KEY_KERN),
 	FUSE_DUAL_OPT_KEY("push_symlinks_in",	KEY_KERN),
-	FUSE_OPT_KEY("nosync_unmount",		KEY_KERN),
 #if __FreeBSD_version >= 1200519
 	FUSE_DUAL_OPT_KEY("intr",		KEY_KERN),
 #endif
@@ -134,21 +132,6 @@ void fuse_kern_unmount(const char *mountpoint, int fd)
 	unmount(mountpoint, MNT_FORCE);
 }
 
-/* Check if kernel is doing init in background */
-static int init_backgrounded(void)
-{
-	unsigned ibg;
-	size_t len;
-
-	len = sizeof(ibg);
-
-	if (sysctlbyname("vfs.fuse.init_backgrounded", &ibg, &len, NULL, 0))
-		return 0;
-
-	return ibg;
-}
-
-
 static int fuse_mount_core(const char *mountpoint, const char *opts)
 {
 	const char *mountprog = FUSERMOUNT_PROG;
@@ -194,20 +177,12 @@ mount:
 	}
 
 	if (pid == 0) {
-		if (! init_backgrounded()) {
-			/*
-			 * If init is not backgrounded, we have to
-			 * call the mount util backgrounded, to avoid
-			 * deadlock.
-			 */
+		pid = fork();
 
-			pid = fork();
-
-			if (pid == -1) {
-				perror("fuse: fork() failed");
-				close(fd);
-				exit(1);
-			}
+		if (pid == -1) {
+			perror("fuse: fork() failed");
+			close(fd);
+			_exit(EXIT_FAILURE);
 		}
 
 		if (pid == 0) {
--- a/lib/mount_bsd.c
+++ b/lib/mount_bsd.c
@@ -17,9 +17,7 @@
 #include <sys/param.h>
 #include "fuse_mount_compat.h"
 
-#include <sys/stat.h>
 #include <sys/wait.h>
-#include <sys/user.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
@@ -27,8 +25,6 @@
 #include <fcntl.h>
 #include <errno.h>
 #include <string.h>
-#include <paths.h>
-#include <limits.h>
 
 #define FUSERMOUNT_PROG		"mount_fusefs"
 #define FUSE_DEV_TRUNK		"/dev/fuse"
--- a/lib/mount_bsd.c
+++ b/lib/mount_bsd.c
@@ -12,6 +12,7 @@
 #include "fuse_i.h"
 #include "fuse_misc.h"
 #include "fuse_opt.h"
+#include "util.h"
 
 #include <sys/param.h>
 #include "fuse_mount_compat.h"
@@ -151,7 +152,7 @@ static int init_backgrounded(void)
 static int fuse_mount_core(const char *mountpoint, const char *opts)
 {
 	const char *mountprog = FUSERMOUNT_PROG;
-	int fd;
+	long fd;
 	char *fdnam, *dev;
 	pid_t pid, cpid;
 	int status;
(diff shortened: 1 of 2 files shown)


vassilit:
> [Not modified is mount_bsd.d as I cannot test it.](https://github.com/libfuse/libfuse/pull/902#issue-2184278505)

Good idea, I will modify and test it a bit later.

vassilit:
Following the information I got in the [answer](https://github.com/libfuse/libfuse/discussions/1100) on the discussion page, I've done some proper tests again waiting for FUSE_INIT to return. 
While it's true that the delta t between `fuse_main()` process exiting and FUSE_INIT is inferior without the double fork, the total delta t between `fuse_main()` invocation and FUSE_INIT is slightly better with a double fork. It seems the deadlock if any is not blocking, but it is not worth the candle to potentially break something, that's why I removed the [commit](https://github.com/libfuse/libfuse/commit/f1335d5290ba64f84850c04c0bf6b3af16d87359) for now but if you have any enlightening information about the utility of this double-fork it would be great to get to the bottom of this. I will do posix_spawn however.

For the record, the total delta t is about 20% better (average of 800µs) with a double-fork, but there are spikes of more than 250% (2000µs) once in ~50 invocation in a row.
Without the double-fork, no spikes but an average of 1000µs.

vassilit:
A small detail to look. Why a double-fork if `setsid()` isn't called ? Why not a simple fork without `waitpid()` ? The actual implementation wait for `exit(0)` to succeed, which seems pretty useless. The grandchild process calls `perror()`, no wonder it needs a terminal.
If not in the foreground, soon enough `fuse_daemonize()` will fork again anyway.
Also, if `exec(mount_fusefs)` fails for whatever reason, `free(fdnam)` will probably segfault.

bsbernd:
Would you mind to put the double fork patch into a different PR? All the other patches are simple, but for the fork I need to have a closer look and I also need to release 3.17 - except of the fork all of it could go to the 3.17 branch - easier if we merge it a few days later.

vassilit:
No problem. However, I removed the dangerous one and this last one is simple and tested and does not change the flow, so no need look into the synchronous/asynchronous behaviour of the FreeBSD kernel.
It correct a possible segfault (which is plausible, if /sbin is not in the PATH for whatever reason for example).

vassilit:
> Would you mind to put the double fork patch into a different PR? 

I've removed the last commit about the simplification of the double-fork.
It will go into a new PR, but it'll be easier for me to base it on the new master after you merge this one so I'll wait.

vassilit:
Rebased for two small corrections:

- Remove unused header `sys/sysctl.h` in Remove useless options
- Add a commit to clean other unused headers

bsbernd:
Could you add you "Signed-off-by:"?

bsbernd:
Could we also add a BSD test? That uses this here? https://github.com/marketplace/actions/freebsd-vm?

bsbernd:
I don't think anything is using that .cirrus.yml file

bsbernd:
Untested, suggested by the AI tool we use at work (Augment)
```
---
name: 'Build & Test (FreeBSD)'
on:
  push:
    branches:
      - master
      - '[0-9]+.[0-9]+'  # This will match branches like 3.17, 3.18, 4.0, etc.
  pull_request:
    branches:
      - master
      - '[0-9]+.[0-9]+'

permissions:
  contents: read

jobs:
  build:
    runs-on: ubuntu-latest
    name: FreeBSD Build
    steps:
      - name: Checkout
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2

      - uses: actions/setup-python@v5
        with:
          python-version: '3.12'

      - name: Build and Test on FreeBSD
        uses: vmactions/freebsd-vm@v1
        with:
          usesh: true
          prepare: |
            pkg update -f
            pkg install -y \
              clang \
              doxygen \
              gcc \
              valgrind \
              meson \
              ninja \
              python3 \
              pkgconf \
              libtool

          run: |
            pip install -r requirements.txt
            test/ci-build.sh

      - name: Upload logs on failure
        uses: actions/upload-artifact@v4
        if: failure()
        with:
          name: freebsd-logs
          path: |
            build/meson-logs/meson-log.txt
            build/meson-private/
          if-no-files-found: ignore```

**Comment by vassilit** [comment]
I could put something in place, but I think this should go to another PR.

I would have question, like where is your [`iwyu.imp`](https://github.com/libfuse/libfuse/blob/42045175964ca3f46dd8f49409ac0cc97e51289e/.github/workflows/iwyi-check.yml#L44C42-L44C50) ?

**Comment by vassilit** [comment]
It should be also determined what is the minimum version of FreeBSD we target.

**Comment by bsbernd** [comment]
The iwyu test is actually not running yet, it needs some work - will do that when I find some time.

**Comment by asomers** [comment]
This has caused a regression.  Now, every single time that a fusefs file system unmounts, the daemon prints:
```
unmounting /tmp/mnt failed: Invalid argument
```
That's because after somebody does `/sbin/umount <MOUNTPOINT>`, the kernel sends `FUSE_DESTROY` to the daemon, and the daemon tries to unmount the mountpoint again.  The double-unmount isn't new; only the printed warning is.  I'm going to open a new issue for this.

PR #1146, opened 2025-02-20
Replace fork by posix_spawn in mount_bsd.c

vassilit:
Analog to #902


vassilit:
checkpatch is plain wrong on alignment.

It complains when I do:
```
|tab   |if (status) {
|tab   ||tab   |fuse_log(FUSE_LOG_CRIT, "spawning " FUSERMOUNT_PROG " failed: %s (%d)\n",
|tab   ||tab   |         strerror(status), status);
```
But accept doing:
```
|tab   |if (status) {
|tab   ||tab   |fuse_log(FUSE_LOG_CRIT, "spawning " FUSERMOUNT_PROG " failed: %s (%d)\n",
|tab   ||tab   ||tab   | strerror(status), status);
```
In `/.dir-locals.el`, libfuse define tab width at 8. What if the function we call is less than 8 char ? How do we align ?
Also, the first variant allows any tab size, while the second will misalign.

bsbernd:
checkpatch.pl is a copy of `<linux>/scripts/checkpatch.pl` - unlikely that it is wrong.

bsbernd:
If next indentation is less than a tab, it needs to be indented by spaces only.

vassilit:
> checkpatch.pl is a copy of `<linux>/scripts/checkpatch.pl` - unlikely that it is wrong.

Yes, you are right. "Plain wrong" was a wrong choice of word. The kind of wording that they would use in <linux> in the past.

> If next indentation is less than a tab, it needs to be indented by spaces only.

Actually, it seems that you are talking about alignment, not indentation. I'm used to «indent with tabs and align with spaces». I now understand that we are enforcing Linux kernel coding style here (which is not bad, just  a bit different) of «indent with tabs and align with tabs+spaces if needed». I will use that here from now on, I was not aware ; I've changed the commit to comply.

bsbernd:
Yeah, the project was using linux style from the beginning - no good reason to change the style. Already not ideal that some files that got introduced a few years back have a different style (4 char spaces).

bsbernd:
I was not aware of `dir-locals.el`and not using emacs either - that needs to be fixed to linux style.

vassilit:
For reference, `HAVE_ENVIRON_DECL` is needed only when compiling for Debian GNU/kFreeBSD.
It doesn't interfere in other cases and allows to be future-proof if ever FreeBSD decides one day to declare `environ` in the libc.

bsbernd:
@vassilit I will look at the PR later today. Could you have a look at the emacs config file PR, when you have time?
https://github.com/libfuse/libfuse/pull/1148

asomers:
Sorry @vassilit I'm not sure I'll have the time to review this.  I'll try to get to it this weekend.  But if I don't, I might never.

asomers:
> > Sorry @vassilit I'm not sure I'll have the time to review this. I'll try to get to it this weekend. But if I don't, I might never.
> 
> I understand. Also, a full review is not necessary, just a quick comment about your views on synchronously calling `nmount()` directly vs spawning mount_fusefs in the background without waiting for it like it's done today.

I'm not up to speed on what this problem is.  But it's certainly possible to call `nmount` directly.  In fact, that's what the fuse3 Rust crate does.  `mount_fusefs` doesn't do anything that you can't do by yourselves.  It does some argument sanity checking, realpath(3), etc.

vassilit:
The double-fork was introduced in https://github.com/libfuse/libfuse/commit/2efe48fc775fbc66e63694e365078f22092251a8

vassilit:
I've got to the bottom of this. My PR was right from the start. My cautious questions in this thread about the potential dead-lock due to an old comments was due to my wrong assumption that _init_backgrounded_ was false by default. But when you look at the SYSCTL_INT in FreeBSD it is hard-wired to 1.

So before @asomers removal of the sysctl, libfuse did not double-fork, as it was conditional on `!init_backgrounded()` (poorly named IMHO). After the removal, the function that I removed in 124ceb7 called `sysctlbyname`, which returned -1 (error) and returned `ibg` which was uninitialised (undetermined), so libfuse sometimes double-forked, sometimes not. After my commit 124ceb7, the double-fork was always (wrongly) done. The double-fork is (and always was, but it was not done when the sysctl existed) completly useless as FreeBSD init was always done asynchronously.
This double-fork and potential dead-lock in the comments was the blocker of this PR, but now I don't have anymore a zombie problem and #1240 was merged more than a year ago and no one reported a dead-lock.

TLDR: the following comment from 2006 that I deleted in 124ceb7 was never applicable to FreeBSD, so removing the double-fork and do waitpid is the right thing to do.
```c
			/*
			 * If init is not backgrounded, we have to
			 * call the mount util backgrounded, to avoid
			 * deadlock.
			 */
```
(this comment is apparently inherited from macFUSE, which was never applicable to FreeBSD)

bsbernd:
@vassilit Good timing, I'm about to enable BSD tests: https://github.com/libfuse/libfuse/pull/1631 I.e. not only compilation but running the actual tests.

vassilit:
(rebasing to run FreeBSD CI)

bsbernd (approved):

asomers:
> > I'm not up to speed on what this problem is. But it's certainly possible to call `nmount` directly. In fact, that's what the fuse3 Rust crate does. `mount_fusefs` doesn't do anything that you can't do by yourselves. It does some argument sanity checking, realpath(3), etc.
> 
> Yes, calling nmount directly works and it is understood. My question was about an old comment in the `mount_bsd.c` code:
> 
> ```
>             /*
>              * If init is not backgrounded, we have to call the mount util
>              * backgrounded, to avoid deadlock.
>              */
> ```
> 
> As per your [commit](https://svnweb.freebsd.org/base?view=revision&revision=347544), init is never "backgrounded".
> 
> I've checked [fuse3 crate](https://github.com/Sherlock-Holo/fuse3/blob/17eae92592c4aab5a9dd30d5b09a92987c0d4823/src/path/session.rs#L49C1-L51C19) and it is mounting «semi-asynchronously».
> 
> The owner of the [fuse-rs](https://github.com/zargony/fuse-rs/issues/9#issuecomment-32167809) crate thinks that osxfuse `fuse_vfsop_mount` [...] then synchronously waits for the FUSE_INIT reply. This is about osxfuse but they seem very similar.
> 
> Looking at the current code of FreeBSD, it seems that init reply is waited asynchronously for a long time, so it's safe to call `nmount()` in mount_bsd.c and wait for its return. Actually, I can already tell it works. But I just want to be sure with you that it will always work, because I haven't understood the history behind this dead-lock comment .... I guess `fuse_internal_send_init` was blocking at some point of time in the past that would explain the comment, but I couldn't find it clearly. It's more a question to err on the side of caution. I [stress tested](https://gist.github.com/vassilit/e00d59135e3204aabcc4cca8c333c140) libfuse a bit with the direct synchronous implementation and it always worked.

Good sleuthing!  And as it turns out, last week I coincidentally implemented "foreground init" in FreeBSD's fuse module.  However, it causes some programs to break, so I'm not going to commit it.  If I understand correctly, the deadlock that can happen with "foreground init" comes from the fact that the nmount(2) syscall will block until the daemon responds to the FUSE_INIT message.  Obviously, responding before nmount(2) returns requires either a multithreaded daemon or a multiprocessing one.  But some fuse servers expect nmount(2) to return, _then_ they read the FUSE_INIT message and respond to it.

bsbernd:
@asomers Did you see how Linux' sync-FUSE_INIT works? By implementation it needs the new mount API (which still gives some headache and sometimes hickup in the tests - new issue was found last week when I added support for VM runs with recent kernels). Independent how BSD is going to implement it, there are conditions for fuse_daemonize_early* usage - I *guess* the daemons your commit breaks will need it - there needs to be negotiation between kernel and libfuse if sync-INIT can be used.

PR #1149, opened 2025-02-20
FreeBSD: test runner

vassilit:
Draft version of a yet simple runner that just build libfuse under FreeBSD and show the compile log.


vassilit:
Are we on the same page to only support successful compilation, and running tests only on officially supported FreeBSD versions ?
(i.e. 14 and 13 at the moment).
And a best-effort to support 15-CURRENT internally.

What's the policy on Linux ? I can't find where it documented.
For EL, I remember seeing that RHEL7 was still supported by libfuse regardless of its EOL status.
i.e. do we have a min-GLIBC, min-MUSL version ?
As for the kernel, I understand that modern libfuse3 should be compatible with any kernel >2.6.14 (when FUSE became officially supported) ?

bsbernd:
Right now tests for Linux only happen on latest Ubuntu. You have noticed that I'm adding more tests to the project...
If we can get *any* BSD test, it would already help.

bsbernd:
@Nikratio do you know by any chance? Do you have a login for https://cirrus-ci.com?

bsbernd:
@vassilit could you give more details what is the issue with the BSD VM?

Nikratio:
@bsbernd I can log into cirrus-ci.com using my Github account, and then it shows me the most recent builds for libfuse.

bsbernd:
@vassilit This seems to do something for FreeBSD - can't we just use it and get reports?

bsbernd:
@Nikratio Sorry for my late reply, do you know if we can access these cirrus reports somehow in github?

.github/workflows/bsd.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  build_bsd:
+    runs-on: ubuntu-latest
+    name: Build under FreeBSD
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Build
+      uses: vmactions/freebsd-vm@v1
```
bsbernd, line 23: We should use git commit ID for the action.

.github/workflows/bsd.yaml
```diff
@@ -0,0 +1,33 @@
...
+name: 'Build (FreeBSD)'
+on:
+  push:
+    branches:
+      - master
+      - 'fuse-[0-9]+.[0-9]+*'
+  pull_request:
+    branches:
+      - master
+      - 'fuse-[0-9]+.[0-9]+*'
```
bsbernd, line 13: Permission settings are missing.

vassilit:
Hello, I will come back to this in a few days.

Having Cirrus-CI results on Github alongside Github actions seems [possible](https://github.com/freebsd/freebsd-src/pull/1634/checks?check_run_id=39845044646).

Nikratio:
> @Nikratio Sorry for my late reply, do you know if we can access these cirrus reports somehow in github?

No idea, sorry.

PR #1152, opened 2025-02-24
USDT request tracepoints

joannekoong:
Add user statically-defined tracepoints for observability into request lifecycle (request receive, request processing, request reply).

Output paste from running usdt.bt and "cat file.c":
```
libfuse:request_receive hit, err=0
libfuse:request_process hit, opcode=1, unique=88
libfuse:request_reply hit, unique=88, len=144, err=0, reply_err=0
libfuse:request_receive hit, err=0
libfuse:request_process hit, opcode=14, unique=90
libfuse:request_reply hit, unique=90, len=32, err=0, reply_err=0
libfuse:request_receive hit, err=0
libfuse:request_process hit, opcode=15, unique=92
libfuse:request_reply hit, unique=92, len=1257, err=0, reply_err=0
libfuse:request_receive hit, err=0
libfuse:request_process hit, opcode=25, unique=94
libfuse:request_reply hit, unique=94, len=16, err=0, reply_err=0
libfuse:request_receive hit, err=0
libfuse:request_process hit, opcode=18, unique=96
libfuse:request_reply hit, unique=96, len=16, err=0, reply_err=0
```

### DIFF
--- a/example/usdt.bt
+++ b/example/usdt.bt
@@ -1,0 +1,19 @@
+#!/usr/bin/env bpftrace
+
+// To run, do `sudo bpftrace usdt.bt`
+
+usdt:../build/lib/libfuse3.so:libfuse:request_receive
+{
+    printf("libfuse:request_receive hit, err=%d\n", arg0);
+}
+
+usdt:../build/lib/libfuse3.so:libfuse:request_process
+{
+    printf("libfuse:request_process hit, opcode=%u, unique=%u\n", arg0, arg1);
+}
+
+usdt:../build/lib/libfuse3.so:libfuse:request_reply
+{
+    printf("libfuse:request_reply hit, unique=%lu, len=%u, err=%u, reply_err=%d\n",
+	    arg0, arg1, arg2, arg3);
+}
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -18,6 +18,7 @@
 #include "fuse_opt.h"
 #include "fuse_misc.h"
 #include "mount_util.h"
+#include "usdt.h"
 #include "util.h"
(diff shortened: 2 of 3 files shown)


bsbernd:
@joannekoong I wonder if adding usdt.h should be done with a git submodule. I see that upstream did changes to that file 2 months ago - we would need to regularly pull in usdt changes?

joannekoong:
Hi @bsbernd. The libbpf usdt.h encourages copying the single-file source file directly. The [documentation](https://github.com/libbpf/usdt) states:

> It is intentional that usdt.h has no dependencies and is single-file. It will remain single-file library going forward. This is meant to cause minimal hurdles in adoption inside applications. It is encouraged to just copy the file into your project and bypass any sort of dependency management or packaging issues.

bsbernd:
@joannekoong Ok, issues already start with checkpatch.pl - it is different project, follows different style and then fails. At best it should be disabled for that file, but I don't like the approach too much. And every time we do an update for that file, we need to pull in remote changes that are not under our control.

joannekoong:
> And every time we do an update for that file, we need to pull in remote changes that are not under our control.

The file is not expected to be updated. I synced with Andrii, who is the author of the file and one of the bpf maintainers, and he said the intention of the file is to be easily copy and pasted to wherever it's used instead of adding it a dependency.

bsbernd:
@joannekoong Ok, fine, let's take it. I will grumble a bit when the file gets updated ;) Would you mind to add a meson option for this? Besides that one might want to run without, it also might cause issues on BSD?

joannekoong:
Sorry for the late reply, I'll add a meson option for this if you don't get to it first.

PR #1153, opened 2025-02-25
Update fuse_lowlevel_notify_store() documentation

luis-henrix:
Added a warning on a possible server deadlock if the server tries to push data using fuse_lowlevel_notify_store() when replying to another FUSE request (for example, from FUSE_READ).

(Feel free to rephrase the text I'm adding to the documentation.)

### DIFF
--- a/doc/README.notifications
+++ b/doc/README.notifications
@@ -1,0 +1,37 @@
+During the life-cycle of a user-space filesystem the usual flow is:
+
+  1. User-space application does a filesystem-related syscall
+  2. Kernel VFS calls into the FUSE kernel driver
+  3. FUSE kernel redirects request to the user-space filesystem
+  4. User-space server replies to request
+  5. FUSE returns reply to VFS
+  6. User-space application gets reply from the kernel
+
+However, there are occasions where the filesystem needs to send notifications to
+the kernel that are not in reply to any particular request.  If, for example,
+when a READ request of 4096 bytes results in the filesystem having more data
+available for the specific inode, it may be useful to provide this extra data to
+the kernel cache so that future read operations will be faster.
+
+FUSE provides mechanisms for a user-space server to send the kernel certain
+types of asynchronous notifications.  Currently, these are the available
+notifications:
+
+|-------------+----------------------------------|
+| Operation   | libfuse function                 |
+|-------------+----------------------------------|
+| POLL        | fuse_lowlevel_notify_poll        |
+| INVAL_INODE | fuse_lowlevel_notify_inval_inode |
+| ENTRY       | fuse_lowlevel_notify_inval_entry |
+| STORE       | fuse_lowlevel_notify_store       |
+| RETRIEVE    | fuse_lowlevel_notify_retrieve    |
+| DELETE      | fuse_lowlevel_notify_delete      |
+| RESEND      | -                                |
+|-------------+----------------------------------|
+
+One important restriction is that these asynchronous operations SHALL NOT be
+performed while executing other FUSE requests.  Doing so will likely result in
+deadlocking the user-space filesystem server.  In the example above, if the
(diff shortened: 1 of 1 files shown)


include/fuse_lowlevel.h
```diff
@@ -1845,6 +1845,10 @@ int fuse_lowlevel_notify_delete(struct fuse_session *se,
  * If the stored data overflows the current file size, then the size
  * is extended, similarly to a write(2) on the filesystem.
  *
+ * To avoid a deadlock this function must not be called while executing
+ * a related filesystem operation (e.g. while replying to a FUSE_READ
+ * request).
+ *
```
bsbernd, line 1851: I think all notification should not be called while executing fuse kernel requests. I wonder if we should better create a file like docs/README-notifications.1st?

luis-henrix: OK, that probably makes more sense.  Let me try to write something to share through this MR.  Thanks @bsbernd

luis-henrix:
@bsbernd I've just pushed a new commit that adds an extra README file into the doc directory.  Let me know what you think about it.

doc/README.notifications
```diff
@@ -0,0 +1,37 @@
...
+| DELETE      | fuse_lowlevel_notify_delete      |
+| RESEND      | -                                |
+|-------------+----------------------------------|
+
+One important restriction is that these asynchronous operations SHALL NOT be
+performed while executing other FUSE requests.  Doing so will likely result in
+deadlocking the user-space filesystem server.  In the example above, if the
+server is replying to a READ request and has extra data to add to the kernel
+cache, it needs to reply to the READ request first, and, e.g., signal a
+different thread to do the STORE.
```
bsbernd, line 37: How about "it either needs to reply to the READ request first or signal a different thread to do the STORE".

doc/README.notifications
```diff
@@ -0,0 +1,37 @@
...
+  3. FUSE kernel redirects request to the user-space filesystem
+  4. User-space server replies to request
+  5. FUSE returns reply to VFS
+  6. User-space application gets reply from the kernel
+
+However, there are occasions where the filesystem needs to send notifications to
+the kernel that are not in reply to any particular request.  If, for example,
+when a READ request of 4096 bytes results in the filesystem having more data
+available for the specific inode, it may be useful to provide this extra data to
+the kernel cache so that future read operations will be faster.
```
bsbernd, line 14: Sounds a bit awkward for me, but I'm not a native speaker. The AI system at work suggests

Augment
When a filesystem has more data available than requested for an inode, providing this extra data to the kernel cache can optimize subsequent read operations.

bsbernd:
Thank you very much @luis-henrix Sorry for late reviews, rather swamped at the moment.

doc/README.notifications
luis-henrix: I'm not sure I understand your comment.  The server **needs** to reply to the READ request, right?  It may or may not push the extra data using the STORE, but that has to be done as a separate operation.

I'm giving as an example signaling a different thread, but to be honest I don't fully understand the threading model(s) provided by libfuse so I may be taking a very simplistic approach.

bsbernd: I'm now actually confused as well. Isn't my phrase exactly what you wrote before, just a bit shorter?  The important part is that fuse-notify operations are blocking - if the same thread handles fuse operations it would be dead-lock.

luis-henrix:
> I'm now actually confused as well. Isn't my phrase exactly what you wrote
> before, just a bit shorter?  The important part is that fuse-notify operations
> are blocking - if the same thread handles fuse operations it would be dead-lock.

Sorry, I guess I misread your suggestion.  Sorry for messing up.  But I
see you've merged this doc already.  Thanks a lot, Bernd.

Cheers,
-- 
Luís

PR #1156, opened 2025-03-07
fuse_session_receive_buf: Fix the pipe buf size

bsbernd:
This fixes dynamic buffer allocation in commit
0e0f43b79b9b ("Reallocate fuse_session buffer transparently for extended max writes")

I noticed that when I increased the default fuse buf size as possible in recent kernels.

### DIFF
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -72,14 +72,14 @@ struct fuse_session {
 	int broken_splice_nonblock;
 	uint64_t notify_ctr;
 	struct fuse_notify_req notify_list;
-	size_t bufsize;
+	_Atomic size_t bufsize;
 	int error;
 
 	/* This is useful if any kind of ABI incompatibility is found at
 	 * a later version, to 'fix' it at run time.
 	 */
 	struct libfuse_version version;
-	bool buf_reallocable;
+	_Atomic bool buf_reallocable;
 };
 
 struct fuse_chan {
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -79,7 +79,9 @@ struct fuse_session {
 	 * a later version, to 'fix' it at run time.
 	 */
 	struct libfuse_version version;
-	_Atomic bool buf_reallocable;
+
+	/* true if reading requests from /dev/fuse are handled internally */
+	bool buf_reallocable;
 };
 
 struct fuse_chan {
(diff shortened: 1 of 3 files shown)


bsbernd:
@joannekoong could you have a look?

lib/fuse_i.h
```diff
@@ -72,14 +72,14 @@ struct fuse_session {
 	int broken_splice_nonblock;
 	uint64_t notify_ctr;
 	struct fuse_notify_req notify_list;
-	size_t bufsize;
+	_Atomic size_t bufsize;
```
joannekoong, line 75: Does bufsize need to be _Atomic? AFAICS, it's only set when a session is created (fuse_session_new_versioned()) and when handling the init in do_init() which gets called before multi-threading becomes relevant.

lib/fuse_i.h
```diff
@@ -72,14 +72,14 @@ struct fuse_session {
...
 	int error;
 
 	/* This is useful if any kind of ABI incompatibility is found at
 	 * a later version, to 'fix' it at run time.
 	 */
 	struct libfuse_version version;
-	bool buf_reallocable;
+	_Atomic bool buf_reallocable;
```
joannekoong, line 82: do we need to also set buf_reallocable atomically in _fuse_session_receive_buf()?

joannekoong (commented):
Thanks for adding this in @bsbernd!

lib/fuse_i.h
bsbernd: I think we need _Atomic pr READ_ONCE/WRITE_ONCE, because the compiler *could* optimize out the 2nd read in _fuse_session_receive_buf`. Initially I had added READ_ONCE/WRITE_ONCE, but lines became much longer and there were several instances of that in that function. Using _Atomic is so much easier and I would guess that is similar micro perf impact.

lib/fuse_i.h
bsbernd: Ah yes, thanks, I wanted to do that, but somehow had forgotten it.

lib/fuse_i.h
joannekoong: Sounds good, I'm not familiar with how the compiler optimizes out reads, so I'll defer to you on this

lib/fuse_i.h
bsbernd: Actually I don't think we needed to do anything about further atomic operations anymore - this was just to avoid READ_ONCE/WRITE_ONCE.

Though, I added another patch, which simplifies/reduces buf_reallocable. It doesn't even need to be an atomic anymore. If you have time, could you check that 2nd patch @joannekoong ?

Issue #1163, opened 2025-03-13
3.17-rc1: Change capabilties from `#define` to `enum` - deal breaker?


bsbernd:
### Discussed in https://github.com/libfuse/libfuse/discussions/1162

<div type='discussions-op-text'>

<sup>Originally posted by **HereThereBeDragons** March 13, 2025</sup>
Hello @bsbernd 

I tried using 3.17-rc1 with cvmfs.
However, I stumbled upon a breaking change for us:
in `fuse_common.h` the capabilities are changed from `#define`s to an `enum`.

This is not very good for us as in our code we support multiple different (and old) OS and with this libfuse versions.
We use the preprocessors `#define`s to guard old OS/libfuse from unkown method signatures introduced by newer versions.

Having the `#define`s for capabilities also allows to check for backported capabilities (e.g. RHEL9 still uses 3.10 but has backported quite a few features) and not be limited just by the version check.

Now my question is:
- Do you have a better idea how to support multiple libfuse versions in the same code not using `#define`s as a guard?
- Is there any real advantage of having switched to the `enum`?
- Do you plan to reintroduce the `#define`s?

Thanks
Laura

</div>

PR #1164, opened 2025-03-13
fuse_common.h: Convert back FUSE_CAP_ from enum to defines

bsbernd:
Some applications use that for detection of features between distributions/libfuse version.

Fixes: https://github.com/libfuse/libfuse/issues/1163


HereThereBeDragons:
yes thank you, that works smoothly!
just tested it with your branch.
```
FUSE library version: 3.18.0-rc0
CernVM-FS: mounted cvmfs on /mnt/test
```

Issue #1171, opened 2025-03-20
fuse: both 'want' and 'want_ext' are set


bsbernd:
> Shouldn't this still set `want = 0`?
> 
> I'm using GVFS's FUSE daemon and I suspect this change is the one that resulted in:
> 
> ```
> 
> gvfsd[62597]: fuse: both 'want' and 'want_ext' are set
> 
> ```
> 
> Which means the GVFS FUSE daemon fails to initialize.
> 
> I've proposed a GVFS [change](https://gitlab.gnome.org/GNOME/gvfs/-/merge_requests/248/diffs?commit_id=77581bc426025bd04bd7ccb3b61e9c20114bc44b) that fixes it though. 

 _Originally posted by @tambry in [24f5b12](https://github.com/libfuse/libfuse/commit/24f5b129c4e1b03ebbd05ac0c7673f306facea1a#r154022913)_

bsbernd:
For myself, gvfsfusedaemon.c uses high level API.

tambry:
> Look into this function please
> I.e. it is supposed to only complain if conn->want and conn->want_ext are set.

Correct. This however doesn't seem to be the case.  
However just by reviewing code I can't find a reason for as to why this might be.

If `se->conn.want = 0;` was necessary because `se->conn` isn't zero-initialized on allocation then `se->conn.want` might now be garbage.  
gvfsd-fuse for example simply ORs the relevant bitfields and doesn't zero-initialize the field so `want` might be nonsensical.  
While wrong, it doesn't look like this should affect this.

> @tambry could you help me to reproduce? Is there a way that I compile gvfsd or run it manually outside of gnome?

I think you might be able to build as standalone and then run the `gvfs-daemon.service`?  
But honestly I'm not sure. I'm not a developer of gvfs but rather an user for whom it broke with Debian's upgrade of libfuse 3.14→3.17 so I investigated. I was already somewhat familiar since I had two small gvfsd patches pending for fixing MTP for my phone.

> For reproducing (also with another file system) - did you run into this with binary compiled against previous libfuse or with recompilation. If the issue exists it should not matter, but that this should not have come up in the first place.

Upgrading from libfuse 3.14→3.17 results in the same gvfsd-fuse binary failing to start with the `want` and `want_ext` error.  
So libfuse broke gvfsd-fuse with no recompiles or changes necessary in gvfsd-fuse.

[...]

PR #1172, opened 2025-03-22
fuse: Fix want flag conversion

bsbernd:
32-bit conn->want flags been left to be ABI compatible to 3.10, even though the so version was changed.
The more recent way is to use fuse_set_feature_flag(), which will use conn->want_ext.

Given that we now have two flags (want and want_ext), we need to convert and that brought several issues
- If the application sets conn->want, that needs to be set into the lower 32 bit of  conn->want_ext. As the application might actually unset values, it really has to be a copy and not just 'or' - fixed now.
- convert_to_conn_want_ext() actually needs to check for _modified_ conn->want and conn->want_ext
- convert_to_conn_want_ext() must consider being called from high and lowlevel interfact, with different want_ext_default and want_default values. It is only a failure, if the application changed both, conn->want and conn->want_ext. This function was failing in issue #1171, because high level fuse_fs_init() was changing values and then lowlevel do_init() was incorrectly failing on that.

This also adds a new test (test_want_conversion) and sets values into example/{hello.c,hello_ll.c}

Also some more internal users of conn->want are converted to fuse_{set,unset}_feature_flag().

closes: https://github.com/libfuse/libfuse/issues/1171

### DIFF
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -8,8 +8,11 @@
 
 #include "fuse.h"
 #include "fuse_lowlevel.h"
+#include "util.h"
 
+#include <stdint.h>
 #include <stdbool.h>
+#include <errno.h>
 
 #define MIN(a, b) \
 ({									\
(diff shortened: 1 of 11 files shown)


lib/fuse_i.h
```diff
@@ -224,3 +227,32 @@ int fuse_loop_cfg_verify(struct fuse_loop_config *config);
...
+	 */
+	if (conn->want != want_default &&
+	    fuse_lower_32_bits(conn->want_ext) != conn->want) {
+		if (conn->want_ext != want_ext_default) {
+			fuse_log(FUSE_LOG_ERR,
+				 "fuse: both 'want' and 'want_ext' are set\n");
+			return -EINVAL;
+		}
+		conn->want_ext &= ~0xffffffffULL; /* zero lower 32 bits */
+		conn->want_ext |= conn->want; /* set lower 32 bits from want */
```
tambry, line 254: Would this maybe read clearer?
```
conn->want_ext = (uint64_t)fuse_lower_32_bits(conn->want_ext) | conn->want;
```

gcsideal:
Is this targeted to 3.17.x? If not, would it be safe to backport these changes to that branch?

bsbernd:
@gcsideal Yes, definitely. Just updated code a bit to address the review from @tambry (thanks for the review!) and try to make it better readable.
I let tests pass and will then cherry-pick into 3.17.x

lib/fuse_i.h
tambry: Ah, right, oops. The now-changed variant is definitely better. :slightly_smiling_face:

PR #1181, opened 2025-03-25
example/hello_ll.c: Improve the compilation comments

izxl007:
When building and installing FUSE3 using Meson on CentOS8, the fuse3.pc
file is installed in the /usr/local/lib64/pkgconfig directory. However,
pkg-config does not search for fuse3.pc in this directory, leading to GCC
compilation failures.

This patch improves the compilation comments, helping users successfully
execute the GCC command.

### DIFF
--- a/example/hello_ll.c
+++ b/example/hello_ll.c
@@ -14,6 +14,10 @@
  *
  *     gcc -Wall hello_ll.c `pkg-config fuse3 --cflags --libs` -o hello_ll
  *
+ * Note: If the pkg-config command fails due to the absence of the fuse3.pc
+ *     file, you should configure the path to the fuse3.pc file in the
+ *     PKG_CONFIG_PATH variable.
+ *
  * ## Source code ##
  * \include hello_ll.c
  */


example/hello_ll.c
```diff
@@ -14,6 +14,11 @@
  *
  *     gcc -Wall hello_ll.c `pkg-config fuse3 --cflags --libs` -o hello_ll
  *
+ * Note: On CentOS, if the libfuse library is manually built and installed
+ *     using Meson, use the following compilation command:
+ *     gcc -Wall hello_ll.c `PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig \
+ *     pkg-config fuse3 --cflags --libs` -o hello_ll
```
bsbernd, line 20: Dunno, so you basically suggest that we add a comment for every distribution? The given path is also not generic, as it is a meson option.

example/hello_ll.c
```diff
@@ -14,6 +14,11 @@
  *
  *     gcc -Wall hello_ll.c `pkg-config fuse3 --cflags --libs` -o hello_ll
  *
+ * Note: On CentOS, if the libfuse library is manually built and installed
```
bsbernd, line 17: Are you sure this is centos specific? If so which version? Maybe

pkg-config might not search for its config in all installation paths, this can be handled by

* gcc -Wall hello_ll.c `PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig \
* pkg-config fuse3 --cflags --libs` -o hello_ll
* The given path might need to be adjusted to the fuse installation directory.

bsbernd (changes requested):
The commit subject is rather misleading - this is not a fix, but just code comment. And CentOS is both, too specific and too generic. Too specific in the sense that the same might happen for other distributions, to generic in the sense that there are several centos versions out there. I don't think it should mention CentOS at all.

example/hello_ll.c
izxl007: > Are you sure this is centos specific? If so which version? Maybe
> 
> pkg-config might not search for its config in all installation paths, this can be handled by
> 
> * gcc -Wall hello_ll.c `PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig \
> * pkg-config fuse3 --cflags --libs` -o hello_ll
> * The given path might need to be adjusted to the fuse installation directory.

I encountered this problem on CentOS 8. I’ve also verified that the issue doesn’t exist on Ubuntu 22.04. I don’t have access to other operating system environments, so I haven’t been able to test further.

On CentOS 8, the installation path for fuse3.pc is /usr/local/lib64/pkgconfig.

example/hello_ll.c
izxl007: > Dunno, so you basically suggest that we add a comment for every distribution? The given path is also not generic, as it is a meson option.

If the pkg-config command fails due to fuse3.pc, prompt the user to configure the PKG_CONFIG_PATH variable. 
Is this a better way?

bsbernd: I suggest to update the subject to something like "example/hello_ll.c: Improve the compilation comment"

bsbernd: And then remove the distribution specific part from the comment.

bsbernd:
@izxl007 Thanks, looks better now. Could you please merge/rebase the commits into one single commit?

izxl007:
> @izxl007 Thanks, looks better now. Could you please merge/rebase the commits into one single commit?

Done

bsbernd:
@izxl007 Thank you, could you please also update your commit message to make checkpatch happy?

bsbernd (approved):

PR #1183, opened 2025-03-26
example: Add a compilation instruction in README.compile

izxl007:
When building FUSE3 with Meson on CentOS 8, the fuse3.pc file gets
installed in /usr/local/lib64/pkgconfig. Since pkg-config doesn't search
this path by default, GCC compilation fails due to missing FUSE3 flags.

This patch adds an instruction for setting PKG_CONFIG_PATH variable in 
README.compile to fix GCC compilation issues.

### DIFF
--- a/example/README.compile
+++ b/example/README.compile
@@ -1,0 +1,4 @@
+Note:
+  * If the pkg-config command fails due to the absence of the fuse3.pc file,
+  you should configure the path to the fuse3.pc file in the PKG_CONFIG_PATH
+  variable.


bsbernd:
Maybe there should be an `example/README.compile` and then compilation instructions should be given ones? Somehow it does not make sense to make to duplicate the same comment in all files.

bsbernd:
Thank you @izxl007 !

izxl007:
> Thank you @izxl007 !

You're welcome. I'm happy to receive comments that improve the code.

PR #1185, opened 2025-03-31
Fix build with kernel < 5.9

giuliobenetti:
linux/close_range.h is only available since kernel 5.9 and https://github.com/torvalds/linux/commit/60997c3d45d9a67daf01c56d805ae4fec37e0bd8 resulting in the following build failure:

../util/fusermount.c:40:10: fatal error: linux/close_range.h: No such file or directory

So let's check for header presence and emit HAVE_LINUX_CLOSE_RANGE_H accordingly and check for it when including <linux/close_range.h>.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -72,7 +72,7 @@ private_cfg.set_quoted('PACKAGE_VERSION', meson.project_version())
 # Test for presence of some functions
 test_funcs = [ 'fork', 'fstatat', 'openat', 'readlinkat', 'pipe2',
                'splice', 'vmsplice', 'posix_fallocate', 'fdatasync',
-               'utimensat', 'copy_file_range', 'fallocate', 'close_range' ]
+               'utimensat', 'copy_file_range', 'fallocate' ]
 foreach func : test_funcs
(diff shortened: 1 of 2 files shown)


bsbernd:
Hi @giuliobenetti , thank you! I just wonder if we then shouldn't replace all of HAVE_CLOSE_RANGE by HAVE_LINUX_CLOSE_RANGE_H. We currently do
```
diff --git a/meson.build b/meson.build
index 0487639c..cbcd70d0 100644
--- a/meson.build
+++ b/meson.build
@@ -72,7 +72,7 @@ private_cfg.set_quoted('PACKAGE_VERSION', meson.project_version())
 # Test for presence of some functions
 test_funcs = [ 'fork', 'fstatat', 'openat', 'readlinkat', 'pipe2',
                'splice', 'vmsplice', 'posix_fallocate', 'fdatasync',
-               'utimensat', 'copy_file_range', 'fallocate' ]
+               'utimensat', 'copy_file_range', 'fallocate', 'close_range' ]
```
Maybe that needs to be pulled out here and be replaced by a more extensive test?

PR #1189, opened 2025-04-04
Fix static_assert build failure with C++ version < 11

giuliobenetti:
At the moment build fails due to lack of static_assert: https://gitlab.com/jolivain/buildroot/-/jobs/9606292537 this means that the check per date is not enough, so let's use meson to check if static_assert() is present or not and simplify fuse_static_assert() definition by only checking HAVE_STATIC_ASSERT.


bsbernd:
Thank you!

PR #1195, opened 2025-04-10
mount: Add FUSE_KERN_DEVICE env variable to specify fuse kernel device

menglu0105:
For kernel development it might be necessary to load a module with renamed symbols and renamed /dev/<devicenode>. Reason is that for example ubuntu kernels have fuse compiled in and it is not possible to replace it at run time. And fuse might also be used for other file systems - a different device node is then needed.

### DIFF
--- a/lib/mount.c
+++ b/lib/mount.c
@@ -49,6 +49,7 @@
 #define FUSERMOUNT_PROG		"fusermount3"
 #define FUSE_COMMFD_ENV		"_FUSE_COMMFD"
 #define FUSE_COMMFD2_ENV	"_FUSE_COMMFD2"
+#define FUSE_KERN_DEVICE_ENV	"FUSE_KERN_DEVICE"
 
 #ifndef MS_DIRSYNC
 #define MS_DIRSYNC 128
@@ -506,7 +507,7 @@ static int fuse_mount_sys(const char *mnt, struct mount_opts *mo,
 			  const char *mnt_opts)
 {
 	char tmp[128];
-	const char *devname = "/dev/fuse";
+	const char *devname = getenv(FUSE_KERN_DEVICE_ENV) ?: "/dev/fuse";
 	char *source = NULL;
 	char *type = NULL;
 	struct stat stbuf;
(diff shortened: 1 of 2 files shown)


lib/mount.c
```diff
@@ -528,7 +529,11 @@ static int fuse_mount_sys(const char *mnt, struct mount_opts *mo,
 	fd = open(devname, O_RDWR | O_CLOEXEC);
 	if (fd == -1) {
 		if (errno == ENODEV || errno == ENOENT)
-			fuse_log(FUSE_LOG_ERR, "fuse: device not found, try 'modprobe fuse' first\n");
+		{
+			fuse_log(FUSE_LOG_ERR, "fuse: device %s not found\n", devname);
+			if (strcmp(devname, "/dev/fuse") == 0)
+				fuse_log(FUSE_LOG_ERR, "fuse: try 'modprobe fuse' first\n", devname);
+		}
```
bsbernd, line 536: Maybe let's avoid the check for the exact dev name and just

```
fuse_log(FUSE_LOG_ERR, "fuse: device %s not found. Kernel module not loaded?\n", devname);
``

File: util/fusermount.c [hunk 2]
```diff
@@ -1189,7 +1192,7 @@ static int try_open_fuse_device(char **devp)
 	int fd;
 
 	drop_privs();
-	fd = try_open(FUSE_DEV, devp, 0);
+	fd = try_open(devname, devp, 0);
```
bsbernd, line 1195: I think better to have
```
const char *devname = getenv(FUSE_KERN_DEVICE_ENV) ?: FUSE_DEV;
```
here than as global static variable.

bsbernd: However, looking more closely, the code is a bit confusing

```
mount_fuse
    ...
    char *dev;
    ...
    fd = open_fuse_device(&dev);
        try_open_fuse_device
            try_open --> strdup to retrieve the pointer
```

Unless I'm missing something, we can just use

const char *dev = getenv(FUSE_KERN_DEVICE_ENV) ?: FUSE_DEV;

in `mount_fuse` and avoid the strdup() altogether?

bsbernd (changes requested):

util/fusermount.c
menglu0105: Hi Bernd, comment addressed. Since open_fuse_device() is the only caller of try_open_fuse_device() which is also the only caller of try_open() - and no one is even using the "silent" flag of try_open() - I squashed all the logic into open_fuse_device() itself. Please help to review again, thank you!

bsbernd (commented):
LGTM, could you please also update the commit message and add a short comment about the code consolidation and avoided strdup?

bsbernd (approved):

PR #1177, opened 2025-03-24
Add fuse-over-io-uring support

bsbernd:

PR #1200, opened 2025-04-21
Famfs changes to libfuse

jagalactic:
These libfuse changes go with the famfs patch that I expect to send out later this evening. I will edit this PR to point to the famfs/fuse kernel patch set on lore, once it's been posted.

Two new low-level fuse messages/responses are added: GET_FMAP and GET_DAXDEV. Both are documented to a reasonable extent in the kernel patch set.

In addition, famfs needs to set some additional kernel mount options, and this enables that.

The famfs fuse kernel patchset and documentation can be viewed here: https://lore.kernel.org/linux-fsdevel/20250421013346.32530-1-john@groves.net/T/#t

Cheers,
John

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2105,11 +2105,7 @@ void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
 		if (inargflags & FUSE_FLOCK_LOCKS)
 			se->conn.capable_ext |= FUSE_CAP_FLOCK_LOCKS;
 		if (inargflags & FUSE_AUTO_INVAL_DATA)
-			se->conn.capable_ext |= FUSE_CAP_AUTO_INVAL_DATA;
-		if (inargflags & FUSE_DO_READDIRPLUS)
-			se->conn.capable_ext |= FUSE_CAP_READDIRPLUS;
-		if (inargflags & FUSE_READDIRPLUS_AUTO)
-			se->conn.capable_ext |= FUSE_CAP_READDIRPLUS_AUTO;
+			se->conn.capable |= FUSE_CAP_AUTO_INVAL_DATA;
 		if (inargflags & FUSE_ASYNC_DIO)
 			se->conn.capable_ext |= FUSE_CAP_ASYNC_DIO;
 		if (inargflags & FUSE_WRITEBACK_CACHE)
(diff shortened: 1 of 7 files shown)


lib/fuse_versionscript
```diff
@@ -200,6 +200,7 @@ FUSE_3.17 {
 		fuse_set_fail_signal_handlers;
 		fuse_log_enable_syslog;
 		fuse_log_close_syslog;
+		fuse_add_kernel_mount_opt;
```
bsbernd, line 203: That needs to go into a new 3.18 section. See here
https://github.com/libfuse/libfuse/pull/1177/commits/ab33d7f8dcad1c20577f5461d790e3f9b497af14

('m still reviewing and cleaning it up myself)

lib/fuse_lowlevel.c
```diff
@@ -2379,6 +2386,30 @@ static void do_destroy(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
...
+}
+
+static void
+do_get_daxdev(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
+{
+	struct fuse_session *se = req->se;
+	(void)inarg;
+
+	if (se->op.get_daxdev)
+		se->op.get_daxdev(req, nodeid); /* Use nodeid as daxdev_index */
```
bsbernd, line 2408: Just quickly scanning through it, no reply?

jagalactic: If I understand the question... Replies for get_fmap() and get_daxdev() are handled in the fuse server, if it handles those messages.  https://github.com/cxl-micron-reskit/famfs/blob/c6bdc79929b9988dd2ee84d54d74c8af622ac50c/src/famfs_fused.c#L681

If this breaks from a norm, let me know and I'll try to be more normy ;)

Another question: although the reply payload for get_daxdev is fixed-size, the reply payload for get_fmap should be variable - but is currently fixed at a max. Can you point me to example code for a variable-sized reply payload? I kinda just fumbled around until I got this part working, and then focused on several much bigger challenges...

Thanks!

bsbernd: Just confused because reply types are not documented, see the other comment I had just added.

bsbernd: Regarding dynamic reply, maybe as `fuse_reply_buf`, `fuse_reply_data or `fuse_reply_iov?

jagalactic:
Closing because I pushed an updated PR for famfs (also not ready to merge, but FYI)

PR #1202, opened 2025-04-23
Check if pthread_setname_np() exists before use it

giuliobenetti:
Since pthread_setname_np() is the only pthread function that requires NPTL and it basically only set thread name, let's check if pthread_setname_np() does exist, otherwise let's not call pthread_setname_np() to shrink dependencies.


bsbernd:
Thanks @giuliobenetti

PR #1207, opened 2025-04-24
Fix meson function tests

bsbernd:
Several meson tests were incorrectly failing

Checking for function "static_assert" : NO (cached)
Checking for function "pthread_setname_np" : NO (cached)
Check usable header "#include <linux/close_range.h>" : NO (cached)

These functions get now tested with compilation tests and get found on my system.

Checking if "static_assert check" compiles: YES
Checking if "pthread_setname_np check" compiles: YES
Checking if "close_range check" compiles: YES

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -59,6 +59,8 @@ include_default = '''
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <assert.h>     /* For static_assert */
+#include <pthread.h>    /* For pthread_setname_np */
 '''
 args_default = [ '-D_GNU_SOURCE' ]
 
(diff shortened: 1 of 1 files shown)


bsbernd:
@giuliobenetti We had a bit of an issue with meson tests - it was incorrectly failing.

giuliobenetti:
> @giuliobenetti We had a bit of an issue with meson tests - it was incorrectly failing.

Oops, I assumed it worked correctly and didn't test with and without nptl. So I've setup a glibc and a uclibc buildroot with libfuse-3.17.2 and your patch. I comment on the code for what I find.

meson.build
```diff
@@ -59,6 +59,8 @@ include_default = '''
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <assert.h>     # For static_assert
```
giuliobenetti, line 62: `# For static_assert` should be `/* # For static_assert */` otherwise a warning of extra token is thrown. Same for the other include

meson.build
```diff
@@ -72,32 +74,64 @@ private_cfg.set_quoted('PACKAGE_VERSION', meson.project_version())
...
+            unsigned int flags = CLOSE_RANGE_UNSHARE;
+            return close_range(3, ~0U, flags);
+        }
+    '''
+}
+
+foreach name, code : special_funcs
+    private_cfg.set('HAVE_' + name.to_upper(),
+        cc.compiles(code,
+                   args: ['-D_GNU_SOURCE', '-I/usr/include'] + args_default,
```
giuliobenetti, line 114: This gives problems with cross-compilation, because it looks into host folder

giuliobenetti: Indeed in Buildroot it can't find any of the 3 special functions

giuliobenetti: '-I/usr/include' on Buildroot side is useless and I think it's useless in general because that should be a standard path of the toolchain. Am I right?

giuliobenetti: cc.compiles() is not enough, we could go with it with -Werror or even better cc.links(), that way I've tested with NPTL and linuxthread and it recognizes correctly pthread_setname_np(). As is now it doesn't work because of:
```
- cc.compiles()
- -I/usr/include
```

giuliobenetti (changes requested):
Without '-I/usr/include' everything works as expected with Buildroot. I've built 2 toolchains, one with NPTL and another with linuxthreads. NPTL provides pthread_setname_np() while linuxthreads not.

meson.build
bsbernd: Thanks! Hmm, interesting that it worked at all.

meson.build
bsbernd: Thank you, I changed to `cc.links()`, added `-Werror` and removed `-I/usr/include`

giuliobenetti: Ok, just FYI, -Werror was mandatory to treat the missing function prototypes as errors. But since you're using cc.links() here you can remove -Werror. We only need to make sure cc.links() is there since the beginning of meson since I've seen other projects using cc.compiles() with -Werror option enabled instead of cc.links().
Anyway it works as is, simply it returns with error when it can't find the function prototype instead of failing during linking.

Thank you for the solution provided that fixed my patch :-)

bsbernd: I don't have a strong opinion about it - changed it back to cc.compiles with the -Werror. Thank a lot for your help on this!

PR #1210, opened 2025-04-25
meson.build: make special_funcs check more reliable

giuliobenetti:
Unfortunately while cross-compiling with build tools like Buildroot it happens to have repeated flags or anything that could lead to a warning. This way the check fails because of a warning not related to the special function. So let's use cc.links() and increase minimum meson_version to 0.60 since cc.links() has been added during that version.

--
Pardon @bsbernd!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('libfuse3', ['c'],
         version: '3.18.0-rc0',  # Version with RC suffix
-        meson_version: '>= 0.51.0',
+        meson_version: '>= 0.60.0',
         default_options: [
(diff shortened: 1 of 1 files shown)


bsbernd (approved):

bsbernd:
Thanks @giuliobenetti !

giuliobenetti:
> Thanks @giuliobenetti !

Sorry  I didn't catch it before :-)

bsbernd:
No worries at all,  we would need to have a huge test matrix here in github to catch that - will hard. So you work and help is highly welcome :)

PR #1216, opened 2025-04-29
fuse_lowlevel.c: Fix a small spelling mistake

izxl007:


bsbernd:
Thanks @izxl007 I actually need to test if interrupts work.

PR #1217, opened 2025-04-30
Ensure want is synced to want_ext for applications directly modifying them

sur5r:
This is a related to #1171 and #1172.

Rationale behind this change is:

If an application uses both high-level functions to modify `want` and direct bit manipulation, the check in `convert_to_conn_want_ext` will fail. Seen in ceph-fuse.

`fuse_apply_conn_info_opts` has to keep `want` and `want_ext` in sync so direct modifications to `want` can be analyzed.

Now if `want` has additional bits set, those can be copied over to `want_ext` as the intention of the application is clear.

### DIFF
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -257,6 +257,15 @@ static inline int convert_to_conn_want_ext(struct fuse_conn_info *conn,
 					   uint64_t want_ext_default,
 					   uint32_t want_default)
 {
+	/* Check if want has bits set that want_ext doesn't.
+	 * In this case assume the application has used a mix of functions
+	 * and direct modification of want and copy those bits over to want_ext
+	 */
+	uint32_t want_diff = conn->want ^ fuse_lower_32_bits(conn->want_ext);
+
+	if ((conn->want_ext & want_diff) == 0)
+		conn->want_ext |= want_diff;
+
 	/*
 	 * Convert want to want_ext if necessary.
 	 * For the high level interface this function might be called


bsbernd:
@sur5r Thank you, let's ignore the checkpatch issues for now. Sorry couldn't look into it yesterday evening anymore. I would first like to understand why it it is failing for ceph-fs and why existing tests are not catching it.

bsbernd:
I will look into it in the evening. The interesting part is that I had especially copied `example/hello.c` to tests/, to test the old way of setting flags - which is want ceph seems to do. Hmm.

sur5r:
AIUI, it's not the old way per se. It's the combination of `fuse_apply_conn_info_opts` which uses the new way internally and *then* setting flags the old way.

lib/fuse_i.h
```diff
@@ -257,6 +257,15 @@ static inline int convert_to_conn_want_ext(struct fuse_conn_info *conn,
...
 					   uint32_t want_default)
 {
+	/* Check if want has bits set that want_ext doesn't.
+	 * In this case assume the application has used a mix of functions
+	 * and direct modification of want and copy those bits over to want_ext
+	 */
+	uint32_t want_diff = conn->want ^ fuse_lower_32_bits(conn->want_ext);
+
+	if ((conn->want_ext & want_diff) == 0)
+		conn->want_ext |= want_diff;
```
bsbernd, line 267: Issue is that the application might delete default flags from either of both and this might add them back.

sur5r: Yes, that's unfortunate. It seems the whole `want_ext` stuff is too complex for a minor release as it breaks both ABI and API ways impossible to fix.

I wonder if this should be guarded by a feature flag (`#define FUSE_ONLY_WANT_EXT` or somesuch) so applications can declare that they don't touch `want`.

bsbernd: Well, my personal wish was to entirely break the ABI and to make 'want' 64-bit. We had the .so bump anyway. Issue is that the creator of this project needs compatibility for an enterprise distro. 
I *think* your change in `fuse_apply_conn_info_opts` should be enough. I'm going to create test for these two functions (fuse_parse_conn_info_opts/fuse_apply_conn_info_opts).

If nothing works, we would need another .so bump (to .so.5) and do what I initially planned.

sur5r: No, it's not enough for the `ceph-fuse` case. That was my first attempt.

The sequence in that code is:

1. `fuse_parse_cmdline`
2. `fuse_parse_conn_info_opts`
3. In `fuse_lowlevel_ops->init` (`do_init`): (each guarded by a check against `capable`)
   1. `want |= FUSE_CAP_SPLICE_MOVE`
   2. `want |= FUSE_CAP_DONT_MASK`
   3. `want |= FUSE_CAP_EXPORT_SUPPORT`

Now `want` has up to three bits set which `want_ext` doesn't.

sur5r: I have to admit I'm a bit confused why my patch to `fuse_parse_conn_info_opts` is necessary in the first place as `conn.want` gets initialized from `conn.want_ext` in `_do_init`...

sur5r: Another finding: The docs for `fuse_apply_conn_info_opts` only mention `want`, not `want_ext`. So the current state of the code does not match the expectations of library users.

bsbernd: Ok yeah, with fuse_parse_conn_info_opts() want and want_ext are changed from default ans then also other flags added to either of want or want_ext. So just updating there won't help. I think a solution is to move convert_to_conn_want_ext() into a C file and then to make want_ext_default and want_default global variables (or with setters) - fuse_parse_conn_info_opts() needs to update them. All not beautiful.

sur5r: See https://github.com/sur5r/libfuse/commit/7c84c67c696e222441581101c568b14cf42effde for a variant that works with `ceph-fuse` as well.

bsbernd: That commit has the issue that it might add flags that the application might not want, not an ideal solution either. https://github.com/sur5r/libfuse/commit/7c84c67c696e222441581101c568b14cf42effde

And yeah, changing want works, because ceph sets want in its init method. But for example gvfsd already got update and uses fuse_set_feature_flag(), which sets conn_ext. Too late for the fuse API, but not ideal that structs are exposed at all - we need a time machine ;)

bsbernd: @sur5r didn't manage to work on it yet - will do during a long flight tomorrow.

gcsideal: > @sur5r didn't manage to work on it yet - will do during a long flight tomorrow.

Any news? It was two weeks before.

bsbernd: https://github.com/libfuse/libfuse/pull/1223

bsbernd: Sorry, hadn't managed to complete it during the flights and had been too busy the other days.

bsbernd:
@sur5r sorry for the long delay, would you have time to test https://github.com/libfuse/libfuse/pull/1223?

sur5r:
Yes, already saw the new PR. Didn't have time today, will try and test tomorrow.

gcsideal:
> Yes, already saw the new PR. Didn't have time today, will try and test tomorrow.

Please note that it is not straightforward to backport the changes for v3.17.2 - at least I'm not sure I'm doing it correctly. It may also change the ABI.

bsbernd:
@gcsideal https://github.com/libfuse/libfuse/pull/1225

bsbernd:
Regarding ABI, hmm, the ABI checker didn't complain, but yeah, I see that previously inlined functions now become part of the API/ABI. Dunno what to do about it - I don't see a way out.

gcsideal:
> @gcsideal #1225

Thanks, proposed Debian package is ready [1]. Please test it @sur5r .
[1] dget -x https://people.debian.org/~gcs/fuse3_3.17.2-2.dsc

gcsideal:
> @gcsideal Thanks. I did a quick test and it's looking good so far.

Thanks for the confirmation. For some reason, when I upgraded fuse and rebooted, I got this for fuse mount points:
d????????? ? ?   ?     ?            ? doc
d????????? ? ?   ?     ?            ? gvfs

Then downgraded to 3.17.2 and the next reboot it was working as expected. Upgraded fuse again and now everything are correct. I don't know what was the intermediate case.
Final question if there will be a 3.17.3 release (soon) or not.

bsbernd:
@gcsideal Ooops, that is weird. If you should ever get this again, could you run in foreground with debugging enabled (`-d`) to get some basic debug output for me? I think there had been readdir-plus issue that Amir had fixed that could result in that behavior, but would be good to see a bit more context.

Very unlikely that I have time to release 3.17.3 today, but tomorrow or on Thursday. At some point I need to work on a simple action button to make the release...

Going to merge my PRs for this now.

PR #1219, opened 2025-05-12
RFC: fuse_lowlevel: added helpers to support FUSE server restart

luis-henrix:
This pull-request is just an RFC, as I'm not really sure there's interest in having support for this in libfuse.

I've spent some time trying to understand what would be required in order to restart a libfuse server.  And the functions added with this commit seem to be the minimal set, specially the one that re-initializes the (opaque) session object.

My understanding is that a FUSE server needs to keep track of the `/dev/fuse` file descriptor, the `struct fuse_conn_info` with the kernel parameters, and the file-system mount point.  Then, libfuse simply needs to set the corresponding fields in the `struct fuse_session`.  Obviously, the **actual** work is file-system dependent, and a file-system needing to recover it's state probably needs to keep track of much more data.

### DIFF
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -1894,6 +1894,43 @@ int fuse_lowlevel_notify_store(struct fuse_session *se, fuse_ino_t ino,
 int fuse_lowlevel_notify_retrieve(struct fuse_session *se, fuse_ino_t ino,
 				  size_t size, off_t offset, void *cookie);
 
+/**
+ * Resend all pending requests
+ *
+ * Ask the kernel to resend all requests that are currently pending.  This
+ * is useful mostly for FUSE servers that try to recover previous
+ * connections when, for example, the server crashed.
+ *
+ * @param se the session object
+ */
+int fuse_lowlevel_notify_resend(struct fuse_session *se);
+
+/**
+ * Verify if a request is a new request or a resent.
+ */
+int fuse_lowlevel_req_is_resend(fuse_req_t req);
+
+/**
+ * Re-initialize a FUSE session using a previously stored file descriptor
+ *
+ * Using a FUSE connection that resulted from a previously initialized
+ * session, initialize a new session object.  This is useful when trying to
+ * recover a session that was previously broken, for example after a server
+ * crash or after a server upgrade.
+ *
+ * This function can be used *instead* of the fuse_session_mount(), whic is
+ * used for new FUSE sessions.  Note that all the parameters but the session
+ * need to be explicitly store by the FUSE server from the session it is
+ * trying to recover.
+ *
+ * @param se a new session created with fuse_session_new()
+ * @param conn a connection object kept from a previous session
+ * @param fd the file descriptor from the previous session
+ * @param mountpoint the mount point to restore
+ */
+int fuse_session_reinitialize(struct fuse_session *se,
+			      struct fuse_conn_info *conn,
+			      int fd, const char *mountpoint);
 
 /* ----------------------------------------------------------- *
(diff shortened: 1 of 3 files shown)


bsbernd:
Hi Luis, yes I think we will need this, but I think the entire restart approach is very unsafe right now - it does not handle LOOKUP and FORGET - fuse-client/kernel also needs to restore all existing LOOKUPs.

luis-henrix:
> Hi Luis, yes I think we will need this, but I think the entire restart approach is very unsafe right now - it does not handle LOOKUP and FORGET - fuse-client/kernel also needs to restore all existing LOOKUPs.

I haven't delve too deep into this, so I'm not sure I understand what's missing from the kernel.  Couldn't that be handled by simply invalidating **all** the dentries?

luis-henrix:
> If all the inodes are given up, resending requests doesn't work - resend
> would be for given up inodes.

Sorry for getting back to this after a long time, but I've been thinking
again on this topic.  While working a bit more on my patch that adds a
work queue to do the dentries clean-up[1], I realised that the
infra-structure it adds could be used to support the FUSE server restart.

Here's what I thought:

If a FUSE connection keeps track of all its dentries (as my patch is
doing), then, when the kernel gets a `NOTIFY_RESEND`, it could also send a
`LOOKUP` for each of its dentries.  Sure, it doesn't really need to be a
real `LOOKUP`, but maybe something more like a `BATCH_LOOKUP`.  Or this
should most likely be part of a `NOTIFY_RESEND_v2` so that the old
behaviour isn't modified.

I haven't thought about any of the details, and I'm sure this idea has a
lot of holes.  But my understanding is that this would address your
concerns regarding this draft PR.  Do you have any thoughts on this?  Or
did I misunderstood your comments?

Oh, and this approach would also require my work queue patch to be
accepted, of course :-)

[1] ***@***.***/

Cheers,
-- 
Luís

bsbernd:
I think almost - dentries are probably not suitable. Let's assume the last inode link was removed, but the inode is still opened. Or let's assume multiple hard links. I.e. we want to operate on inode level and not dentry level - using your first patch versions that went over all inodes is probably more suitable?

luis-henrix:
On Fri, Jul 11 2025, Bernd Schubert wrote:

> bsbernd left a comment (libfuse/libfuse#1219)
>
> I think almost - dentries are probably not suitable. Let's assume the last inode
> link was removed, but the inode is still opened. Or let's assume multiple hard
> links. I.e. we want to operate on inode level and not dentry level - using your
> first patch versions that went over all inodes is probably more suitable?

Right, I see what you mean, it makes sense of course.  But this probably
brings the opposite scenario: while the FUSE server is restarting a new
file has been created on the file system (for example, in a networked file
system), and the kernel had a negative dentry for it because there was a
previous lookup.

I guess I'll have to work on an RFC kernel patch to try to this out and
see what works and what doesn't.

Cheers,
-- 
Luís

bsbernd:
The lookup count is for inodes. The issue here is `fuse_evict_inode` after fuse server restart. Lot's of file systems store the pointer to the inode into fi->fh, which then becomes the node id on the kernel side. If these file systems receive a forget with a node-id (i.e. pointer to an inode) that does not exist, it would just trigger random memory access - either crash or all weird things going on. So on fuse-server restart the lookup needs to be restored - based on the inode and not dentry.

luis-henrix:
OK, I've spent already some time playing with this. I will now send an RFC email (not a patch) to the fsdevel mailing-list.  I want to see whether what I'm doing is 1) what you actually had in mind and 2) if this would be acceptable in mainline.  Thanks a lot for the feedback, @bsbernd

PR #1228, opened 2025-05-22
Fix build in musl libc

luis-henrix:
Function fuse_set_thread_name() assumes that pthread_t is an unsigned long and fails to compile in musl libc with the following:

```
../lib/util.c: In function 'fuse_set_thread_name': ../lib/util.c:48:28: error: passing argument 1 of \
      'pthread_setname_np' makes pointer from integer without a cast [-Wint-conversion]
```
Fix fuse_set_thread_name() by using pthread_t type instead.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -4840,7 +4840,7 @@ static void *fuse_prune_nodes(void *fuse)
 	struct fuse *f = fuse;
 	int sleep_time;
 
-	fuse_set_thread_name(pthread_self(), "fuse_prune_nodes");
+	fuse_set_thread_name("fuse_prune_nodes");
 
 	while(1) {
 		sleep_time = fuse_clean_cache(f);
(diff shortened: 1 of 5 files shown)


bsbernd:
@luis-henrix I had done it like this because of commit 2dcc57693c3ff1d4bcaa52f103633b1fc0042ba4 from @giuliobenetti . I.e. I had tried to avoid including a pthread dependency here. Would it work to cast?

bsbernd:
@luis-henrix Ah yes, dropping the id the better way, thank you!

luis-henrix:
> @luis-henrix Ah yes, dropping the id the better way, thank you!

Heh! you're too fast for me! I was writing a reply to your original question :-)

Anyway, the cast would work: the patch would need to cast every `pthread_self()` to `unsigned long` and then cast `tid` back to `pthread_t`.  Which is a bit ugly, I guess.

PR #1231, opened 2025-05-28
Allow applications to retrieve the req payload (io-uring only)

bsbernd:
With io-uring the req owns the payload buffer, the application can directly access it and copy data into it.

fuse_buf_copy_one() already has a check for dstmem == srcmem and skips data copies.

fuse_reply_data
   fuse_reply_data_uring
       fuse_buf_copy
           fuse_buf_copy_one

### DIFF
--- a/lib/fuse_uring.c
+++ b/lib/fuse_uring.c
@@ -190,6 +190,30 @@ static int fuse_uring_commit_sqe(struct fuse_ring_pool *ring_pool,
 	return 0;
 }
 
+int fuse_req_get_payload(fuse_req_t req, char **payload, size_t *payload_sz,
+			 void **mr)
+{
+	struct fuse_ring_ent *ring_ent;
+
+	/* Not possible without io-uring interface */
+	if (!req->is_uring)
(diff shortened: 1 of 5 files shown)


lib/fuse_uring.c
```diff
@@ -181,6 +181,29 @@ static int fuse_uring_commit_sqe(struct fuse_ring_pool *ring_pool,
...
 }
 
+int fuse_req_get_payload(fuse_req_t req, void **payload, size_t *payload_sz,
+			 void **mr)
+{
+	struct fuse_ring_ent *ring_ent =
+		container_of(req, struct fuse_ring_ent, req);
+
+	/* Not possible without io-uring interface */
+	if (!req->is_uring)
```
izxl007, line 191: It would be better to move this check to the beginning of the function.

bsbernd: Thanks for spotting it, fixed. I will first add some usage for that to fuse_passthrough_hp.cc before merging it. Just urgently needed it yesterday for our file system.

PR #1240, opened 2025-06-05
mount_bsd.c: Fix race between actual mounting and returning to the caller

arrowd:
I was trying to debug an application that mounts a FUSE filesystem and then immediately tries to access it. I was constantly getting ENOENT after mounting despite `fuse_session_mount()` returning success.

It turned out that the code doesn't really wait for the actual mounting program to finish, which allows the caller to proceed with execution before the FS gets mounted.

### DIFF
--- a/lib/mount_bsd.c
+++ b/lib/mount_bsd.c
@@ -187,7 +187,7 @@ mount:
 		if (pid == 0) {
 			const char *argv[32];
 			int a = 0;
-			int ret = -1; 
+			int ret = -1;
 
 			if (! fdnam)
 			{
@@ -214,6 +214,7 @@ mount:
 			_exit(EXIT_FAILURE);
 		}
 
+		waitpid(pid, &status, 0);
 		_exit(EXIT_SUCCESS);
 	}
 


bsbernd:
Please correct me if I'm wrong, but I don't think we need the double fork if we add a waitpid there.

arrowd:
Yes, I'm now obscured by this too. @asomers do you have an idea why double-forking is used there?

bsbernd:
@arrowd something like this?

```
static int fuse_mount_core(const char *mountpoint, const char *opts)
{
    const char *mountprog = FUSERMOUNT_PROG;
    long fd;
    char *fdnam, *dev;
-	pid_t pid, cpid;
+	pid_t pid;
    int status;
    int err;

    fdnam = getenv("FUSE_DEV_FD");

    if (fdnam) {
        err = libfuse_strtol(fdnam, &fd);
        if (err || fd < 0) {
            fuse_log(FUSE_LOG_ERR, "invalid value given in FUSE_DEV_FD\n");
            return -1;
        }

        goto mount;
    }

    dev = getenv("FUSE_DEV_NAME");

    if (! dev)
        dev = (char *)FUSE_DEV_TRUNK;

    if ((fd = open(dev, O_RDWR)) < 0) {
        perror("fuse: failed to open fuse device");
        return -1;
    }

mount:
    if (getenv("FUSE_NO_MOUNT") || ! mountpoint)
        goto out;

    pid = fork();
-	cpid = pid;

    if (pid == -1) {
        perror("fuse: fork() failed");
        close(fd);
        return -1;
    }

    if (pid == 0) {
-		pid = fork();
-
-		if (pid == -1) {
-			perror("fuse: fork() failed");
-			close(fd);
-			_exit(EXIT_FAILURE);
-		}
-
-		if (pid == 0) {
-			const char *argv[32];
-			int a = 0;
-			int ret = -1; 
-
-			if (! fdnam)
-			{
-				ret = asprintf(&fdnam, "%ld", fd);
-				if(ret == -1)
-				{
-					perror("fuse: failed to assemble mount arguments");
-					close(fd);
-					_exit(EXIT_FAILURE);
-				}
-			}
-
-			argv[a++] = mountprog;
-			if (opts) {
-				argv[a++] = "-o";
-				argv[a++] = opts;
-			}
-			argv[a++] = fdnam;
-			argv[a++] = mountpoint;
-			argv[a++] = NULL;
-			execvp(mountprog, (char **) argv);
-			perror("fuse: failed to exec mount program");
-			free(fdnam);
-			_exit(EXIT_FAILURE);
-		}
-
-		_exit(EXIT_SUCCESS);
+		/* Child process */
+		const char *argv[32];
+		int a = 0;
+		int ret = -1; 
+
+		if (! fdnam) {
+			ret = asprintf(&fdnam, "%ld", fd);
+			if(ret == -1) {
+				perror("fuse: failed to assemble mount arguments");
+				close(fd);
+				_exit(EXIT_FAILURE);
+			}
+		}
+
+		argv[a++] = mountprog;
+		if (opts) {
+			argv[a++] = "-o";
+			argv[a++] = opts;
+		}
+		argv[a++] = fdnam;
+		argv[a++] = mountpoint;
+		argv[a++] = NULL;
+		execvp(mountprog, (char **) argv);
+		perror("fuse: failed to exec mount program");
+		free(fdnam);
+		_exit(EXIT_FAILURE);
    }

-	if (waitpid(cpid, &status, 0) == -1 || WEXITSTATUS(status) != 0) {
+	/* Parent process */
+	if (waitpid(pid, &status, 0) == -1 || WEXITSTATUS(status) != 0) {
        perror("fuse: failed to mount file system");
        if (close(fd) < 0)
            perror("fuse: closing FD");
        return -1;
    }

out:
    return fd;
}````

bsbernd:
@vassilit was also recently looking at this code. Double forking something like a standard technique to avoid zombies, if the parent gets killed before calling waitpid. Maybe your suggestion to add a waitpid is the optimal solution.

bsbernd:
@arrowd I merged it, as it is obviously correct, we can simplify the code any time later.

asomers:
With this change, does the mount program wait until after the server sends the `FUSE_INIT` response?

bsbernd:
`FUSE_INIT` is only send after mount is complete. In general, waiting to the complete the mount till FUSE_INIT has been received and responded to makes a lot sense, but needs work.

With io-uring it is even more painful - the mount is only accepting requests when fuse-server (libfuse) has done ring setup - after `FUSE_INIT`.

vassilit:
I made a simple program to test the readiness of the filesystem after the forked fuse_main() process returns: https://gist.github.com/vassilit/e00d59135e3204aabcc4cca8c333c140

We see that on FreeBSD, waiting for FUSE_INIT helps. See also https://github.com/libfuse/libfuse/discussions/1100

`waitpid()` on `mount_bsd.c` used to be prone to a deadlock (see https://github.com/libfuse/libfuse/pull/1146#issuecomment-2691801292), hence the initial double-fork and no waitpid. However, I think this comment is now irrelevant and waiting for mount_fusefs to returns is safe. We should check the status however, as mount_fusefs can fail.

PR #1243, opened 2025-06-09
help: Remove duplicate io_uring options from fuse_cmdline_help

izxl007:
The io_uring options are currently shown in both fuse_cmdline_help() and fuse_lowlevel_help(), which creates unnecessary duplication. Since io_uring is a low-level I/O feature, it makes more sense to only show these options in fuse_lowlevel_help().

This change:
- Removes io_uring options from fuse_cmdline_help()
- Keeps them in fuse_lowlevel_help() where they belong
- Removes the FIXME comment that is no longer needed

This is purely a documentation improvement and does not affect any functionality. Users will still see all available options when using --help, just organized in a more logical way.

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -139,11 +139,7 @@ void fuse_cmdline_help(void)
 	       "    -o max_idle_threads    the maximum number of idle worker threads\n"
 	       "                           allowed (default: -1)\n"
 	       "    -o max_threads         the maximum number of worker threads\n"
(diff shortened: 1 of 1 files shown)


bsbernd:
These options are also valid for high level interface.

bsbernd:
Thanks @izxl007 I'm merging it. Actually makes sense, I also see the double flags. I though I had seen missing options for some example/ - we should better fix the reason for that then.

PR #1247, opened 2025-06-13
mount_util.c: check if utab exists before update

httpstorm:
Do not attempt to update `/run/mount/utab` if it doesn't exist.
Note: if this path ever changes, `utab` updates will break.

Fixes the following error when mounting iPhone documents and containers using `ifuse`:
``` bash
ifuse /mnt --container com.httpstorm.httpstorm
mount: mounting ifuse on /mnt failed: Invalid argument
```

On OpenWRT by default `mount-utils` is not installed and `utab` does not exist.
`/bin/mount` is a symlink to `/bin/busybox` and does not support updating of `utab`.
If `mount-utils` is installed `/run/mount/` exists, but does not contain `utab`.
The `mount-utils` instance is under `/usr/bin/mount`, so a hard-coded call to `/bin/mount` will still fail.
Using `/usr/bin/mount` succeeds but does not create `utab`.

I am not sure if this is the correct approach of solving the issue.
The regression was introduced in 74b1df2e84e836a1710561f52075d51f20cd5c78 which does not explain why the symlink check was removed.
The original symlink check was introduced in 7f571e320e39a5dd0bb7ab2c1b2a40b174879b18.

Build and run tested on OpenWRT main, using `ifuse` master, where `mtab` is indeed a symlink:
``` bash
ll /etc/mtab 
lrwxr-xr-x    1 root     root            12 Jun 12 03:58 /etc/mtab -> /proc/mounts
```

@bsbernd @Nikratio

### DIFF
--- a/lib/mount_util.c
+++ b/lib/mount_util.c
@@ -75,6 +75,10 @@ static int mtab_needs_update(const char *mnt)
 
 		if (err == EROFS)
 			return 0;
+
+		res = access("/run/mount/utab", F_OK);
+		if (res == -1)
+			return 0;
 	}
 
 	return 1;


bsbernd (changes requested):
I'm afraid not so easy, see https://github.com/libfuse/libfuse/pull/894

I.e. restoring the symlink check would introduce regression, there is a longer discussion about it in that thread. The ideal way would be probably to detect util-linux, but I don't know how to do that. Maybe check if not linux? Or an meson configure option that triggers automatically on OSX?

bsbernd:
Hmm ok, so this happens with OpenWRT on linux? OpenWRT is not using util-linux then?

bsbernd:
I forgot the name of this almighty binary that is used in initramfs and embedded projects, maybe we can add a detection for it?

httpstorm:
I am building OpenWRT master for WRT3200ACM on macOS host. I'll check the thread, but please keep in mind that I don't have any experience with `util-linux`, `meson` and friends.

bsbernd:
If we agree on an approach I might be able to help.
What is actually giving the error, libfuse on osx or libfuse inside of OpenWRT (my last usage of OpenWRT was about 15 years ago, without looking it up, I guess it is still Linux based).

httpstorm:
Yes, OpenWRT is Linux based. The error is on OpenWRT when I run `ifuse`. I have the build environment ready so if you come up with any ideas, I can test them.
First some host tools and the toolchain are compiled along with a cross compiler for the target architecture. Next these are used to compile the Linux kernel followed by target packages. Finally a firmware image is created.

bsbernd:
So "mount" is a symlink to /bin/busybox on your system? We either add a detection for that or we run "mount --version" (using posix_spawn)  and then check for util-linux.

httpstorm:
`mount` is indeed a symlink to `busybox`:
``` bash
which -a mount
/bin/mount
lrwxr-xr-x    1 root     root             7 Jun 12 03:58 /bin/mount -> busybox*
```

I don't think that checking for `busybox` is a good idea. OpenWRT can be compiled with the full featured `mount` tool.
On the other hand, checking if `mtab` is a symlink and the target is read-only should work:

The `nano` editor says it is read-only and attempting to change it returns an error:
``` bash
/proc/mounts is meant to be read-only
Error writing /proc/self/mounts: Invalid argument

lrwxr-xr-x    1 root     root            12 Jun 12 03:58 /etc/mtab -> /proc/mounts
lrwxrwxrwx    1 root     root            11 Jun 12 03:58 /proc/mounts -> self/mounts
-r--r--r--    1 root     root             0 Jun 13 13:12 /proc/self/mounts
```

`/proc/self/mounts` is updated automatically on OpenWRT. I am using this patch.

``` bash
ifuse /mnt --documents com.httpstorm.httpstorm
cat /proc/self/mounts
…
ifuse /mnt fuse.ifuse rw,nosuid,nodev,relatime,user_id=0,group_id=0 0 0
```

bsbernd:
/proc/mounts is always symlink on modern system. The point is the util-linux doesn't even try to update /etc/mtab, but uses `/run/mount/utab`. Seems like busybox does not support that and therefore needs the symlink check. Enabling just the symlink check in general is not good either, as it would break adding custom x-* options. So the generic solution is either to check if mount is a symlink, or better to parse the version output of mount to check if it is util-linux.

httpstorm:
I did some tests:
- On Ubuntu 24.04.2 LTS: `/proc/self/mount` is read-only, `/run/mount/utab` is writable only by root
- On OpenWRT: `/proc/self/mount` is read-only, `/run/mount` do not exist

If I disable `busybox/mount` and compile with `mount-utils`, `/run/mount` is there, but `/run/mount/utab` does not exist.
I think I found another bug in `libfuse3`: it is hard-coded to use `/bin/mount`, which does not exist when `busybox/mount` is disabled:
``` bash
ifuse /mnt --container com.httpstorm.httpstorm 
fuse: failed to execute /bin/mount: No such file or directory

which -a mount
/usr/bin/mount

ln -s /usr/bin/mount /bin/mount
which -a mount
/usr/bin/mount
/bin/mount

# with the full featured mount, this is successful even without my patch
ifuse /mnt --container com.httpstorm.httpstorm
```

At this point I agree that checking if `mount` is a symlink to `busybox` will solve the issue.
The other option I see is to check if `/run/mount/utab` exists. If it doesn't `return 0`:
``` C
if (S_ISLNK(stbuf.st_mode) && does_not_exist("/run/mount/utab"))
	return 0;
```

httpstorm:
Checking if `mount` is a link to `busybox` will work, but first we need to decide if we are going to use `/usr/bin/mount` or
`/bin/mount`. Then there is the complication that this is a `setuid` binary, so we need to be careful what we run. We also need to be consistent if we the check is performed in both `mtab_needs_update` and `add_mount`. It would be better if we don't have to check twice. We can also use `access("/run/mount", R_OK)` or `access("/run/mount/utab", R_OK)` or `W_OK` in `mtab_needs_update`, but if both versions of `mount` are present, the decision should rather be based on the one we choose to use. Another reason not to use `access` from the man page (though for a simple check if the file exists, it should be ok):
> set-user-ID and set-group-ID applications should restore the effective user or group ID, and perform
     actions directly rather than use access() to simulate access checks for the real user or group ID.

On OpenWRT there is a config option to make `/etc/mtab` a regular file (usually if the kernel is compiled without `proc`, though such use is discouraged. `CONFIG_BUSYBOX_CONFIG_FEATURE_MTAB_SUPPORT`

httpstorm:
``` bash
mount --version
# busybox mount does not support --version, it prints the list of mount points
# the full featured mount reports util-linux
mount from util-linux 2.41 (libmount 2.41.0: btrfs, namespaces, idmapping, fd-based-mount, assert, debug)
```

bsbernd:
@httpstorm Sorry for my late reply. Are you sure the patch is right, because you had written above
>If I disable busybox/mount and compile with mount-utils, /run/mount is there, but /run/mount/utab does not exist.

httpstorm:
@bsbernd Hey, no worries. I'm glad to have you back.
Yes, this patch fixes the errors on OpenWRT. There is no `utab` file, so that feature won't work.
If you like, later we can introduce another patch to detect what version and path is used for the `mount` command.
Or at least avoid calling `/bin/mount` if it doesn't exist. But considering the above, it won't matter for OpenWRT.

I wrote a small program to test the calls on Ubuntu as regular user, root and `setuid` binary, to make sure I know how it works and that there is no change. As long as the hard-coded path to `/run/mount/utab` remains the same, there should be no regression.

I also tested OpenWRT where the new behaviour is as follows:
1. With `busybox`
`/run/mount` does not exist, so `access` returns -1, `mtab_needs_update` returns 0, `/bin/mount` does not support the parameters and would fail, but it is not called.

2. With `mount-utils`
`/run/mount/utab` does not exist, so `access` returns -1, `mtab_needs_update` returns 0, `/bin/mount` does not exist and calling it would fail, but is not called. I'm not aware if `utab` can be enabled through configuration. By default it does not exist. 

3. Both `busybox` and `mount-utils` (theoretical, not tested)
`/run/mount/utab` does not exist, so `access` returns -1, `mtab_needs_update` returns 0, `/bin/mount` does not support the parameters and would fail, but it is not called.

bsbernd:
Thank you @httpstorm for the detailed description and test. We can also cherry-pick it to 3.17.x and new release definitely follows this week. 
However, I also want to check what exactly creates utab, I wonder a bit if it is dynamically created and that we might take the wrong decision if "/" would be a fuse file system (I have done these things in the past myself).

httpstorm:
I have a [PR](https://github.com/openwrt/packages/pull/26762) to merge the fixes in OpenWRT. A new release is definitely welcome, so we don't have to maintain patches. Would you like me to cherry pick 6787608e5655b3d833a541a9f0d5a3159ae51f0e and 3793b1748ad151c8043dee1db198fffa3dbb5a67 for 3.17.x?

I can try to do more research about `utab` on OpenWRT, but I'm lost. There might be some kernel module or configuration that needs to be enabled. Or I can give you guidance to configure and build for x64 [1, 2]. You can test it on a VM or boot from flash-drive.
Note that OpenWRT is targeted for routers and embedded devices, so it's optimised to be light-weight. Many features can be enabled through `make menuconfig` or `make kernel_menuconfig`. It is pretty usable for servers too.

[1] [OpenWRT Build System Usage](https://openwrt.org/docs/guide-developer/toolchain/use-buildsystem)
[2] [OpenWRT build environment on macOS](https://httpstorm.com/openwrt/help/OpenWRT%20build%20on%20macOS.htm)

httpstorm:
@bsbernd 
I noticed a new release `fuse-3.17.3` is finally out, however while preparing a PR for OpenWRT, I saw that none of my fixes made it there.

[1] https://github.com/libfuse/libfuse/commit/6787608e5655b3d833a541a9f0d5a3159ae51f0e
[2] https://github.com/libfuse/libfuse/commit/3793b1748ad151c8043dee1db198fffa3dbb5a67

bsbernd:
@httpstorm Just back from vacation yesterday. Sorry, looks like I had missed the patches when cherry picking commits. Let's aim for another release end of this week. Also feel free to do cherry picks yourself (ideally with git cherry-pick -x).

httpstorm:
@bsbernd 
I'm doing cherry-pick for the first time, so please let me know if I managed to do it correctly. 🍒 https://github.com/libfuse/libfuse/pull/1309

sertonix:
For the record: When util-linux umount was installed and is replaced with busybox umount (without reboot) the way that the check is done still detects the system as using util-linux which causes umount to fail (happened on an alpine linux system). I can't think of a better check so not opening an issue

PR #1248, opened 2025-06-13
gitignore: add *.patch

httpstorm:
When working with patches, git shows them as uncommitted changes.
Ignore `*.patch` to keep the list of changes tidy.

@bsbernd @Nikratio

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 *.gz
 \#*#
 *.orig
+*.patch
 *~
 .DS_Store
 Makefile.in


bsbernd:
Thanks, should not hurt. Btw, no need to CC me, github sends me a notification automatically. And Nikolaus got too busy to be maintainer.

httpstorm:
Thanks for the quick merge! I guess GitHub adds CC when I use `@` . It's a common practice in OpenWRT. I guess I should avoid it here.

PR #1251, opened 2025-06-16
fuse: add support to FUSE_NOTIFY_INC_EPOCH

luis-henrix:
This patch adds support for the FUSE INC_EPOCH notify.  This new operation simply increments the FUSE connection epoch value, allowing to invalidate all the dentries next time they are revalidated.

### DIFF
--- a/example/notify_inval_entry.c
+++ b/example/notify_inval_entry.c
@@ -67,6 +67,12 @@
  * To use the function fuse_lowlevel_notify_expire_entry() instead of
  * fuse_lowlevel_notify_inval_entry(), use the command line option --only-expire
  *
+ * Another possible command-line option is --inc-epoch, which will use the FUSE
+ * low-level function fuse_lowlevel_notify_increment_epoch() instead.  This will
+ * function will force the invalidation of all dentries next time they are
+ * revalidated.  Note that --inc-epoch and --only-expire options are mutually
+ * exclusive.
+ *
  * ## Compilation ##
  *
  *     gcc -Wall notify_inval_entry.c `pkg-config fuse3 --cflags --libs` -o notify_inval_entry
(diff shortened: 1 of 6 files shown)


bsbernd:
Thanks @luis-henrix ! Would it be possible to add a test, similar to the other notify files.

luis-henrix:
> Thanks @luis-henrix ! Would it be possible to add a test, similar to the other notify files.

Sure, I can have a look at it and see what I can do.  It may be tricky to
validate the result, as it's not very visible from user-space.

Cheers,
-- 
Luís

bsbernd:
Just a small test to check the if the call works

bsbernd:
Or kind of usage example

luis-henrix:
OK, I've modified the notify_inval_entry.c example (and test_notify_inval_entry test) so that it has an option for using increment epoch instead.

luis-henrix:
Hmm... did I screwed up the style?  I don't think so, right?  The examples have a style that's different from the kernel code, so I just followed that.

bsbernd:
Yeah, I started to just run 'clang-format -i <file>' on files before doing modifications (as preparation patch). Too much hassle to keep the different styles. Without checkpatch I don't see if someone uses whitespaces instead of tabs, with checkpatch it fails on these files.

bsbernd:
I can just ignore checkpatch here and will then run clang-format myself. A bit a pity that we loose simple `git annotate/blame`.

luis-henrix:
On Wed, Jun 18 2025, Bernd Schubert wrote:

>  /** Minor version number of this interface */
> -#define FUSE_KERNEL_MINOR_VERSION 42
> +#define FUSE_KERNEL_MINOR_VERSION 44
>
> Just notice this now, somehow I feel there is something missing. 7.43 added
> timeouts, but this is not in this patch? Can we just sync the entire
> fuse_kernel.h? I typically do as preparation patch.

Oh, that's right.  I noticed it too and forgot to mention it.  What happen
was that libfuse commit cd503168287e ("Update include/fuse_kernel.h to
version 7.43") was incomplete: it missed bumping the minor version.  But
all the other changes related with timeout were there.

Because that commit wasn't yet included in any libfuse release, I didn't
bother including a separate commit to fix that (i.e. a one-liner changing
version to 43).  But I can do that if that's what you're asking.

Cheers,
-- 
Luís

bsbernd:
Ah, thanks @luis-henrix for the explanation!

PR #1258, opened 2025-06-22
Clarify offset field is signed despite uint64_t type

izxl007:


bsbernd:
I think we can just add a comment here, that uint64_t offset is derived from linux kernel loff_t and is therefore signed.

static void _do_lseek(fuse_req_t req, const fuse_ino_t nodeid,
		      const void *op_in, const void *in_payload)
{
	(void)in_payload;
	const struct fuse_lseek_in *arg = op_in;
	struct fuse_file_info fi;

	memset(&fi, 0, sizeof(fi));
	fi.fh = arg->fh;

	if (req->se->op.lseek)
		req->se->op.lseek(req, nodeid, arg->offset, arg->whence, &fi);
	else
		fuse_reply_err(req, ENOSYS);
}

izxl007:
The relevant information has been updated, and this PR is ready to be merged.

izxl007:
> The relevant information has been updated, and this PR is ready to be merged.

@bsbernd

Issue #1075, opened 2024-12-07
funded update for macos?


r2evans:
> libfuse does not have any active, regular contributors or developers.  The current maintainer continues to apply pull requests and tries to make regular releases, but unfortunately has no capacity to do any development beyond addressing high-impact issues. 

I recognize the effort you've put into it so far. Often the "unable to actively develop" is based on funding issues, not necessarily interest. If that is the case for you, I would be interested in discussing _funding_ the work required to bring libfuse and sshfs up-to-date on macos (I'm on 15.1.1). I'm not familiar with the issues blocking the update, though I can imagine they are not trivial (otherwise it would likely have been done by now). If you're amenable to discussion, let me know.

If not, and if somebody has time (whether or not the primary maintainer here), it might be useful to enumerate some of the biggest hurdles to updating for macos. This would also be helpful to help quantify the time to develop, useful in estimating the cost of said development.

Thank you, I appreciate your efforts so far and do not want to diminish your decision to not actively maintain.

bsbernd:
Hi @r2evans ,

sorry for late reply, all weekend busy with something totally different.

Do you actually still need libfuse on OSX? There is https://osxfuse.github.io/?

Thanks,
Bernd

saxbophone:
> Do you actually still need libfuse on OSX? There is https://osxfuse.github.io/?
> 

It looks like macFUSE is stuck at implementing just the 2.9.7 API of libfuse only: https://github.com/macfuse/macfuse/issues/390

bsbernd:
Sorry, I actually meant https://www.fuse-t.org/

bsbernd:
@r2evans Besides that I don't have a single OSX system and any related hardware, time is the main factor. I'm not sure if funding would help me here - time is the issue again and then also with French bureaucracy (I'm tempted to write bureau_crazy_ - declaring tax and social contributions is a full time job...) funding might become a personal nightmare (unless you know someone who could help with declaring such things in France)...

saxbophone:
> Sorry, I actually meant https://www.fuse-t.org/

Ah cool. It looks like that one's a little behind the libfuse API too, but thanks for signposting me towards a hopefully more stable/maintained fork :)

r2evans:
It appears that fuse-t is on the same API version as macfuse, so I don't see (without benchmarking) an impetus to change. It claims to be faster, but it also has un-implemented components (at the same time it is a "drop-in replacement for macfuse" :-/ ...)

[...]

PR #1280, opened 2025-07-09
example/passthrough: support fspacectl()

CismonX:
FreeBSD 14 introduced a new system call, [`fspacectl(2)`](https://man.freebsd.org/cgi/man.cgi?fspacectl(2)).

Currently, it supports one operation mode, `SPACECTL_DEALLOC`, which is functionally equivalent to Linux `fallocate()` with `FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE` flags.

`fspacectl()` calls with `SPACECTL_DEALLOC` is supported on FUSE filesystems via `FUSE_FALLOCATE` with the aforementioned flags.

---

Followup: Ideally there should be tests for this patch, however, since fallocate is not tested at all, I think we could save that for later.

P.S. `fspacectl()` is available in FreeBSD userland via [`truncate(1)`](https://man.freebsd.org/cgi/man.cgi?truncate(1)) with `-d` option.

### DIFF
--- a/example/passthrough_helpers.h
+++ b/example/passthrough_helpers.h
@@ -23,6 +23,23 @@
  * SUCH DAMAGE
  */
 
+static inline int do_fallocate(int fd, int mode, off_t offset, off_t length)
+{
+#ifdef HAVE_FALLOCATE
+	if (fallocate(fd, mode, offset, length) == -1)
+		return -errno;
+	return 0;
+#else  // HAVE_FALLOCATE
+
+#ifdef HAVE_POSIX_FALLOCATE
+	if (mode == 0)
+		return -posix_fallocate(fd, offset, length);
+#endif
+
+	return -EOPNOTSUPP;
+#endif  // HAVE_FALLOCATE
+}
+
 /*
  * Creates files on the underlying file system in response to a FUSE_MKNOD
  * operation
--- a/example/passthrough_helpers.h
+++ b/example/passthrough_helpers.h
@@ -23,6 +23,20 @@
  * SUCH DAMAGE
  */
 
+#ifndef FUSE_EXAMPLE_PASSTHROUGH_HELPERS_H_
+#define FUSE_EXAMPLE_PASSTHROUGH_HELPERS_H_
+
+#include <errno.h>
+#include <fcntl.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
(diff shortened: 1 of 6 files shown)


CismonX:
Did a little refactor before the actual patch.

Wrapped the fallocate implementation into a helper function, so that it could be reused by multiple passthrough servers, without too many `#ifdef`s here and there.

CismonX:
@arrowd Perhaps you could review this patch? Thanks.

arrowd:
This looks OK to me, but I'm nowhere an expert in this area. Let's ask @asomers

example/passthrough_helpers.h
```diff
@@ -23,11 +23,55 @@
...
+	return 0;
+#else  // HAVE_FALLOCATE
+
+#ifdef HAVE_POSIX_FALLOCATE
+	if (mode == 0)
+		return -posix_fallocate(fd, offset, length);
+#endif
+
+#ifdef HAVE_FSPACECTL
+	// 0x3 == FALLOC_FL_KEEP_SIZE | FALLOC_FL_PUNCH_HOLE
```
asomers, line 54: Why don't you actually include the right headers so you can use these symbols, instead of hard coding their values here?

asomers (commented):
Looks pretty good to me.  I just have a few stylistic comments.

example/passthrough_helpers.h
```diff
@@ -23,11 +23,55 @@
  * SUCH DAMAGE
  */
 
+#ifndef FUSE_EXAMPLE_PASSTHROUGH_HELPERS_H_
+#define FUSE_EXAMPLE_PASSTHROUGH_HELPERS_H_
+
+#include <errno.h>
+#include <fcntl.h>
+#include <string.h>
+#include <sys/stat.h>
```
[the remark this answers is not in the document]
bsbernd: Is there a good reason to have sys headers first? To define something system specifically?

example/passthrough_helpers.h
bsbernd: Is there a fuse_kernel.h for BSD and we could use that? Maybe we should better rename fuse_kernel.h to fuse_linux_kernel.h and have it system specific? Because we already have an issue with OSX (https://github.com/libfuse/libfuse/issues/1075)

Issue #1286, opened 2025-07-15
unmounting /tmp/mnt failed: Invalid argument


asomers:
After PR #1133 I see this warning printed every time I unmount a fuse file system.  The problem is that after libfuse receives `FUSE_DESTROY` (indicating that some other process has unmounted the file system), it tries to unmount it again.  That behavior has long been present, but the errors were suppressed until #1133 .  I think the second unmount should be eliminated.  It should only be attempted if the libfuse main loop terminates due to a reason like SIGTERM or an internal error, not if it is closed by FUSE_DESTROY.

Steps to Reproduce
----------------------------

```
> example/hello_ll -d -f /tmp/mnt
FUSE library version: 3.18.0-rc0
dev unique: 1, opcode: INIT (26), nodeid: 0, insize: 56, pid: 15016
INIT: 7.33
flags=0x01030033
max_readahead=0x00010000
   INIT: 7.44
   flags=0x00000020
   max_readahead=0x00010000
   max_write=0x00020000
   max_background=0
   congestion_threshold=0
   time_gran=1
   unique: 1, success, outsize: 80


*** In another terminal do "umount /tmp/mnt" ***


dev unique: 2, opcode: DESTROY (38), nodeid: 0, insize: 40, pid: 15017
   unique: 2, success, outsize: 16
fuse: session exited, terminating workers
unmounting /tmp/mnt failed: Invalid argument⏎   
```
Environment
-------------------
FreeBSD 15.0-CURRENT, 14-STABLE, and 14.3-RELEASE
libfuse 3.17.1 and the current master branch

bsbernd:
Thanks @asomers ! Personally I don't like that auto-unmount too much and I was aware of the double umount - took a few min and created a patch. Although there are more issues and auto-unmount from fusermount daemon mode is requested.

gcsideal:
Is this affect releases 3.17.x? I see you added several commits with this fix, but I get "this commit does not belong to any branch on this repository, and may belong to a fork outside of the repository" on those.

asomers:
> Is this affect releases 3.17.x? I see you added several commits with this fix, but I get "this commit does not belong to any branch on this repository, and may belong to a fork outside of the repository" on those.

Yes.  3.17.x is affected.  3.16.2 and earlier will not print the warning message, but they will still attempt to double-unmount.

bsbernd:
Note that the warning message for the double umount is only part of `lib/mount_bsd.c`

[...]

PR #1287, opened 2025-07-15
Avoid double unmount on FUSE_DESTROY

bsbernd:
This is a long standing issue, a system could have unmounted /path/to/mnt and then fuse-client/kernel would send FUSE_DESTROY, which would then again try a umount.

Given that FUSE_DESTROY is async, that umount might arrive any time later and might possibly unmount a wrong mount point. A warning as in issue #1286 is just minor to that.

Code wise this uses atomics to free the char *, as FUSE_DESTROY might race with a signal and a double free might be possible without proctection. A lock might run into the same issue, if the signal would arrive at the wrong time a double lock would be possible.

Additionally, fuse_session_mount() is updated, to first duplicatate the pointer and to then do the kernel mount - reverting the kernel mount in case of strdup() failure is much harder.

Closes: https://github.com/libfuse/libfuse/issues/1286

### DIFF
--- a/lib/fuse_i.h
+++ b/lib/fuse_i.h
@@ -18,6 +18,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <errno.h>
+#include <stdatomic.h>
 
 #define MIN(a, b) \
 ({									\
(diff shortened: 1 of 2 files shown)


bsbernd:
@asomers if you should have a bit time, would you mind to review?

asomers (approved):
This LGTM and it tests good too.  Thanks for the fast fix!

bsbernd:
Thanks a lot for your review @asomers !

PR #1291, opened 2025-07-16
memfs_ll: fix deadlock in truncate operation

lonuxli:
Remove redundant mutex lock acquisition in the truncate() method to prevent deadlock. The issue occurs when memfs_setattr() already holds the mutex lock and then calls truncate(), which attempts to acquire the same lock again.

### DIFF
--- a/example/memfs_ll.cc
+++ b/example/memfs_ll.cc
@@ -199,7 +199,6 @@ class Inode {
 
 	void truncate(off_t size)
 	{
-		std::lock_guard<std::mutex> lock(mutex);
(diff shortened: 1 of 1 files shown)


bsbernd (approved):
Thank you! Not well tested on my side yet - never found the time after the initial commits.

izxl007:
哈哈，是桃子湖的LiLong吗？

lonuxli:
> 哈哈，是桃子湖的LiLong吗？

不是, 白马湖的

PR #1297, opened 2025-07-21
Passthrough hp syslog redirection

bsbernd:
Just a draft, I'm thinking to add this to fuse_daemonize().


bsbernd:
Will be added in the new damonize api.

PR #1299, opened 2025-07-23
fix splice invalid argument

eddyli1989:


bsbernd:
@eddyli1989 Sorry for my late response, I like the approach in general,. but tests fails. Lots of other work queued, I can only take a deeper look on Wednesday.

bsbernd:
I think using a single worker at startup time is better: https://github.com/libfuse/libfuse/pull/1326

PR #1301, opened 2025-07-24
mount_bsd.c: Actually report mount failures back to caller

arrowd:


arrowd:
Getting @asomers  on the hook.

PR #1304, opened 2025-07-30
Avoid using pthread_cancel for thread termination

SC404:
we are using libfuse in OpenHarmony,so we would like to avoid using pthread_cancel for thread termination. The solution favors cooperative thread termination using exit flags over forced cancellation, leading to more predictable system behavior.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -4969,7 +4969,7 @@ void fuse_stop_cleanup_thread(struct fuse *f)
 {
 	if (lru_enabled(f)) {
 		pthread_mutex_lock(&f->lock);
-		pthread_cancel(f->prune_thread);
+		// pthread_cancel(f->prune_thread);
 		pthread_mutex_unlock(&f->lock);
 		pthread_join(f->prune_thread, NULL);
 	}
--- a/lib/fuse_loop_mt.c
+++ b/lib/fuse_loop_mt.c
@@ -138,9 +138,9 @@ static void *fuse_do_work(void *data)
 		int isforget = 0;
 		int res;
 
-		pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, NULL);
+		// pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, NULL);
 		res = fuse_session_receive_buf_internal(se, &w->fbuf, w->ch);
-		pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, NULL);
+		// pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, NULL);
 		if (res == -EINTR)
 			continue;
 		if (res <= 0) {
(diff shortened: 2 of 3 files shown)


bsbernd:
I don't think we can easily do that. Issue is that waiting for fuse requests goes into a blocking read() and these need to be woken up. Example from musl libc

```
int pthread_cancel(pthread_t t)
{
    static int init;
    if (!init) {
        init_cancellation();
        init = 1;
    }
    a_store(&t->cancel, 1);
    if (t == pthread_self()) {
        if (t->canceldisable == PTHREAD_CANCEL_ENABLE && t->cancelasync)
            pthread_exit(PTHREAD_CANCELED);
        return 0;
    }
    return pthread_kill(t, SIGCANCEL);
}
```

I.e., the interesting part here is the SIGCANEL, which will wake up threads waiting in that blocking read.

bsbernd:
I'm closing this, as just removing pthread_cancel cannot be the solution.

spectral54:
We just need some non-BSD-restart-semantics signal to wake up the read, right? Is there another signal we could use? I'd of course normally reach for SIGUSR1 or SIGUSR2, but worry that we might not set the sigaction late enough, so maybe we should go for a more esoteric signal, even if it's not part of a standard (like SIGIO or something "weird" like that)?

pthread_cancel is very scary and we've seen it cause deadlocks in our system (I think due to non-signal-safe actions being taken in other signal handlers, and no, we can't really fix that, unfortunately)

bsbernd:
Maybe we can also manually send SIGCANCEL. I don't know which signals wake a blocking wait and that also might be OS dependent.  I guess SIGCANCEL does for al OSs

spectral54:
I thought about using sigcancel, and it turns out that there's actually a sigcancel signal handler, which does all of the "this thread is being torn down immediately" stuff we need to avoid :(

Someone reminded me of this race condition a few seconds after I sent my previous comment, so just sending a signal isn't sufficient:

```
if (!fuse_session_exited())
  /* signal arrives between that check and this next line: it gets handled and then read() is called and we have a hang */
  read()
if (fuse_session_exited()) exit_gracefully();
```

I think we'd need to switch to some other form of I/O such as `poll()` + `read()` (which won't block long since if we get past the `poll()` we have stuff to read and so it won't block). We then move the issue to `poll()`, which could be waiting on multiple fds, and our "cancel that" signal is "set the fuse session exited bit, write to an fd to wake up all the `poll`s". Bonus: we only have to do it once instead of once per thread :)

bsbernd:
I wouldn't like to generalize poll + read. With io-uring (which BSD does not have yet), maybe - everything performance critical is then moved to io-uring. But if io-uring is not used, poll + read means two syscalls instead of one. IO-uring does not have that read wake up issue, as I have set up a poll on the teardown eventfd in the rings.

bsbernd:
So if you want to work on removing pthread_cancel and also switch to poll+read in general - it all has to be optional.

PR #1312, opened 2025-08-12
fuse_loop_mt.c: fix close-on-exec flag on clone fd

### LINKED ISSUE #1310: Cloned FUSE device channel is not set to close-on-exec if O_CLOEXEC is not available
This branch of this `#ifndef` where `O_CLOEXEC` is undefined is never taken,
https://github.com/libfuse/libfuse/blob/ba8a3a420e0421561eb6ff6caf17588e748ab164/lib/fuse_loop_mt.c#L274-L276
because a few lines above `O_CLOEXEC` is defined if it previously wasn't.
https://github.com/libfuse/libfuse/blob/ba8a3a420e0421561eb6ff6caf17588e748ab164/lib/fuse_loop_mt.c#L265-L267

CismonX:
Closes: #1310

### DIFF
--- a/lib/fuse_loop_mt.c
+++ b/lib/fuse_loop_mt.c
@@ -271,9 +271,15 @@ static int fuse_clone_chan_fd_default(struct fuse_session *se)
 			strerror(errno));
 		return -1;
 	}
-#ifndef O_CLOEXEC
-	fcntl(clonefd, F_SETFD, FD_CLOEXEC);
-#endif
+	if (!O_CLOEXEC) {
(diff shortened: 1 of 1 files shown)


bsbernd:
@CismonX Thank you, LGTM. Could you add your Signed-off? And this probably should be part of 3.17.4 (just wanted to release it).

CismonX:
@bsbernd The original commit already has a Signed-off.
Nevertheless, I updated the commit message so that checkpatch won't fail.

bsbernd (approved):
LGTM, thank you.

PR #1317, opened 2025-08-14
fuse_uring: fix race by locking thread counters and add usleep to wait loop

lixianming19951001:


bsbernd:
Thanks @lixianming19951001 Let's do it a bit differently, see here https://github.com/libfuse/libfuse/pull/1318

PR #1320, opened 2025-08-19
tests: move struct size assertions into a test

CismonX:
These checks are meant for libfuse maintainers only, and should not be exposed to users.


bsbernd:
Thanks a lot @CismonX !

PR #1330, opened 2025-09-09
fuse_lowlevel.c: allow passing ops as NULL

In-line:
During testing of LACT using various sanitizers I uncovered issue with libfuse.so leaking memory and reading uninitialized memory. While this was due to ABI misusage on the side of `fuser` (Rust wrapper for libfuse.so) behaviour of `fuse_session_new` is very unintuitive, it makes sense to handle it.

Also this code has made it's way into production applications, which further makes sense to fix it in all sides.

Related
https://github.com/ilya-zlobintsev/LACT/pull/740
https://github.com/cberner/fuser/pull/390
https://github.com/Alogani/easy_fuser/issues/60

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -4051,6 +4051,14 @@ fuse_session_new_versioned(struct fuse_args *args,
 	int err;
 	struct fuse_session *se;
 	struct mount_opts *mo;
+	struct fuse_lowlevel_ops null_ops = { 0 };
+
+	// Empty operations list may is necessary to create a no-op fuse session for mounting
+	if (op == NULL) {
+		fuse_log(FUSE_LOG_ERR, "fuse: warning: empty op list, falling back to default\n");
+		op = &null_ops;
+		op_size = sizeof(null_ops);
+	}
 
 	if (sizeof(struct fuse_lowlevel_ops) < op_size) {
 		fuse_log(FUSE_LOG_ERR, "fuse: warning: library too old, some operations may not work\n");
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -2117,6 +2117,8 @@ fuse_session_new_versioned(struct fuse_args *args,
  * If not all options are known, an error message is written to stderr
  * and the function returns NULL.
  *
+ * To create a no-op session just for mounting pass op as NULL.
+ *
  * Option parsing skips argv[0], which is assumed to contain the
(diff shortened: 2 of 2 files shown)


lib/fuse_lowlevel.c
```diff
@@ -4051,6 +4051,13 @@ fuse_session_new_versioned(struct fuse_args *args,
 	int err;
 	struct fuse_session *se;
 	struct mount_opts *mo;
+	struct fuse_lowlevel_ops fallback_ops = { 0 };
+
+	if(op == NULL) {
```
bsbernd, line 4056: Nit: indentation: if (op)

lib/fuse_lowlevel.c
```diff
@@ -4051,6 +4051,13 @@ fuse_session_new_versioned(struct fuse_args *args,
...
+		fuse_log(FUSE_LOG_ERR, "fuse: warning: empty op list, falling back to default\n");
+		op = &fallback_ops;
+		op_size = sizeof(fallback_ops);
+	}
```
bsbernd, line 4060: To be honest, I don't see what the purpose is - why not return NULL? What is the purpose of a "null_op", if it won't do anything?

bsbernd (commented):
@In-line you basically added in the commit message the diagnosis, but not the reason  - missing is why external tools are mis-using the ABI.

In-line:
@bsbernd please feel free to suggest the exact message, if this one isn't okay.

bsbernd:
@In-line I need to read through the fuser code to be sure - don't have time for this. The current commit messages suggests to that fuse_session_new_versioned() should return an error when op is NULL, but I'm rather sure that would break fuser. 
If we let through the commit as it is, I'm sure that either me or someone else would come up in a few years with a commit to change it - which is why we need a code comment explain why a no-op is needed. 

`// Empty operations list may is necessary to create a no-op fuse session for mounting`

-> Please use linux style (checkpatch.pl will complain anyway. And needs to explain that external libfuse users need this and a bit more detailed.


I had quickly check fuser yesterday or the day before, and I think the no-op session is used for mounting. But I'm still not sure about the exact interaction with libfuse here.
I'm not sure if I will have time before Saturday to go through fuser code - maybe @cberner could fill in the exact commit message? If not I will do ASAP.

bsbernd:
@In-line Could you please also add your Signed-off-by line?

bsbernd:
@In-line Updated PR is here: https://github.com/libfuse/libfuse/pull/1335

PR #1331, opened 2025-09-12
libfuse: fix COPY_FILE_RANGE interface

bsbernd:
The FUSE protocol uses struct fuse_write_out to convey the return value of copy_file_range, which is restricted to uint32_t.  But the COPY_FILE_RANGE interface supports a 64-bit size copies.

Currently the number of bytes copied is silently truncated to 32-bit, which is unfortunate at best.

Implement the COPY_FILE_RANGE_64 interface which is identical to the old one, except the number of bytes copied is returned in a 64-bit value.

The library interface remains the same.

If the kernel does not support the new interface or the server is running as a 32-bit process, limit the copy size to size to UINT_MAX - 4096.

Reported-by: Florian Weimer <fweimer@redhat.com>
Closes: https://lore.kernel.org/all/lhuh5ynl8z5.fsf@oldenburg.str.redhat.com/

### DIFF
--- a/include/fuse_kernel.h
+++ b/include/fuse_kernel.h
@@ -235,6 +235,10 @@
  *
  *  7.44
  *  - add FUSE_NOTIFY_INC_EPOCH
+ *
+ *  7.45
+ *  - add FUSE_COPY_FILE_RANGE_64
+ *  - add struct fuse_copy_file_range_out
  */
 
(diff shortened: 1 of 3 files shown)


bsbernd:
@szmi I slightly updated it, I think the code is easier to read with a new bit.

PR #1347, opened 2025-10-12
Use `listmount` and `statmount` to count filesystems

blinsay:
### background

I've been fuzz-testing a FUSE implementation by running a loop that creates and mounts a new FUSE, runs some operations against it, and unmounts. The fuzz test does this repeatedly from a single thread, and should never be mounting more than one FUSE at a time.

Despite that, occasionally I see the following error, with `max_mount` set to the default value of 1000:

```
fusermount3: too many FUSE filesystems mounted; mount_max=N can be set in /etc/fuse.conf
```

After manually digging in `/proc` it seemed like the fuzzing wasn't leaving unmounted filesystems around, so I copied [`count_fuse_fs`](https://github.com/libfuse/libfuse/blob/564d79e4d3132924d33f0a0848b9d54c84b42a17/util/fusermount.c#L569-L587) and wrote some code based on [the new `statmount` and `listmount`](https://lwn.net/Articles/950569/) APIs see if I could get a better view of what was going on:

<details>

<summary>count_fuse_fs.c</summary>

```c
#include <stdio.h>
#include <mntent.h>
#include <syscall.h>
#include <linux/mount.h>
#include <unistd.h>
#include <errno.h>
#include <stdint.h>
#include <string.h>


static int count_fuse_fs_listmount(void) {
	struct mnt_id_req req = {
		.size = sizeof(struct mnt_id_req),
	      	.mnt_id = LSMT_ROOT,
	};
	uint64_t mnt_ids[128];
	unsigned char smbuf[1024];
	int count = 0;

	int n = syscall(SYS_listmount, &req, &mnt_ids, 128, 0);
	if (n == -1) {
		return errno;
	}

	for (int i = 0; i < n; i++) {
		req.mnt_id = mnt_ids[i];
		req.param = STATMOUNT_SB_BASIC | STATMOUNT_FS_TYPE | STATMOUNT_MNT_ROOT | STATMOUNT_MNT_POINT;
		int ret = syscall(SYS_statmount, &req, &smbuf, 1024, 0);
		if (ret) {
			if (errno != ENOENT) {
				perror("statmount failed");
			}
			continue;
		}


		struct statmount *sm = (struct statmount *)smbuf;
		if ((sm->mask & STATMOUNT_FS_TYPE) && strcmp(&sm->str[sm->fs_type], "fuse") == 0) {
			count++;
		}
	}	

	return count;
}

static int count_fuse_fs(const char* mtab) {
	struct mntent *entp;
	int count = 0;
	FILE *fp = setmntent(mtab, "r");
	if (fp == NULL) {
		fprintf(stderr, "failed to open %s: %s\n", mtab, strerror(errno));
		return -1;
	}
	while ((entp = getmntent(fp)) != NULL) {
		if (strcmp(entp->mnt_type, "fuse") == 0 || strncmp(entp->mnt_type, "fuse.", 5) == 0) {
			count ++;
		}
	}
	endmntent(fp);
	return count;
}

int main(int argc, char *argv[]) {
	int proc_count = count_fuse_fs("/proc/mounts");
	int statmount_count = count_fuse_fs_listmount();

	if ((proc_count > 0) || (statmount_count > 0)) {
		printf("/proc/mounts=%d listmount=%d\n", proc_count, statmount_count);
	}
}
``` 

</details>

In a few hours of fuzzing, I haven't seen `listmount` ever report a count of more than one FUSE, while every few minutes I see counting lines in `/proc/mounts` report that hundreds of FUSEs exist. 

```
$ while true; do ./count_fuse_fs; sleep 0.1; done/proc/mounts=0 listmount=1
...
/proc/mounts=0 listmount=1
/proc/mounts=1 listmount=1
/proc/mounts=916 listmount=0
/proc/mounts=1 listmount=1
...
/proc/mounts=1 listmount=1
/proc/mounts=1107 listmount=0
/proc/mounts=1 listmount=0
```

There's a delay between the two counts being called so I expect to occasionally see different results, but if the fuzz test was actually leaving a few hundred unmounted FUSEs around I would expect `listmount` to report that at some point. This really does seem like [the classic problem with parsing `/proc`](https://stackoverflow.com/questions/5713451/is-it-safe-to-parse-a-proc-file).

### the change

I took a stab at using `listmount` and `statmount` to count filesystems when they're available. There are no libc wrappers for these functions, so they're called directly through `syscall`. I'm sure there's plenty else to fix here, especially the commit message, but before we get into details I want to make sure folks would be comfortable using `listmount` and `statmount` without libc wrappers.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -99,6 +99,25 @@ special_funcs = {
             unsigned int flags = CLOSE_RANGE_UNSHARE;
             return close_range(3, ~0U, flags);
         }
+    ''',
+    'listmount': '''
+	#include <syscall.h>
+	#include <linux/mount.h>
+	#include <unistd.h>
+	#include <stdint.h>
+
+	int main(int argc, char *argv[]) {
+	  struct mnt_id_req req = {
+	    .size = sizeof(struct mnt_id_req),
+		  .mnt_id = LSMT_ROOT,
+	  };
+	  uint64_t mnt_ids[1];
+
+	  int n = syscall(SYS_listmount, &req, &mnt_ids, 1, 0);
+	  if (n == -1) {
+	    return -1;
+	  }
+	}
     '''
 }
 
--- a/util/fusermount.c
+++ b/util/fusermount.c
@@ -40,6 +40,12 @@
 #include <linux/close_range.h>
 #endif
 
+#if defined HAVE_LISTMOUNT
+#include <linux/mount.h>
+#include <syscall.h>
+#include <stdint.h>
+#endif
+
 #define FUSE_COMMFD_ENV		"_FUSE_COMMFD"
 #define FUSE_KERN_DEVICE_ENV	"FUSE_KERN_DEVICE"
 
@@ -566,7 +572,7 @@ static int unmount_fuse(const char *mnt, int quiet, int lazy)
 	return res;
 }
 
-static int count_fuse_fs(void)
+static int count_fuse_fs_mtab(void)
 {
 	struct mntent *entp;
 	int count = 0;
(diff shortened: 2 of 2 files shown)


bsbernd:
@blinsay that is an interesting finding! Could you add a fall back to the 'old' count_fuse_fs(), if the new version fails? Like `count_fuse_fs_ls_mnt` and `count_fuse_fs_mtab`, and then `count_fuse_fs_ls_mnt` calls into `count_fuse_fs_ls_mnt` and if that fails for whatever reason, it falls back to `count_fuse_fs_mtab`?

blinsay:
> Could you add a fall back to the 'old' count_fuse_fs(), if the new version fails?

Sure can. What's the rationale here? I get the sense that if the syscall is available, it shouldn't be any less reliable than reading mtab.

bsbernd:
Well, what is if libfuse was build on libc that supports listmount, but the booted kernel then doesn't have the syscall? Without fallback there also needs to be handling if `mnt_ids[n], size_t n` is just too small.

blinsay:
> Well, what is if libfuse was build on libc that supports listmount, but the booted kernel then doesn't have the syscall? 

Ah sure, of course. 

> Without fallback there also needs to be handling if mnt_ids[n], size_t n is just too small.

`mnt_req_id` has a `param` field that lets you continue, so I should be able to deal with that without a failure.

I may not have time for this for the next couple days, but I'll be back with those changes. Thanks for the feedback.

blinsay:
@bsbernd updated! We should now handle any number of mounts, and falls back to reading `mtab` if `listmount` ever fails. It didn't seem like it was worth checking for `ENOSYS` specifically, so the fallback is unconditional. I also picked `SMBUF_SIZE` and `MNT_ID_LEN` pretty arbitrarily - my system has 65 mounts, so I rounded up hoping that this would mean we mostly do a single `listmount` call.

bsbernd (dismissed):
Thank a lot, LGTM!

blinsay:
@bsbernd is there an easy way to run `./checkpatch.pl` locally? I don't see it documented anywhere.

bsbernd:
@blinsay I had created a gist for myself `https://gist.github.com/bsbernd/e48f75de0ea886ebcf09689400e39d5c`

`/checkpatch.pl --max-line-length=100 --no-tree --ignore MAINTAINERS,SPDX_LICENSE_TAG,COMMIT_MESSAGE,FILE_PATH_CHANGES,EMAIL_SUBJECT,AVOID_EXTERNS,GIT_COMMIT_ID,ENOSYS_SYSCALL,ENOSYS,FROM_SIGN_OFF_MISMATCH,QUOTED_COMMIT_ID,PREFER_ATTRIBUTE_ALWAYS_UNUSED,PREFER_DEFINED_ATTRIBUTE_MACRO -g HEAD`

blinsay:
@bsbernd all set, not sure what to do about the signed-off-by line though. Thanks for your help!

bsbernd:
@blinsay Could you please also update the commit message with a description why this is needed - kind of what you did for the PR (I personally don't fill PR requests, but only commit messages). And then

Signed-off-by: My Name <my.name@example.com>

blinsay:
@bsbernd I think we're good, I'm passing `./checkpatch.pl` locally. Let me know if I can help with anything else.

bsbernd:
Thanks a lot @blinsay !

PR #1351, opened 2025-10-21
io-uring: register ring fd

joannekoong:
Register the ring fd which allows us to avoid fdget/fdput for every io_uring_enter() syscall.

### DIFF
--- a/lib/fuse_uring.c
+++ b/lib/fuse_uring.c
@@ -370,6 +370,8 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
 		return rc;
 	}
 
+	io_uring_register_ring_fd(ring);
+
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


bsbernd:
@joannekoong Thank you. I had seen and considered it before, but this note makes me hesitate 

```
NOTES
       When the ring descriptor is registered, it is stored internally in the struct io_uring structure.
       For  applications that share a ring between threads, for example having one thread do submits and
       another reap events, then this optimization cannot be used as each thread may  have  a  different
       index for the registered ring fd.
```

I.e. should this an option that can be set in ->init()? Something like `single_ring_issuer`?

joannekoong:
Hi @bsbernd, sorry for the late reply. Just saw that you commented on this. Thanks for looking at this. 

This is okay to use because libfuse doesn't share rings between threads, even for configurable queues. I ddon't think this needs to be set in the init reply, the fuse kernel is completely unconcerned with this.

bsbernd:
@joannekoong I still don't think it is correct by default.
 We don't have a libfuse/example for it, but fuse_req_t can be passed to another thread and the current ring thread continues to handle more CQE / fuse_req. The thread that has taken over the request might handle it and reply. I.e. each fuse application needs to decide itself if it is safe to set that.
 I'm also going to add an example for thread offload. I have this discussion the 2nd time this week - the other time in review or compound requests :)

joannekoong:
Thread offloading is a good point, I'll add this behind a config flag. I will update this PR! :)

bsbernd:
Let's close this for now, it is trivial to add, but first needs an option for single-issuer.

PR #1357, opened 2025-10-29
fuse_lowlevel.h: remove incorrect documentation

z-s-e:
It seems fuse_reply_open() cannot actually return -ENOENT under recoverable conditions, so remove the corresponding paragraph claiming otherwise.

For a discussion on this see
https://github.com/libfuse/libfuse/discussions/1262


bsbernd:
Thanks @z-s-e ! I squashed the patches into one commit.

PR #1367, opened 2025-11-20
Add support for synchronous FUSE_INIT 

bsbernd:


bsbernd:
@szmi could you take a look? I think it works for privileged mounts. Will look into fusermount changes for unprivileged tomorrow.

szmi:
I'm not sure about limiting to a single request (INIT).  Long time ago there was an issue with selinux wanting to perform getxattr during super block initialization.  I'm not sure how that was worked around, but allowing more than one request would be the proper way to address this and similar cases.

But this leads to difficulties.  Now we don't know which request is related to mount and which is not.  I have some ideas which would further complicate the userspace/kernel interface, so I'm reluctant to go that way.

Thoughts?

bsbernd:
Thanks for review @szmi I was quite busy over the weekend, because sync INIT causes issues with io-uring - we cannot start the ring threads from the INIT handler, as as these get destroyed after `fuse_daemonize`. 

selinux is exactly why I'm looking into this - it currently does not work. I didn't test with sync INIT yet, though, we just have a backtrace where mounting blocks because of getxattr during mount.

I will look into letting more requests through with the initial thread. The other issue with io-uring is that we cannot block the mount then until io-uring initialization is complete. Main issue there is a lock order issue - we cannot hold the bg lock while holding the queue lock. Typical lock order is queue-lock, then bg lock. But on switching from /dev/fuse rw to io-uring lock order gets reversed.

I guess this needs a larger discussion - we should probably continue on fsdevel.

bsbernd:
@szmi eventually got the time to update it (had started the update during x-mas vacation and then didn't come to it until now). See commit description the 1st commit please, there is a bit preparation work needed.

szmi:
There's one situation that this doesn't handle well: if the fs crashes while processing the INIT request, the mount syscall will hang, because it's holding a ref to the device file.

Fixing this requires:

 - libfuse: switching to the new mount API
 - kernel: in fuse_parse_param() storing a reference to the fuse_conn instead of the fd
 - spawning a new process for mount, not just a new thread
 - after FSCONFIG_SET_FD and before FSCONFIG_CMD_CREATE closing the devfd

It's a bit more complicated, but may be worth it, otherwise any failure or crash during the initialization will result in a hang (resolvable only with with echo 1 > ../abort), which is a regression compared to the async mount case.

bsbernd:
Thanks for looking into it @szmi . How about creating a mount watchdog process that would do the conn abort if the parent is gone?

szmi:
It's not sufficient to watch for parent death, since the parent is still there, just hung.

Timing out the mount would work, but the issue with timeouts is that there's never a "good" value.  If too short it will break real cases.  If too long it's useless.

I'd prefer the good old "abort the conn when the dev fd is closed", which works pretty well, just needs tweaking for the sake of sync init.

szmi:
Maybe it's sufficient to do unshare(CLONE_FILES) in the mount thread before FSCONFIG_CMD_CREATE so that at least the other threads can be killed, which would result in the devfd being released and the mount thread able to continue and be killed also.

bsbernd:
Thank you @szmi Means we first need to switch to the new mount API.

bsbernd:
@szmi The part I don't understand is how unshare(CLONE_FILES) would help. Doesn't that create a reference on devfd and destruction of the other threads wouldn't help? This reminds me a bit on the early io-uring patches where I had introduced a special io-uring CLONE ioctl and detection if these are fds of the io-uring threads and then just ignored these `fuse_dev_release()`

hbirth:
LGTM

lib/mount.c
```diff
@@ -524,50 +581,87 @@ static int fuse_mount_sys(const char *mnt, struct mount_opts *mo,
...
+ */
+int fuse_kern_do_mount(const char *mnt, struct mount_opts *mo,
+		       const char *mnt_opts)
+{
+	char *source = NULL;
+	char *type = NULL;
+	int res;
+	const char *devname = fuse_mnt_get_devname();
+	res = -ENOMEM;
+	source = fuse_mnt_build_source(mo->fsname, mo->subtype, devname);
```
kchen-ddn, line 658: The 'source' and 'type' may be changed below after mount failed with ENODEV, we'd better free  them and then call asprintf again for their modification.

example/passthrough_hp.cc
```diff
@@ -243,6 +245,9 @@ static void sfs_init(void *userdata, fuse_conn_info *conn)
 
 	/* Try a large IO by default */
 	conn->max_write = 4 * 1024 * 1024;
+
+	/* required here for async init */
+	fuse_daemonize_success();
```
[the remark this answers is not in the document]
bsbernd: @kchen-ddn Removed at all in ->init() and done automatically at the end of `_do_init`

lib/mount_fsmount.c
```diff
@@ -0,0 +1,477 @@
...
+
+	errno = err;
+}
+
+/*
+ * Apply VFS superblock flags to the filesystem context.
+ * Only handles flags that are filesystem parameters (ro, sync, dirsync).
+ * Mount attributes (nosuid, nodev, etc.) are handled separately via fsmount().
+ */
+static int set_ms_flags(int fsfd, unsigned long ms_flags)
```
kchen-ddn: This function never returns success.
The ms_flags can have other flags set, e.g. MS_NOSUID and MS_NODEV, etc.
In that case the for loop will iterate to the last fme of strflags, i.e. {0,0}.
Then 'fuse: set (null) option: Invalid argument' will be printed and then error returned.

Even if the ms_flags only contains MS_SYNCHRONOUS, MS_DIRSYNC and MS_LAZYTIME, after the for loop, it will return EOPNOTSUPP.

Should we only handle MS_SYNCHRONOUS, MS_DIRSYNC and MS_LAZYTIME in this function and leave the other flags to ms_flags_to_mount_attrs()?

lib/mount.c
kchen-ddn: I mean, here we do strcpy and sprintf to 'source' or 'type'. Should we free them and asprintf() again?

lib/mount.c
bsbernd: Good catch @kchen-ddn 
In the "Refactor mount code / move common functions to mount_util.c" patch it is still save - still uses malloc. Needs to be fixed in the commit that switches to asprintf.

bsbernd: Actually it was safe, because subtype was just removed - the allocated array by asprintf() was guaranteed to be larger. But I still added the free and new asprintf() to ensure it cannot be broken in the future.

lib/mount.c
bsbernd: I did that now, please check again, but it was actually safe, because subtype removal needs less space.

lib/mount_fsmount.c
bsbernd: @kchen-ddn Thank you, that would have been a real issue. Given that tests did not find it, I need to add a test for it. Should be handled now, but I don't like the private maps in `mount_fsmount.c` yet. I think I'm going to add it to `const struct mount_flags mount_flags[]` in mount_utils.c

kchen-ddn: Yes, the test could pass because it failed back to old API. I only noticed it because I was trying to do SELinux mount, which eventually hung at getxattr.

lib/mount.c
kchen-ddn: Yes, thanks!

lib/mount_fsmount.c
```diff
@@ -0,0 +1,488 @@
...
+		ret = fsconfig(fsfd, FSCONFIG_SET_FLAG, fme->string, NULL,
+			       0);
+		if (ret) {
+			int save_errno = errno;
+
+			fprintf(stderr, "fuse: set %s option: %s\n",
+				fme->string, strerror(save_errno));
+			log_fsconfig_kmsg(fsfd);
+
+			return -save_errno;
```
kchen-ddn, line 138: This line we return an errno, but the return value should be the not handled flags.

kchen-ddn (changes requested):

lib/mount_fsmount.c
bsbernd: Thank you, updated. There was another issue, the loop did not check for map exhaustion - out of bound access.

PR #1368, opened 2025-11-22
example/passthrough_hp: fix race between forget_one and do_lookup

lonuxli:
During fsstress stress testing using passthrough_hp as the backend, the backend process crashes. The root cause is that when forget_one() and do_lookup() concurrently process the same inode, do_lookup may return either an invalid inode or a different inode reusing the same memory address.

This can lead to abnormalities in the inode nlookup count. Since the value of inode.nlookup determines the inode's lifecycle, and considering the locking order requirements between the inode lock and fs lock, using the inode lock alone does not resolve the issue effectively. Therefore, the fs lock is now used to protect inode.nlookup.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -78,6 +78,7 @@
 #include "cxxopts.hpp"
 #include <mutex>
 #include <syslog.h>
+#include <atomic>
 
 #include "passthrough_helpers.h"
 
@@ -130,7 +131,7 @@ struct Inode {
 	int generation{ 0 };
 	int backing_id{ 0 };
 	uint64_t nopen{ 0 };
-	uint64_t nlookup{ 0 };
+	std::atomic<uint64_t> nlookup{ 0 };
 	std::mutex m;
 
 	// Delete copy constructor and assignments. We could implement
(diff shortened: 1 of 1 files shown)


bsbernd:
@lonuxli Thank you, instead of extending the global lock, could we make nlookup atomic? I'm currently working another issue and only have a minute for this here. 

I wonder if we could improve it if `nlookup` would be an atomic. Quite similar to your current patch, except in forget_one(), when it would keep the lock as it is, but then re-check under the lock if the lookup count is still 0.

lonuxli:
Ok, I will fix it by use atomic.

bsbernd:
Thanks @lonuxli !

bsbernd:
Hmm, interesting, this leads to a bug

https://github.com/libfuse/libfuse/actions/runs/19633500073/job/56218545538#step:6:3368

lonuxli:
> Hmm, interesting, this leads to a bug
> 
> https://github.com/libfuse/libfuse/actions/runs/19633500073/job/56218545538#step:6:3368

I apologize that my patch might have introduced a bug. I spent quite some time reviewing the code but couldn't identify the issue, and I haven't been able to reproduce it in my testing. From the debug log, it seems the kernel function fuse_backing_open() is returning -EPERM, but I'm not clear why this is happening. I'll continue investigating to find the root cause.

@bsbernd Do you have any suggestions?

Thanks!

lonuxli:
> Hmm, interesting, this leads to a bug
> 
> https://github.com/libfuse/libfuse/actions/runs/19633500073/job/56218545538#step:6:3368

This appears to be caused by my mistaken removal of the inode lock in forget_one. When multiple forget_one calls run concurrently, it can lead to UAF (use-after-free) issues.

I will fix it as soon as possibe.

@bsbernd  Thanks.

PR #1371, opened 2025-11-25
example/passthrough_hp: fix race in forget_one()

lonuxli:
When multiple threads concurrently call forget_one() on the same inode, a use-after-free memory issue can occur.
```
 forget_one()               forget_one()
 ----------------           ---------------
           <inode.nlookup == 2>
 inode.nlookup -= 1
                            inode.nlookup -= 1
           <inode.nlookup == 0>
 if (!inode.nlookup)
   fs.inodes.erase()
                            if (!inode.nlookup) {} //UAF
```
Fix it by restoring the inode lock protection in forget_one().


bsbernd:
Thank you @lonuxli ! I should have noticed in previous review.

PR #1377, opened 2025-11-26
lib/fuse.c: fix deallocation in fuse_session_loop_remember

### LINKED ISSUE #1376: Crash (invalid pointer passed to free) with `-o remember=1`
**Describe the bug**
FUSE server using the high-level API crashes on `free(fbuf.mem);` in `fuse_session_loop_remember`. Since commit 752b59ac0876, internally allocated buffer must be freed with `fuse_buf_free`, not plain `free`.

**To Reproduce**
Run a FUSE server with `-o remember=1`, for instance the `hello` example:
```
./example/hello -f -s -o remember=1 /tmp/fu/
```
Press Ctrl-C. Glibc will diagnose invalid free:
```
free(): invalid pointer
Aborted
```
**Expected behavior**
The library should not pass an invalid pointer to `free`.

amonakov:
Starting from commit 752b59ac0876, the buffer must be freed with fuse_buf_free, not plain free.

Fixes #1376.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -4626,7 +4626,7 @@ static int fuse_session_loop_remember(struct fuse *f)
 		}
 	}
 
-	free(fbuf.mem);
+	fuse_buf_free(&fbuf);
 	return res < 0 ? -1 : 0;
 }
 


bsbernd:
Thank @amonakov Looks like we don't run a test with this function.

PR #1382, opened 2025-12-03
disable default FUSE_CAP_AUTO_INVAL_DATA in FUSE_INIT handler

bsbernd:
FUSE_AUTO_INVAL_DATA has a known data corruption case and handling it in kernel is complicated, this should not be enabled by default

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -1010,10 +1010,15 @@ void fuse_lib_help(struct fuse_args *args);
 /* Do not call this directly, use fuse_new() instead */
 struct fuse *_fuse_new_30(struct fuse_args *args,
 			  const struct fuse_operations *op, size_t op_size,
+			  unsigned int user_apiabi_version,
 			  struct libfuse_version *version, void *user_data);
 struct fuse *_fuse_new_31(struct fuse_args *args,
 			  const struct fuse_operations *op, size_t op_size,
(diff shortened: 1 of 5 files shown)


bsbernd:
@szmi any objections against this?

szmi:
I very much agree with the change, but have fears about regressions.

I'd do it with FUSE_USE_VERSION and library versioning to make sure existing binaries or plain recompile (without adjusting version) doesn't change behavior.

It's a pain for sure and it would be nice to batch multiple of these into one version update, but I'm not sure how feasible that is.

trapexit:
Is there a write up on this that can be included in the docs/code for devs?

bsbernd:
@trapexit I have some code additions that to store `FUSE_USE_VERSION` in `struct fuse_session`. The complicated part is in high level with ABI compatibility. A pity that I didn't add it in v3.17 when I added in the libfuse version.

Will also add a warning to FUSE_CAP_AUTO_INVAL_DATA, once I'm through with coding. Unfortunately doing too many things in parallel...

trapexit:
@bsbernd I should have been more clear. I was referring to dealing with FUSE_AUTO_INVAL_DATA. It's one thing to warn not to use it. It's another to explain the issue and offer details on how to replace its use.

PR #1387, opened 2025-12-10
fuse-io-uring: Handle EAGAIN/EINTR in fuse-io-uring

bsbernd:
Especially FUSE_IO_URING_CMD_REGISTER can easily get -EGAIN - if it arrives in kernel before FUSE_INIT
was finally handled. Easiest solution is to send it again. I'm not sure if FUSE_IO_URING_CMD_COMMIT_AND_FETCH might fail with -EINTR or -EGAIN - not from fuse client/kernel point of view, but possibly by io-uring. Better if we handle that as well.

### DIFF
--- a/lib/fuse_uring.c
+++ b/lib/fuse_uring.c
@@ -59,6 +59,9 @@ struct fuse_ring_queue {
 	size_t req_header_sz;
 	struct io_uring ring;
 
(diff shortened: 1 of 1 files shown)


bsbernd:
I need split up it, it includes some cleanup

PR #1405, opened 2025-12-29
add support for more mount options to avoid xfstests failure

lostjeffle:
"mount: add relatime mount option " to avoid xfstests generic/003 skip
"mount: add remount mount option" to avoid xfstests generic/294 and generic/452 failure

### DIFF
--- a/lib/mount.c
+++ b/lib/mount.c
@@ -126,6 +126,7 @@ static const struct fuse_opt fuse_mount_opts[] = {
 	FUSE_OPT_KEY("nolazytime",		KEY_KERN_FLAG),
 	FUSE_OPT_KEY("symfollow",		KEY_KERN_FLAG),
 	FUSE_OPT_KEY("nosymfollow",		KEY_KERN_FLAG),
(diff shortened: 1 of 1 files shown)


bsbernd:
@lostjeffle I agree, see commit e66c921 and 81ad52c7. But then also see dba6b3983af34f30de01cf532dff0b66f0ed6045, the latter came up after discussing at LFSMM in 2023.

lostjeffle:
>  But then also see [dba6b39](https://github.com/libfuse/libfuse/commit/dba6b3983af34f30de01cf532dff0b66f0ed6045), the latter came up after discussing at LFSMM in 2023.

@bsbernd So these time variant mount options are senseless in writeback mode as it is the FUSE server (rather than the kernel)'s responsibility to update atime in writeback mode.

lostjeffle:
I could drop commit "mount: add relatime mount option " and update the PR if commit "mount: add remount mount option" still makes sense.

bsbernd:
@lostjeffle I'm actually not sure if remount works without changes to fuse kernel. Did you check? Unfortunately I currently don't get much time to work on fuse myself and I didn't have time to check.

PR #1414, opened 2026-01-06
Famfs support into libfuse

jagalactic:
Famfs - the Fabric-attached Memory File System - supports organizing disaggregated shared memory (e.g. CXL) as a scale-out file system. This patch goes with the famfs V3 kernel patch set (which will go out within about a day - and I'll add a link to it here as soon as it's out).

I think both this and the kernel patch series are mergeable, but I look forward to review comments.

Thanks!

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -53,7 +53,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@5d4e8d1aca955e8d8589aabd499c5cae939e33c7 # v3.29.5
+      uses: github/codeql-action/init@cdefb33c0f6224e58673d9004f47f7cb3e328b89 # v3.29.5
       with:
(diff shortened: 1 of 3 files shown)


jagalactic:
Here is the full patch bundle (kernel, libfuse and ndctl) on Lore: https://lore.kernel.org/linux-fsdevel/20260107153244.64703-1-john@groves.net/T/#t

bsbernd:
Hi John, I'm going to convert it to draft until merged in the kernel.

jagalactic:
> Hi John, I'm going to convert it to draft until merged in the kernel.

Thanks Bernd. I'm getting ready to post V4 of the kernel patch set, and it drops the need for a custom kernel mount opt (-o shadow). So that commit is or will be dropped from the libfuse patch as well. I pushed that, but I'm not sure whether it updated the pull request or not.

bsbernd:
Hi John, it shows " @jagalactic jagalactic force-pushed the famfs-6.19 branch from af84639 to 0b16c7d 1 hour ago"

PR #1423, opened 2026-01-16
Explain fuse_out_header.error is -errno

stepancheg:


bsbernd:
@stepancheg I'm sorry for late reply. I'm afraid I cannot take it, as fuse_kernel.h is a copy of  `<linux>/include/uapi/linux/fuse.h` - i.e. if I apply it, the next time we update `fuse_kernel.h` your update will be gone. Feel free to submit it to linux-fsdevel@vger.kernel.org <linux-fsdevel@vger.kernel.org> with `fuse: ` as subject. Also feel free to add to `doc/libfuse-operations.txt` - this is pure libfuse documentation for now.

PR #1430, opened 2026-01-29
Add support for the FUSE_SYNCFS operation

megari:
Add support for handling the FUSE_SYNCFS operation in both the low-level and the high-level library.

Since Linux kernel 6.18, FUSE_SYNCFS requests have been passed on to all fuseblk file servers. Add the libfuse bits required to allow filesystems to implement the operation.

### DIFF
--- a/include/fuse.h
+++ b/include/fuse.h
@@ -867,6 +867,24 @@ struct fuse_operations {
 	 */
 	int (*statx)(const char *path, int flags, int mask, struct statx *stxbuf,
 		     struct fuse_file_info *fi);
+
+	/**
+	 * Synchronize the filesystem.
+	 *
+	 * Causes all dirty file data and filesystem metadata to be written to
+	 * underlying persistent storage.
+	 *
+	 * Supported since Linux kernel 6.18, and only on fuseblk file servers.
+	 *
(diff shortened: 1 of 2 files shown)


lib/fuse_lowlevel.c
```diff
@@ -3468,6 +3486,7 @@ static struct {
 	[FUSE_COPY_FILE_RANGE] = { do_copy_file_range, "COPY_FILE_RANGE" },
 	[FUSE_COPY_FILE_RANGE_64] = { do_copy_file_range_64, "COPY_FILE_RANGE_64" },
 	[FUSE_LSEEK]	   = { do_lseek,       "LSEEK"	     },
+	[FUSE_SYNCFS]	   = { do_syncfs,      "SYNCFS"      },
```
bsbernd: Thanks, this missed the new ops2 table, which is used with io-uring, could you please also do that?

bsbernd (changes requested):
Add the op code to the ops2 table please

megari:
FWIW, I fixed up the checkpatch.pl complaints.

bsbernd (approved):

bsbernd:
Thank you7 @megari !

megari:
My pleasure, @bsbernd !

PR #1441, opened 2026-02-28
very minor fix: add assert to prevent NULL pointer dereference in try_get_path

mikhailnov:
Svace statical analyzer reported a DEREF_AFTER_NULL.EX defect at lib/fuse.c:1022

```
After having been compared to a NULL value at
fuse.c:1017, pointer 'name' is passed as 3rd parameter in
call to function 'lookup_node' at fuse.c:1026, where it is
dereferenced at fuse.c:856. (CWE476)
```

This is actually a false positive, but there is already one assert(), we can add another one to ensure correct usage of this function.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -1019,6 +1019,7 @@ static int try_get_path(struct fuse *f, fuse_ino_t nodeid, const char *name,
 
 	if (wnodep) {
 		assert(need_lock);
+		assert(name != NULL);
 		wnode = lookup_node(f, nodeid, name);
 		if (wnode) {
 			if (wnode->treelock != 0) {


bsbernd:
@mikhailnov Sorry, I was traveling. Could you please add your Signed-off-by line?

mikhailnov:
Done

bsbernd:
@mikhailnov Sorry, still checkpatch warnings.

mikhailnov:
Ah, I see, "fixed" it now

bsbernd:
@mikhailnov keep it as it is please, I will try to suppress that "WARNING: Non-standard signature: Co-authored-by:"

mikhailnov:
Ok

bsbernd:
@mikhailnov Could you actually change to "Co-developed-by:" because that is what checkpatch.pl handles. Seems there is no suppression option for your signature and I wouldn't like to modify it, because the script is an unmodified copy from Linux.

mikhailnov:
probably a good idea, done

bsbernd:
@mikhailnov Thank you and sorry for all the trouble with checkpatch. I gave up on the last warnings. Checkpatch is basically there to verify that indentation is right and the Signed-off line is there - the other parts are bit over strict.

mikhailnov:
I understand, no problem ))
Thanks!

Issue #1443, opened 2026-03-03
Should not use io_uring_submit_and_wait to wait cqe in fuse-over-uring implementation


z47z:
In fuse_uring_thread. There‘s a infinite loop to wait and handle cqe.  io_uring_submit_and_wait involved assumes
that sqe must be submitted in uring_thread. Not a good idea, sqe can be submitted in any thread ideally. Only one constraint should be sufficient is SPSC of SQ with kernel thread.
```
	while (!atomic_load_explicit(&se->mt_exited, memory_order_relaxed)) {
		io_uring_submit_and_wait(&queue->ring, 1);

		pthread_mutex_lock(&queue->ring_lock);
		queue->cqe_processing = true;
		err = fuse_uring_queue_handle_cqes(queue);
		queue->cqe_processing = false;
		pthread_mutex_unlock(&queue->ring_lock);
		if (err < 0)
			goto err;
	}
```

bsbernd:
@z47z I'm afraid I'm lost, what hinders you to submit from multiple threads? It is not the preferred io-uring way, but I don't see what would hinder you to submit from other threads - fuse_uring_commit_sqe() has that. There is just an optimization to avoid submitting for every completion, as that requires a syscall.

z47z:
@bsbernd  I found submit sqe in different thread could cause double submition of  one commit_id by tracing kernel func 'fuse_uring_commit_fetch' . If this usage is not a good practice. How about removing ring_lock. Or not use io_uring_submit_and_wait, just wait cqe on this point.

bsbernd:
I don't see you you can double submit, it takes an SQE and one thread submits it, which is not important. 

io_uring_submit_and_wait() is useful when there is just a single ring thread doing the work.

z47z:
By modifying passthrough_hp. Add an extra thread to send_reply all requests, could produce following results.    
       
<img width="783" height="479" alt="Image" src="https://github.com/user-attachments/assets/16f0da5f-e666-46cd-bbea-f143de6c2cc4" />

<img width="648" height="583" alt="Image" src="https://github.com/user-attachments/assets/35258ffa-c22d-48c9-b8de-62598270f467" />

bsbernd:
Please provide the patch that triggers that behavior and description how to reproduce.

z47z:
```
From b51b84d8adfa348d1fd8c0cac195d341f75078c7 Mon Sep 17 00:00:00 2001
[...]

PR #1449, opened 2026-03-13
fix: enforce permission checks for allow_other mounts

### LINKED ISSUE #15: Bug : The inodes entries cached by fuse are not refreshed upon returning from setattr().
When the inodes are cached by fuse (non-zero attr_timeout), and the standard_permissions are used, the following sequence :
```
chmod 755 trydir
rm -rf trydir
mkdir trydir
echo file > trydir/file
ls -li trydir/file    
chmod 000 trydir
chmod 444 trydir/file
ls -li trydir/file
```
outputs :

```
293 -rw-rw-r-- 1 linux linux 5 2009-08-25 17:50 trydir/file
293 -r--r--r-- 1 linux linux 5 2009-08-25 17:50 trydir/file
```
The "chmod 000" should have led to an EACCES error for the subsequent commands.

This was reported in 2009 (http://article.gmane.org/gmane.comp.file-systems.fuse.devel/8099), and according to http://article.gmane.org/gmane.comp.file-systems.fuse.devel/15300 some work is still needed on it.

This means that for being Posix compliant, fuse file systems cannot either use the cache or have the permissions checked by the kernel.

ShukantPal:
Closes #15

When default_permissions is not set and allow_other is enabled, the kernel's fuse_permission() returns 0 for most permission checks without consulting userspace. This means that after a chmod 000, cached dentries still allow access by other users — the kernel never sends FUSE_ACCESS to verify permissions.

This is a two-layer problem requiring fixes in both the kernel and the FUSE daemon:

**Daemon fix (example/passthrough.c)**:
  Fix xmp_access() to check permissions as the requesting user (via fuse_get_context()) rather than as root. Uses raw SYS_setresuid/SYS_setresgid syscalls for per-thread credential switching, since glibc wrappers synchronize across all threads.

**Associated kernel patch (fs/fuse/dir.c fuse_permission)**:
  When allow_other is set and default_permissions is not, add an else-if branch that sends FUSE_ACCESS to userspace for all permission checks. Includes a MAY_NOT_BLOCK guard that returns -ECHILD to force retry outside RCU walk mode, since fuse_access() sends a blocking request to userspace.

```diff
--- a/fs/fuse/dir.c
+++ b/fs/fuse/dir.c
@@ -1607,6 +1607,17 @@ static int fuse_permission(...)
      } else if (mask & (MAY_ACCESS | MAY_CHDIR)) {
          err = fuse_access(inode, mask);
+     } else if (fc->allow_other) {
+         /*
+          * With allow_other, multiple users can access this mount.
+          * Verify permissions via userspace to prevent user A from
+          * accessing files/directories owned by user B with
+          * restrictive permissions.
+          */
+         if (mask & MAY_NOT_BLOCK)
+             return -ECHILD;
+
+         err = fuse_access(inode, mask);
      } else if ((mask & MAY_EXEC) && S_ISREG(inode->i_mode)) {
```

**Tests (xfail without kernel patch)**:
- test_perm_cache_dir_traversal: chmod 000 dir, verify files inside are denied
- test_perm_cache_file_truncate: chmod 000 file, verify truncate is denied
- test_perm_cache_dir_read_after_noexec: remove dir execute bit, verify traversal is denied

I have verified that recompiling the Linux kernel with the patch passes the above added tests.

### DIFF
--- a/example/passthrough.c
+++ b/example/passthrough.c
@@ -38,6 +38,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/stat.h>
+#include <sys/syscall.h>
 #include <dirent.h>
 #include <errno.h>
 #include <sys/time.h>
@@ -96,11 +97,35 @@ static int xmp_getattr(const char *path, struct stat *stbuf,
 
 static int xmp_access(const char *path, int mask)
 {
-	int res;
+	struct fuse_context *ctx = fuse_get_context();
+	int res, saved_errno;
+	int switched = 0;
+
+	/*
+	 * When running as root, temporarily switch to the requesting
+	 * user's credentials so that access() checks permissions
+	 * correctly for multi-user mounts (allow_other).
+	 *
+	 * We use raw syscalls because glibc's setresuid/setresgid
+	 * wrappers synchronize across all threads, but the raw
+	 * syscalls are per-thread on Linux.
+	 */
+	if (getuid() == 0 && ctx->uid != 0) {
+		syscall(SYS_setresgid, ctx->gid, ctx->gid, 0);
+		syscall(SYS_setresuid, ctx->uid, ctx->uid, 0);
+		switched = 1;
+	}
 
 	res = access(path, mask);
+	saved_errno = errno;
+
+	if (switched) {
+		syscall(SYS_setresuid, 0, 0, 0);
+		syscall(SYS_setresgid, 0, 0, 0);
+	}
+
 	if (res == -1)
-		return -errno;
+		return -saved_errno;
 
 	return 0;
 }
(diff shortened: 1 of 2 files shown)


bsbernd:
@ShukantPal This probably works, but increases overhead. For me the issue in  #15 looks like a missing attr invalidation. I'm going to try (without your patch) if that is reproducible, got a bit late - will do that tomorrow.

ShukantPal:
@bsbernd Did you get a chance reproduce the issue? This PR attaches failing unit tests that reproduced the issue (they are marked as expecting to fail), which may aid your repro.

bsbernd:
@ShukantPal sorry, not yet. I will really try during the next days, though, I'm about to switch work (super busy to finish things with the current one). Please be patient a bit. In the order of things, I also first need to look into https://github.com/libfuse/libfuse/issues/1443.

bsbernd:
I'm re-opening it, to avoid that it gets forgotten

PR #1479, opened 2026-04-15
Return `-ENOMEM` instead of `ENOMEM` for consistency

gaul:
Suggested by Claude.


bsbernd (approved):
@gaul Looks good, for next time please change the commit message to "Return negative error code as documented in the header files".

PR #1480, opened 2026-04-15
Fix bugs and harden io-uring, session teardown, and fusermount

abhinavagarwal07:
## Summary

- Fix double `pthread_mutex_destroy` of `se->mt_lock` on every MT shutdown
- Fix dangling `se->uring.pool` pointer after `fuse_uring_stop`
- Fix unchecked `fstat` return in fusermount comm fd validation
- Fix mountpoint string leak in `fuse_session_destroy` when unmount is skipped
- Add `O_CLOEXEC` to procfs open in `grow_pipe_to_max` with portability fallback
- Use `snprintf` instead of `sprintf` for fixed-size path buffer


lib/fuse_uring.c
```diff
@@ -891,7 +907,9 @@ int fuse_uring_start(struct fuse_session *se)
 	int err = 0;
 	struct fuse_ring_pool *fuse_ring;
 
-	fuse_uring_sanity_check(se);
+	err = fuse_uring_sanity_check(se);
+	if (err)
+		return err;
```
bsbernd: Oh yes, that is actually already fixed here, https://github.com/bsbernd/libfuse/commits/uring-reduce-nr-queues/,  run into it last weekend. 


https://github.com/bsbernd/libfuse/commit/1871089a8c1093f9c33d64a0f24adaa1e65dde8c#diff-1c98718941460331f4c4aa982d4d212905af21a8e9eabcb80431e3b7f163696fL894

bsbernd:
Thank you @abhinavagarwal07 !

PR #1481, opened 2026-04-15
fuse_uring: fix teardown UAF by draining in-flight requests

abhinavagarwal07:
Fix an io_uring teardown lifetime race where deferred replies could
outlive ring teardown and touch ring/session state after it had been
freed.

This change wakes and joins ring threads without pthread_cancel(), tracks
dispatched io_uring requests with a per-pool in-flight counter, and waits
for outstanding replies to release before freeing ring resources. If
draining cannot complete within the bounded timeout, teardown reports
failure and leaves the still-referenced objects alive instead of freeing
them.

It also disables FUSE_INTERRUPT handling for io_uring requests and flushes
SQEs queued by callbacks before ring threads exit.

### DIFF
--- a/lib/fuse_uring.c
+++ b/lib/fuse_uring.c
@@ -390,28 +390,73 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
 
 static void fuse_session_destruct_uring(struct fuse_ring_pool *fuse_ring)
 {
+	/*
+	 * Do NOT use pthread_cancel() — it can kill a thread while it
+	 * holds queue->ring_lock inside a filesystem callback, leaving
+	 * the mutex permanently locked and deadlocking any application
+	 * thread that tries to reply via fuse_uring_commit_sqe().
+	 *
+	 * Instead, wake all sleepers, join threads (stopping new
+	 * dispatches), then wait for deferred replies before freeing.
+	 */
+
+	/* Wake all sleepers: sem_wait (init path) and io_uring (main loop) */
 	for (size_t qid = 0; qid < fuse_ring->nr_queues; qid++) {
 		struct fuse_ring_queue *queue =
 			fuse_uring_get_queue(fuse_ring, qid);
 
-		if (queue->tid != 0) {
+		sem_post(&fuse_ring->init_sem);
+
+		if (queue->tid != 0 && queue->eventfd >= 0) {
 			uint64_t value = 1ULL;
 			int rc;
 
 			rc = write(queue->eventfd, &value, sizeof(value));
 			if (rc != sizeof(value))
-				fprintf(stderr,
-					"Wrote to eventfd=%d err=%s: rc=%d\n",
-					queue->eventfd, strerror(errno), rc);
-			pthread_cancel(queue->tid);
+				fuse_log(FUSE_LOG_ERR,
+					 "eventfd=%d write err=%s: rc=%d\n",
+					 queue->eventfd, strerror(errno), rc);
+		}
+	}
+
+	/* Join all threads — no new dispatches after this */
+	for (size_t qid = 0; qid < fuse_ring->nr_queues; qid++) {
+		struct fuse_ring_queue *queue =
+			fuse_uring_get_queue(fuse_ring, qid);
+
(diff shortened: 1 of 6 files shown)


lib/fuse_uring.c
```diff
@@ -390,24 +390,53 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
...
+			fuse_uring_get_queue(fuse_ring, qid);
+
+		if (queue->tid != 0) {
 			pthread_join(queue->tid, NULL);
 			queue->tid = 0;
 		}
 
+		for (size_t idx = 0; idx < fuse_ring->queue_depth; idx++) {
+			while (atomic_load(&queue->ent[idx].req.ref_cnt) > 1)
+				usleep(1000);
```
bsbernd: This might never happen, if the daemon is not processing requests anymore and would hang here forever. I also wouldn't like a poll loop.

bsbernd: I guess the easier way is then to say: We cannot free the queue, because there are still references.

abhinavagarwal07:
Agreed — the poll loop is wrong. Current version does an early return and leaks ring memory instead.

But that's still not enough: a late uring reply dereferences `ring_pool->se` in `fuse_uring_commit_sqe`, and the caller frees `se` after `fuse_uring_stop()` returns. So preserving ring memory alone doesn't make teardown safe.

I think we either need to block until deferred replies drain, or add an abort semantic for outstanding requests. What do you prefer?

lib/fuse_uring.c
```diff
@@ -389,28 +389,73 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
...
+	for (size_t qid = 0; qid < fuse_ring->nr_queues; qid++) {
+		struct fuse_ring_queue *queue =
+			fuse_uring_get_queue(fuse_ring, qid);
+
+		for (size_t idx = 0; idx < fuse_ring->queue_depth; idx++) {
+			if (atomic_load(&queue->ent[idx].req.ref_cnt) > 1) {
+				fuse_log(FUSE_LOG_ERR,
+					"Leaked ring: refs held at shutdown\n");
+				return;
+			}
```
bsbernd, line 448: I think here we should now call `io_uring_queue_exit`, I guess we don't even need to check for references.

lib/fuse_uring.c
```diff
@@ -389,28 +389,73 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
...
+		}
+	}
+
+	/* Join all threads — no new dispatches after this */
+	for (size_t qid = 0; qid < fuse_ring->nr_queues; qid++) {
+		struct fuse_ring_queue *queue =
+			fuse_uring_get_queue(fuse_ring, qid);
+
+		if (queue->tid != 0) {
 			pthread_join(queue->tid, NULL);
```
bsbernd, line 427: In worst case this could wait forever, I really believe we need to keep the pthread_cancel(). What we could do is, is use `pthread_cleanup_push` and `pthread_cleanup_pop` and then release the mutex lock in the push callback function, if self acuqired

abhinavagarwal07: This was also on the older unbounded-join version. Current code uses one bounded monotonic deadline for thread exit + reply drain, so teardown should not wait forever.

lib/fuse_uring.c
abhinavagarwal07: This was on an older version of the diff. Current code calls `io_uring_queue_exit()` only after the ring threads have exited and the in-flight replies have drained.

bsbernd:
@abhinavagarwal07 The diff is much much too large, for a simple cleanup issue. I also go through individual commits, if the 1st commit does something and the 2nd commit reverts that - becomes hard to review. I think the teardown solution is way too complex for what needs to do.

abhinavagarwal07:
Agreed. The PR is still in draft state. I am trying to simplify it. To avoid confusion let me close this PR for now and reopen when its ready.

bsbernd:
I will also try to find a simpler solution.

PR #1487, opened 2026-04-23
test stub for FUSE FORGET over io-uring

dragonylffly:
Enable testing against the kernel feature that optionally delivers FORGET over io-uring when userspace sets FUSE_IO_URING_REGISTER_FORGET_COMMIT on the REGISTER command.

Changes:
- include/fuse_kernel.h: Bump protocol minor to 7.46 and define FUSE_IO_URING_REGISTER_FORGET_COMMIT for fuse_uring_cmd_req.flags on REGISTER.
- lib/fuse_uring.c: Set that flag in fuse_uring_register_ent so the kernel submits FORGET requests on the io-uring ring.
- lib/fuse_lowlevel.c: For FORGET and BATCH_FORGET delivered via io-uring, complete with fuse_reply_err(0) without calling op.forget, to avoid real inode teardown during testing.

For testing the "Send FORGET over io_uring" kernel patch only; currently not intended for upstream merging.

### DIFF
--- a/include/fuse_kernel.h
+++ b/include/fuse_kernel.h
@@ -240,6 +240,9 @@
  *  - add FUSE_COPY_FILE_RANGE_64
  *  - add struct fuse_copy_file_range_out
  *  - add FUSE_NOTIFY_PRUNE
+ *
+ *  7.46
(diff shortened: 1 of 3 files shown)


bsbernd:
Let's close it, until accepted kernel upstream.

PR #1490, opened 2026-04-30
fuse_loop_cfg_convert: restore v1 max_idle_threads semantics

amir73il:
Commit 30a126c5 ("API update for fuse_loop_config additions") introduced fuse_loop_cfg_convert() to translate legacy fuse_loop_config_v1 structs to the new v2 API, but only converted max_idle_threads — it did not set max_threads.

In the v1 API, max_idle_threads was the effective pool cap: the thread pool would grow on demand and threads exceeding max_idle_threads were destroyed after processing each request.  A filesystem that configured max_idle_threads=100 expected up to 100 threads to be available.

After the conversion, max_threads retains its default of 10, silently capping the pool at 10 workers regardless of max_idle_threads.  A busy filesystem with all 10 threads processing long-running requests has no thread available to receive the next kernel request, causing it to hang.

Fix fuse_loop_cfg_convert() to call fuse_loop_cfg_set_max_threads() with the same value as max_idle_threads before calling fuse_loop_cfg_set_idle_threads(), preserving the original v1 semantics.

Also add warnings to both setters when max_idle_threads and max_threads are set in an inconsistent order (max_idle > max_threads).  Reorder the two setter calls in helper.c so that max_threads is set before max_idle_threads, preventing a spurious warning when both are supplied as consistent command-line options.  Add a unit test to document and verify the correct behavior.

### DIFF
--- a/test/test_loop_config.c
+++ b/test/test_loop_config.c
@@ -1,0 +1,288 @@
+/*
+ * Unit tests for fuse_loop_cfg_set_idle_threads() /
+ * fuse_loop_cfg_set_max_threads() interaction.
+ *
+ * No FUSE mount is needed; the tests exercise the setter logic directly.
+ */
+
+#define FUSE_USE_VERSION FUSE_MAKE_VERSION(3, 12)
+
+#include "fuse_i.h"          /* internal struct fuse_loop_config (v2) */
+#include "fuse_lowlevel.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <assert.h>
+
+/* ------------------------------------------------------------------ */
+/* Log capture                                                         */
+/* ------------------------------------------------------------------ */
+
+static char last_log_msg[512];
+static enum fuse_log_level last_log_level;
+
+static void test_log_handler(enum fuse_log_level level, const char *fmt,
+			     va_list ap)
+{
+	last_log_level = level;
+	vsnprintf(last_log_msg, sizeof(last_log_msg), fmt, ap);
+}
+
+static void clear_log(void)
+{
+	last_log_msg[0] = '\0';
+	last_log_level = FUSE_LOG_DEBUG;
+}
+
+/* ------------------------------------------------------------------ */
+/* Helpers                                                             */
+/* ------------------------------------------------------------------ */
+
+#define PASS(name) do { printf("PASS: %s\n", name); } while (0)
+#define FAIL(name, fmt, ...) \
+	do { fprintf(stderr, "FAIL: %s — " fmt "\n", name, ##__VA_ARGS__); \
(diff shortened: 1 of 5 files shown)


amir73il:
@bsbernd we have an fs which sets --max_idle_threads=100 because it expects to have 100 active threads
The result of thread pool limit of 10 after libfuse upgrade to 3.12 was surprising and seems unintended.
I understand that disabling unbounded thread pool was an intentional behavior change, but this case seems
at least reasonable to raise the limit to the specified max_idle_threads

bsbernd:
@amir73il I need to think about it and look into the code what went wrong, just now time right now (will try later in the evening). The actual idea was that max_threads does not have an effect if max_idle_threads is set.

amir73il:
> I need to think about it and look into the code what went wrong, just now time right now (will try later in the evening). The actual idea was that max_threads does not have an effect if max_idle_threads is set.

TBH this is what I expected and that means we want to set max_threads = UINT_MAX in fuse_loop_cfg_convert(), not max_threads = max_idle_threads.

This makes sense to me. Let me know if I should make this change. 

Please take your time. 
No rush!!!!

test/test_loop_config.c
```diff
@@ -0,0 +1,288 @@
...
+	PASS(name);
+}
+
+/* ------------------------------------------------------------------ */
+/* main                                                                */
+/* ------------------------------------------------------------------ */
+
+int main(void)
+{
+	fuse_set_log_func(test_log_handler);
```
bsbernd, line 273: That is a really interesting way to check logs!

bsbernd (approved):
Thanks Amir!

bsbernd:
I'm going to disable that SINGLE_STATEMENT_DO_WHILE_MACRO checkpatch warning, it is a bit weird.

bsbernd:
> > I need to think about it and look into the code what went wrong, just now time right now (will try later in the evening). The actual idea was that max_threads does not have an effect if max_idle_threads is set.
> 
> TBH this is what I expected and that means we want to set max_threads = UINT_MAX in fuse_loop_cfg_convert(), not max_threads = max_idle_threads.
> 
> This makes sense to me. Let me know if I should make this change.

I really thought there was already a protection in `fuse_loop_mt.c`. I believe to remember that I had added something, but maybe had removed that during review or some fixes, don't know anymore. Your way with the warning is probably more elegant, no run time checks needed. 
If we should entirely ignore max_threads when max_idle_threads is set is anyway debatable, reaping threads is so expensive - it was the reason why I initially started to work on that. At best the code would need to be changed to reap when idle with some timeouts. But for that these threads must not go into /dev/fuse read, because kernel randomly assigns work to the threads.

>

PR #1492, opened 2026-05-01
Add a known-issues file documenting the limitations of the current auto-unmount implementation.

### LINKED ISSUE #1491: auto_unmount does not work without allow_root
As I understand it, when specifying auto_unmount, fuse creates a pipe and forks a child that runs fusermount3 while monitors the receive end of the pipe. When that reads EOF, indicating that the parent holding the write end has died for whatever reason, then fusermount3 unmounts the file system again.

Except for me it doesn't. Stracing the fusermount3 process is revealing:

```
strace: Process 17986 attached
recvfrom(5, "", 16, 0, NULL, NULL)      = 0
chdir("/home/seb/TTDE/SDBS/testdata")   = 0
getcwd("/home/seb/TTDE/SDBS/testdata", 65536) = 29
clone(child_stack=0x7ffde03430b0, flags=CLONE_NEWNS) = 18023
wait4(18023, [{WIFEXITED(s) && WEXITSTATUS(s) == 0}], __WCLONE, NULL) = 18023
openat(AT_FDCWD, "/home/seb/TTDE/SDBS/testdata/fuse", O_RDONLY) = -1 EACCES (Permission denied)
exit_group(0)                           = ?
+++ exited with 0 +++
```

fusermount3 fails to verify the mount because it is setuid and root does not have access to a non-root users fuse mounts.

If I specify `allow_root` in the options and add `user_allow_other` in `/etc/fuse.conf` it starts working. However, this is undesirable since it unnecessarily adds security risks.

The proper fix is for fusermount3 to drop root privileges before verifying the mount with `seteuid`. In fact it can probably completely drop root privileges when run in this way (at least on Linux) since unmounting a users fuse does not require root privileges.

I'm on Ubuntu 24.04 with fuse3 3.14.0-5build1.

sebunger:
Closes #1491

### DIFF
--- a/doc/README.known-issues
+++ b/doc/README.known-issues
@@ -1,0 +1,12 @@
+Auto-Umount
+The mechanism currently implemented has several flaws:
+- It interacts badly with mount --move and mount --bind operations
+  on the fuse mount-point (and possibly on parent mount-points), and
+- It currently requires that the allow_root option be specified if
+  the filesystem is mounted by a non-root user which in turn may require
+  user_allow_other in /etc/fuse.conf.
+
+A new mechanism is being worked on which will address these shortcomings.
+
+See https://github.com/libfuse/libfuse/issues/1491 for some further information
+on the subject.


bsbernd:
@sebunger Thank you! Checkpatch will complain, there is a missing line `Signed-of-by: Name <mail@mydress>`

sebunger:
@bsbernd Looks like it failed because the commit isn't signed off or something. Am I supposed to do that? If so, can you tell me how? Haven't worked much with git in the past. I'm more a mercurial-kind-of-guy. ;-)

bsbernd:
@sebunger I was using mercurial until the world (and at that time missing mercurial features) forced me to git ;) Now using a mix of git, stgit and easygit commands. This one should do:

```
git commit --amend
<Add your Signed-off-by: line, see above>
git push -f
```

PR #1498, opened 2026-05-12
fuse: No need to call strlen

vlendec:
This only checks for "." and "..". This can be done cheaper than calling strlen. According to godbolt not much code is required for this.

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -2685,11 +2685,12 @@ static void fuse_lib_lookup(fuse_req_t req, fuse_ino_t parent,
 	struct node *dot = NULL;
 
 	if (name[0] == '.') {
-		int len = strlen(name);
+		bool isdot = (name[1] == '\0');
+		bool isdotdot = ((name[1] == '.') && (name[2] == '\0'));
 
-		if (len == 1 || (name[1] == '.' && len == 2)) {
+		if (isdot || isdotdot) {
 			pthread_mutex_lock(&f->lock);
-			if (len == 1) {
+			if (isdot) {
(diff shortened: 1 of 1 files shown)


lib/fuse.c
```diff
@@ -2685,11 +2685,12 @@ static void fuse_lib_lookup(fuse_req_t req, fuse_ino_t parent,
 	struct node *dot = NULL;
 
 	if (name[0] == '.') {
-		int len = strlen(name);
+		bool isdot = (name[1] == '\0');
+		bool isdotdot = ((name[1] == '.') && (name[2] == '\0'));
```
bsbernd, line 2689: @vlendec I think we might get a buffer overflow for `isdotdot`. How about

```
bool isdot = (name[1] == '\0');
bool isdotdot = false;

if (!isdot) && (name[1] == '.') && (name[2] == '\0'))
   isdotdot = true
```
inside of the `if (name[0] == '.') {` condition

bsbernd (changes requested):

lib/fuse.c
vlendec: I don't see the buffer overflow if we assume the string is null-terminated. The name[1]=='.' in isdotdot is only executed if name[0]=='.', and due to boolean short-curcuiting the check for '\0' is only executed if name[1]=='.'. If name[1] is for example '\0', the second arm of the && expression is not executed.

bsbernd: Ah right sorry.

bsbernd (approved):

PR #1499, opened 2026-05-12
fuse: prevent possible fuse_pollhandle leak

mihalicyn:
Documentation for poll() callback says that:
> The callee is responsible for destroying ph with
> fuse_pollhandle_destroy() when no longer in use.

In fuse_lib_poll() ((struct fuse_lowlevel_ops*)->poll) we need to be more careful:

1. If get_path_nullok() fails, we need to free fuse_pollhandle
2. If we passed execution down to fuse_fs_poll(), then it must release fuse_pollhandle resources when no (struct fuse_operations*)->poll provided by the filesystem driver.

Found this by myself while reading the code as a part of [1] review.

This is not critical at all, because once we return ENOSYS once, kernel never sends FUSE_POLL again. So memleak is unnoticable in practice. Alternatively, we can leak if get_path_nullok() fails all the time, but then we are in a much more serious troubles...

Link: https://github.com/lxc/lxcfs/pull/726 [1]

### DIFF
--- a/lib/fuse.c
+++ b/lib/fuse.c
@@ -2271,8 +2271,12 @@ int fuse_fs_poll(struct fuse_fs *fs, const char *path,
 	int res;
 
 	fuse_get_context()->private_data = fs->user_data;
-	if (!fs->op.poll)
+
+	if (!fs->op.poll) {
+		fuse_pollhandle_destroy(ph);
 		return -ENOSYS;
+	}
+
 	if (fs->debug)
 		fuse_log(FUSE_LOG_DEBUG, "poll[%llu] ph: %p, events 0x%x\n",
 			(unsigned long long) fi->fh, ph,
@@ -4348,6 +4352,8 @@ static void fuse_lib_poll(fuse_req_t req, fuse_ino_t ino,
 		err = fuse_fs_poll(f->fs, path, fi, ph, &revents);
 		fuse_finish_interrupt(f, req, &d);
 		free_path(f, ino, path);
+	} else {
+		fuse_pollhandle_destroy(ph);
 	}
 	if (!err)
 		fuse_reply_poll(req, revents);


bsbernd:
@mihalicyn Thank you, I need to read through this, probably only on Thursday.

lib/fuse.c
```diff
@@ -2271,8 +2271,12 @@ int fuse_fs_poll(struct fuse_fs *fs, const char *path,
 	int res;
 
 	fuse_get_context()->private_data = fs->user_data;
-	if (!fs->op.poll)
+
+	if (!fs->op.poll) {
+		fuse_pollhandle_destroy(ph);
 		return -ENOSYS;
+	}
+
```
bsbernd, line 2279: To be honest,  `fuse_fs_poll`  gives me some headache.  Problem is that this function is exported and can be called from anywhere and we do not know what users of this functions do. What we actually need to to encode FUSE_USE_VERSION into `struct fuse_fs` (and `struct fuse_session`) and then act on the version. Old file systems don't get the the handle destroy, new file systems will get it. However, in the end this means we cannot currently do it, because we don't store FUSE_USE_VERSION yet. I had already tried, but especially with high level ABI compat it complex and I didn't have time, with Darricks container PR (just merged) I got a much easier idea, just not done yet.

lib/fuse.c
```diff
@@ -4347,6 +4351,8 @@ static void fuse_lib_poll(fuse_req_t req, fuse_ino_t ino,
 		err = fuse_fs_poll(f->fs, path, fi, ph, &revents);
 		fuse_finish_interrupt(f, req, &d);
 		free_path(f, ino, path);
+	} else {
+		fuse_pollhandle_destroy(ph);
```
bsbernd, line 4356: This is one is safe, no issue here.

lib/fuse.c
mihalicyn: >Problem is that this function is exported and can be called from anywhere and we do not know what users of this functions do.

yes, but we already have `fs->op.poll(..)` call in here and documentation for `fs->op.poll` says:
```
	 * The callee is responsible for destroying ph with
	 * fuse_pollhandle_destroy() when no longer in use.
```

I.e. we assume that `ph` is valid pointer **or** NULL.

Therefore it must be safe to call `fuse_pollhandle_destroy` directly from `fuse_fs_poll` too.

bsbernd: Well let's try, in worse case I need to revert.

mihalicyn: Sure, let's keep an eye on this.

Thanks, Bernd ;-)

mihalicyn:
I have rebased and force-pushed this now.

bsbernd:
Thank you @mihalicyn Looks like there is an ABI test issue - it believes the PR removes function. I guess I need to add rebase to the ABI checker.

PR #1502, opened 2026-05-20
FUSE_SECCTX: add security context extension parsing and iterator API

kchen-ddn:
Add support for parsing FUSE_SECCTX extensions sent by the kernel during file creation operations. This allows userspace filesystems to receive and persist SELinux/AppArmor security labels on newly created files.

### DIFF
--- a/example/passthrough_hp.cc
+++ b/example/passthrough_hp.cc
@@ -174,6 +174,7 @@ struct Fs {
 	std::string fuse_mount_options;
 	bool direct_io;
 	bool passthrough;
+	bool selinux;
 };
 static Fs fs{};
 
@@ -212,6 +213,10 @@ static void sfs_init(void *userdata, fuse_conn_info *conn)
 		fuse_set_feature_flag(conn, FUSE_CAP_WRITEBACK_CACHE);
 
 	fuse_set_feature_flag(conn, FUSE_CAP_FLOCK_LOCKS);
+
+	/* Enable security context extension for SELinux support */
+	if (fs.selinux)
+		fuse_set_feature_flag(conn, FUSE_CAP_SECURITY_CTX);
 
 	if (fs.nosplice) {
 		// FUSE_CAP_SPLICE_READ is enabled in libfuse3 by default,
@@ -476,12 +481,104 @@ static void sfs_lookup(fuse_req_t req, fuse_ino_t parent, const char *name)
 	}
 }
 
+#ifdef HAVE_SETXATTR
+/**
+ * Stage the requester's SELinux label so the next create is born with it.
+ *
+ * The kernel computes the label a newly created inode should have in the
+ * requesting task's context and sends it in the FUSE security context
+ * extension. passthrough_hp must act on it: a bare mknodat/openat on the
+ * backing filesystem would label the inode from the daemon's own context, not
+ * the requester's. Writing the label to the per-thread fscreate attribute makes
+ * the following create syscall produce a correctly-labelled backing inode
+ * atomically. The override is per worker thread and must be cleared afterwards
+ * with sfs_clear_selinux_fscreate() so later creates are unaffected.
+ *
+ * Demonstrates the security context extension API: fuse_req_secctx_count(),
+ * fuse_req_secctx_reset(), fuse_req_secctx_next().
+ *
+ * @param req FUSE request carrying the security context extension
+ * @return true if a label was staged and must be cleared afterwards
+ */
+static bool sfs_set_selinux_fscreate(fuse_req_t req)
+{
+	if (fuse_req_secctx_count(req) == 0)
+		return false;
+
+	const char *ctx_name;
+	const char *ctx_value;
+	uint32_t ctx_len;
+	bool staged = false;
+
+	while (fuse_req_secctx_next(req, &ctx_name, &ctx_value, &ctx_len) == 0) {
+		if (strcmp(ctx_name, "security.selinux") != 0)
+			continue;
+
+		int fd = open("/proc/thread-self/attr/fscreate",
+			      O_WRONLY | O_CLOEXEC);
+		if (fd == -1) {
+			fuse_log(FUSE_LOG_DEBUG,
+				 "sfs_set_selinux_fscreate: open failed: %s\n",
+				 strerror(errno));
+			break;
+		}
+		/* The label value already includes its trailing NUL. */
+		if (write(fd, ctx_value, ctx_len) == -1)
+			fuse_log(FUSE_LOG_DEBUG,
+				 "sfs_set_selinux_fscreate: write failed: %s\n",
+				 strerror(errno));
+		else {
+			staged = true;
+			/*
+			 * Debug log to verify the label was staged correctly.
+			 * This is important because the underlying filesystem will apply
+			 * a default label anyway, so without this log it's hard to tell
+			 * if our SELinux context handling is actually working.
+			 */
+			fuse_log(FUSE_LOG_DEBUG,
+				 "sfs_set_selinux_fscreate: staged label: %s\n",
+				 ctx_value);
+		}
+		close(fd);
+		break;
+	}
+	fuse_req_secctx_reset(req);
+	return staged;
+}
+
+/* Reset the per-thread fscreate label staged by sfs_set_selinux_fscreate(). */
+static void sfs_clear_selinux_fscreate(void)
+{
+	int fd = open("/proc/thread-self/attr/fscreate", O_WRONLY | O_CLOEXEC);
+	if (fd == -1)
+		return;
+	/* A zero-length write clears the override. */
+	if (write(fd, "", 0) == -1)
+		fuse_log(FUSE_LOG_DEBUG,
+			 "sfs_clear_selinux_fscreate: write failed: %s\n",
+			 strerror(errno));
+	close(fd);
+}
+#else
+static bool sfs_set_selinux_fscreate(fuse_req_t req)
+{
+	(void)req;
+	return false;
+}
+
+static void sfs_clear_selinux_fscreate(void)
+{
+}
+#endif
+
 static void mknod_symlink(fuse_req_t req, fuse_ino_t parent, const char *name,
 			  mode_t mode, dev_t rdev, const char *link)
 {
 	int res;
 	Inode &inode_p = get_inode(parent);
 	auto saverr = ENOMEM;
+
+	bool labeled = sfs_set_selinux_fscreate(req);
 
 	if (S_ISDIR(mode))
 		res = mkdirat(inode_p.fd, name, mode);
@@ -490,6 +587,11 @@ static void mknod_symlink(fuse_req_t req, fuse_ino_t parent, const char *name,
 	else
 		res = mknodat(inode_p.fd, name, mode, rdev);
 	saverr = errno;
+
+	/* Reset the staged label so later creates on this thread are unaffected. */
+	if (labeled)
+		sfs_clear_selinux_fscreate();
+
 	if (res == -1)
 		goto out;
 
@@ -940,10 +1042,18 @@ static void sfs_create(fuse_req_t req, fuse_ino_t parent, const char *name,
 {
 	Inode &inode_p = get_inode(parent);
 
+	bool labeled = sfs_set_selinux_fscreate(req);
+
 	auto fd = openat(inode_p.fd, name, (fi->flags | O_CREAT) & ~O_NOFOLLOW,
 			 mode);
+	auto create_errno = errno;
+
+	/* Reset the staged label so later creates on this thread are unaffected. */
+	if (labeled)
+		sfs_clear_selinux_fscreate();
+
 	if (fd == -1) {
-		auto err = errno;
+		auto err = create_errno;
 		if (err == ENFILE || err == EMFILE)
 			cerr << "ERROR: Reached maximum number of file descriptors."
 			     << endl;
@@ -1024,10 +1134,18 @@ static void sfs_tmpfile(fuse_req_t req, fuse_ino_t parent, mode_t mode,
 {
 	Inode &parent_inode = get_inode(parent);
 
+	bool labeled = sfs_set_selinux_fscreate(req);
+
 	auto fd = openat(parent_inode.fd, ".",
 			 (fi->flags | O_TMPFILE) & ~O_NOFOLLOW, mode);
+	auto create_errno = errno;
+
+	/* Reset the staged label so later creates on this thread are unaffected. */
+	if (labeled)
+		sfs_clear_selinux_fscreate();
+
 	if (fd == -1) {
-		auto err = errno;
+		auto err = create_errno;
 		if (err == ENFILE || err == EMFILE)
 			cerr << "ERROR: Reached maximum number of file descriptors."
 			     << endl;
@@ -1486,6 +1604,7 @@ static cxxopts::ParseResult parse_options(int argc, char **argv)
 		("nocache", "Disable attribute all caching")
 		("nosplice", "Do not use splice(2) to transfer data")
 		("nopassthrough", "Do not use pass-through mode for read/write")
+		("selinux", "Enable SELinux context from FUSE request extension for create/mkdir/mknod/symlink")
 		("single", "Run single-threaded")
 		("o", "Mount options (see mount.fuse(5) - only use if you know what "
 		      "you are doing)",
@@ -1527,6 +1646,7 @@ static cxxopts::ParseResult parse_options(int argc, char **argv)
 
 	fs.nosplice = options.count("nosplice") != 0;
 	fs.passthrough = options.count("nopassthrough") == 0;
+	fs.selinux = options.count("selinux") != 0;
 	fs.num_threads = options["num-threads"].as<int>();
 	fs.clone_fd = options.count("clone-fd");
 	fs.direct_io = options.count("direct-io");
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -198,6 +198,16 @@ static void list_add_req(struct fuse_req *req, struct fuse_req *next)
 	next->prev = req;
 }
 
+/*
+ * Convert extension length units to bytes.
+ * Extension lengths in fuse_in_header.total_extlen are specified in units
+ * of 8 bytes (sizeof(uint64_t)).
+ * TODO: Propose adding this to include/uapi/linux/fuse.h in the kernel.
+ */
+#ifndef FUSE_EXT_SIZE
+#define FUSE_EXT_SIZE(units) ((size_t)(units) * sizeof(uint64_t))
+#endif
+
 static void destroy_req(fuse_req_t req)
 {
 	if (req->flags.is_uring) {
@@ -213,6 +223,9 @@ void fuse_free_req(fuse_req_t req)
 {
 	int ctr;
 	struct fuse_session *se = req->se;
+
+	free(req->secctx);
+	req->secctx = NULL;
 
 	/* XXX: for now no support for interrupts with io-uring
 	 *      It actually might work already, though. But then would add
@@ -2798,7 +2811,8 @@ _do_init(fuse_req_t req, const fuse_ino_t nodeid, const void *op_in,
 			se->conn.capable_ext |= FUSE_CAP_OVER_IO_URING;
 		if (inargflags & FUSE_ALLOW_IDMAP)
 			se->conn.capable_ext |= FUSE_CAP_ALLOW_IDMAP;
-
+		if (inargflags & FUSE_SECURITY_CTX)
+			se->conn.capable_ext |= FUSE_CAP_SECURITY_CTX;
 	} else {
 		se->conn.max_readahead = 0;
 	}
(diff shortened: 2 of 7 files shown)


lib/fuse_lowlevel.c
```diff
@@ -3625,6 +3683,115 @@ fuse_req_opcode_sanity_ok(struct fuse_session *se, enum fuse_opcode in_op)
...
+		 * - 32+:  fuse_ext_header (type field)
+		 */
+		if (ext_hdr->type <= FUSE_MAX_NR_SECCTX && ext_hdr->type > 0) {
+			/* Found security context extension */
+			const struct fuse_secctx_header *sec_hdr =
+				(const struct fuse_secctx_header *)ext_curr;
+
+			req->secctx = malloc(sec_hdr->size);
+			if (req->secctx) {
+				memcpy(req->secctx, sec_hdr, sec_hdr->size);
```
kchen-ddn: Do we need memcpy for uring case?

lib/fuse_lowlevel.c
```diff
@@ -206,6 +206,7 @@ static void destroy_req(fuse_req_t req)
 	}
 	assert(req->ch == NULL);
 	pthread_mutex_destroy(&req->lock);
+	free(req->secctx);
```
bsbernd: Note that io-uring requests never go into this code path.

lib/fuse_lowlevel.c
```diff
@@ -2983,8 +2988,24 @@ _do_init(fuse_req_t req, const fuse_ino_t nodeid, const void *op_in,
...
+		fuse_log(FUSE_LOG_DEBUG, "   flags2=0x%08x\n", outarg.flags2);
+		fuse_log(FUSE_LOG_DEBUG, "   inargflags=0x%016llx\n",
+			 (unsigned long long)inargflags);
+		fuse_log(FUSE_LOG_DEBUG, "   outargflags=0x%016llx\n",
+			 (unsigned long long)outargflags);
+		fuse_log(FUSE_LOG_DEBUG, "   capable_ext=0x%016llx\n",
+			 (unsigned long long)se->conn.capable_ext);
+		fuse_log(FUSE_LOG_DEBUG, "   want_ext=0x%016llx\n",
+			 (unsigned long long)se->conn.want_ext);
+		fuse_log(FUSE_LOG_DEBUG,
```
bsbernd: Dunno, I think these are too many logs.

lib/fuse_lowlevel.c
bsbernd: I think in `fuse_free_req`

```
free(req->secctx);
req->secctx = NULL;
```

lib/fuse_lowlevel.c
```diff
@@ -3442,6 +3463,52 @@ const struct fuse_ctx *fuse_req_ctx(fuse_req_t req)
...
+	if (value)
+		*value = ctx_value;
+	if (value_len)
+		*value_len = fctx->size;
+
+	/* Advance iterator to next context (8-byte aligned) */
+	size_t ctx_size = sizeof(*fctx) + strlen(ctx_name) + 1 + fctx->size;
+	size_t aligned = (ctx_size + 7) & ~7;
+
+	req->secctx_iter_ptr += aligned;
```
bsbernd: `req->secctx_iter_ptr += FUSE_REC_ALIGN(sizeof(*fctx) + strlen(ctx_name) + 1 + fctx->size);`?

lib/fuse_lowlevel.c
```diff
@@ -3625,6 +3692,123 @@ fuse_req_opcode_sanity_ok(struct fuse_session *se, enum fuse_opcode in_op)
...
+			}
+			break; /* Found and copied security context */
+		}
+		/*
+		 * Currently we don't handle the extensions other than secctx,
+		 * so skip to next extension (8-byte aligned)
+		 */
+		size_t aligned_size = (ext_hdr.size + 7) & ~7;
+
+		ext_curr += aligned_size;
```
bsbernd: `ext_curr += FUSE_REC_ALIGN(ext_hdr.size);`?

lib/fuse_lowlevel.c
bsbernd: If you restrict secctx to `CREATE/MKNOD/MKDIR/SYMLINK` we might also go without an alloction for io-uring, because reply and copy into payload only comes after everything else - i.e. also after handling secctx. Although not entirely without risk.

lib/fuse_lowlevel.c
```diff
@@ -3442,6 +3463,52 @@ const struct fuse_ctx *fuse_req_ctx(fuse_req_t req)
...
+
+int fuse_req_secctx_next(fuse_req_t req, const char **name,
+			 const char **value, uint32_t *value_len)
+{
+	/* Check if we have more contexts to iterate */
+	if (req->secctx_iter_index >= req->secctx_count || !req->secctx_iter_ptr)
+		return -ENOENT;
+
+	/* Parse current security context */
+	const struct fuse_secctx *fctx = (const struct fuse_secctx *)req->secctx_iter_ptr;
```
bsbernd, line 3504: Could you add sanity checks here?

```
const char *buf_start = (const char *)req->secctx;
const char *buf_end = buf_start + req->secctx_len;

if (iter_ptr + sizeof(struct fuse_secctx) > buf_end) {
    fuse_log(FUSE_LOG_ERR, "fuse_secctx extends past buffer\n");
    return -EIO;
}
```

lib/fuse_lowlevel.c
```diff
@@ -3442,6 +3463,52 @@ const struct fuse_ctx *fuse_req_ctx(fuse_req_t req)
...
+	const char *ctx_name = (const char *)(fctx + 1);
+	const char *ctx_value = ctx_name + strlen(ctx_name) + 1;
```
bsbernd: For perfect safety maybe `strnlen(ctx_name, buf_end - ctx_name)`?

lib/fuse_lowlevel.c
```diff
@@ -3625,6 +3692,123 @@ fuse_req_opcode_sanity_ok(struct fuse_session *se, enum fuse_opcode in_op)
...
+	 * The extension header is polymorphic:
+	 * - struct { size, nr_secctx } if nr_secctx <= 31 (security context)
+	 * - struct { size, type } if type >= 32 (other extensions)
+	 *
+	 * total_extlen is in 8-byte units and gives the extension size.
+	 *
+	 * We must copy the security context because the source buffer
+	 * will be freed after the operation handler returns.
+	 */
+	size_t ext_bytes = (size_t)total_extlen * 8;
```
bsbernd: Could you define a macro like
```
define FUSE_EXT_SIZE(units) ((size_t)(units) * sizeof(uint64_t))
```

And we can propose this to be added to `include/uapi/linux/fuse.h`

lib/fuse_lowlevel.c
```diff
@@ -3625,6 +3692,123 @@ fuse_req_opcode_sanity_ok(struct fuse_session *se, enum fuse_opcode in_op)
...
+	const char *ext_curr = ext_start;
+
+	/*
+	 * Iterate through all extensions to find security context.
+	 * Use memcpy to avoid unaligned access - extensions may not be
+	 * aligned within the buffer.
+	 */
+	while (ext_curr + sizeof(struct fuse_ext_header) <= ext_end) {
+		struct fuse_ext_header ext_hdr;
+		uint32_t ext_type;
```
bsbernd: I think would be easier to read as `second_field`

bsbernd:
@kchen-ddn Could you add an example how to use it? Ideally in passthrough_hp.cc, so that it gets tested during xfstests?

bsbernd:
@kchen-ddn could you please review https://github.com/libfuse/libfuse/pull/1509 ?

lib/fuse_lowlevel.c
kchen-ddn: Thanks, now it's freed in fuse_free_req(). I haven't skipped the extra allocation for io-uring case for safety.

lib/fuse_lowlevel.c
kchen-ddn: Removed.

lib/fuse_lowlevel.c
kchen-ddn: Done.

lib/fuse_lowlevel.c
kchen-ddn: Done.

lib/fuse_lowlevel.c
kchen-ddn: Done.

lib/fuse_lowlevel.c
kchen-ddn: Done.

lib/fuse_lowlevel.c
kchen-ddn: The macro definition has been added in fuse_lowlevel.c for now.

lib/fuse_lowlevel.c
kchen-ddn: Changed it.

kchen-ddn:
> @kchen-ddn Could you add an example how to use it? Ideally in passthrough_hp.cc, so that it gets tested during xfstests?

I added some code to iterate the secctxs in passsthrough_hp.cc. And for security label to exists, we need to support SELinux, hence the SYNC init is also needed. So the fuse_session_want_sync_init needs to be called, so I added it to the versionscript file.

But I'm not sure if the passthrough_hp file system really need the added handling of security label. Seems that when creating the file in the dest file system, the kernel has already set the initial label of the new file.

example/passthrough_hp.cc
```diff
@@ -476,6 +476,64 @@ static void sfs_lookup(fuse_req_t req, fuse_ino_t parent, const char *name)
...
+ * This demonstrates the security context extension API:
+ * - fuse_req_secctx_count() - get number of security contexts
+ * - fuse_req_secctx_reset() - reset iterator
+ * - fuse_req_secctx_next() - iterate through contexts
+ *
+ * @param req FUSE request containing security context extension
+ * @param parent_fd Parent directory file descriptor
+ * @param name The newly created file name
+ */
+static void apply_selinux_context(fuse_req_t req, int parent_fd, const char *name)
```
bsbernd: `sfs_apply_selinux_context`

example/passthrough_hp.cc
```diff
@@ -476,6 +476,64 @@ static void sfs_lookup(fuse_req_t req, fuse_ino_t parent, const char *name)
...
+				fuse_log(FUSE_LOG_DEBUG,
+					 "apply_selinux_context: Set %s=%s on %s\n",
+					 ctx_name, ctx_value, name);
+			}
+			break;
+		}
+	}
+	fuse_req_secctx_reset(req);
+}
+#endif
```
bsbernd, line 572: Can you just add a `#else condition` condition here

```
#else
static void apply_selinux_context(fuse_req_t req, int parent_fd, const char *name)
{
     (void)req;
     (void)fd;
     (void)name;
}
#endif
```

example/passthrough_hp.cc
```diff
@@ -493,6 +551,11 @@ static void mknod_symlink(fuse_req_t req, fuse_ino_t parent, const char *name,
 	if (res == -1)
 		goto out;
 
+#ifdef HAVE_SETXATTR
```
bsbernd: And then these #ifdefs are not needed anymore

example/passthrough_hp.cc
```diff
@@ -951,6 +1014,11 @@ static void sfs_create(fuse_req_t req, fuse_ino_t parent, const char *name,
 		return;
 	}
 
+#ifdef HAVE_SETXATTR
```
bsbernd: And here.

example/passthrough_hp.cc
```diff
@@ -476,6 +476,64 @@ static void sfs_lookup(fuse_req_t req, fuse_ino_t parent, const char *name)
...
+
+	if (secctx_count == 0)
+		return;
+
+	/* Iterate through security contexts looking for security.selinux */
+	const char *ctx_name;
+	const char *ctx_value;
+	uint32_t ctx_len;
+
+	fuse_req_secctx_reset(req);
```
bsbernd: So here the API is weird - why a reset at the beginning?

lib/fuse_lowlevel.c
```diff
@@ -2843,6 +2857,7 @@ _do_init(fuse_req_t req, const fuse_ino_t nodeid, const void *op_in,
 	LL_SET_DEFAULT(se->op.readdirplus && se->op.readdir,
 		       FUSE_CAP_READDIRPLUS_AUTO);
 	LL_SET_DEFAULT(1, FUSE_CAP_OVER_IO_URING);
+	LL_SET_DEFAULT(1, FUSE_CAP_SECURITY_CTX);
```
bsbernd: I don't think we want to enable this by default, as it causes kernel overhead - file systems that want it, need to enable themselves.

lib/fuse_lowlevel.c
```diff
@@ -3625,6 +3701,121 @@ fuse_req_opcode_sanity_ok(struct fuse_session *se, enum fuse_opcode in_op)
...
+ *
+ * @param req The FUSE request to populate with security context
+ * @param total_extlen Extension length in 8-byte units from the FUSE input header
+ * @param ext_buffer The buffer containing the extensions (for io_uring: op_payload,
+ *                   for non-io_uring: in header buffer)
+ * @param buffer_len The length of ext_buffer (for io_uring: payload_len,
+ *                   for non-io_uring: in->len)
+ */
+static void
+fuse_parse_extensions(struct fuse_req *req, uint32_t total_extlen,
```
bsbernd: fuse_req_parse_extensions()

lib/fuse_lowlevel.c
```diff
@@ -3625,6 +3701,121 @@ fuse_req_opcode_sanity_ok(struct fuse_session *se, enum fuse_opcode in_op)
...
+{
+
+	/* Initialize security context fields */
+	req->secctx = NULL;
+	req->secctx_len = 0;
+	req->secctx_count = 0;
+	req->secctx_iter_index = 0;
+	req->secctx_iter_ptr = NULL;
+
+	if (total_extlen == 0)
```
bsbernd, line 3743: Should be in the caller.

lib/fuse_lowlevel.c
```diff
@@ -3745,8 +3938,16 @@ void fuse_session_process_buf_internal(struct fuse_session *se,
 	}
 
 	fuse_session_in2req(req, in);
+	fuse_parse_extensions(req, in->total_extlen, in, in->len);
```
bsbernd:
```
if (in->total_extlen)
    fuse_parse_extensions()
``

File: lib/fuse_lowlevel.c [hunk 19]
```diff
@@ -3823,7 +4024,16 @@ void fuse_session_process_uring_cqe(struct fuse_session *se,
 {
 	int err;
 
+	/* For io_uring, extensions are in the payload buffer, not appended to 'in' header. */
 	fuse_session_in2req(req, in);
+	fuse_parse_extensions(req, in->total_extlen, op_payload, payload_len);
```
bsbernd: Here as well.

lib/fuse_lowlevel.c
```diff
@@ -3625,6 +3701,121 @@ fuse_req_opcode_sanity_ok(struct fuse_session *se, enum fuse_opcode in_op)
...
+			}
+			break; /* Found and copied security context */
+		}
+		/*
+		 * Currently we don't handle the extensions other than secctx,
+		 * so skip to next extension (8-byte aligned)
+		 */
+		ext_curr += FUSE_REC_ALIGN(ext_hdr.size);
+	}
+}
```
bsbernd, line 3832: I think `fuse_req_secctx_reset` should be called at  the end of this function

bsbernd: Or better above in the if condition that copies.

lib/fuse_lowlevel.c
```diff
@@ -3442,6 +3459,65 @@ const struct fuse_ctx *fuse_req_ctx(fuse_req_t req)
...
+	/* Return the parsed data */
+	if (name)
+		*name = ctx_name;
+	if (value)
+		*value = ctx_value;
+	if (value_len)
+		*value_len = fctx->size;
+
+	/* Advance iterator to next context (8-byte aligned) */
+	req->secctx_iter_ptr += FUSE_REC_ALIGN(sizeof(*fctx) + strlen(ctx_name) + 1 + fctx->size);
```
bsbernd: Using strlen here confuses static analyzers, because of strnlen() above.

Above just

size_t max_strlen = buf_end - ctx_name;

And then use that in strnlen() above and here?

example/passthrough_hp.cc
```diff
@@ -951,6 +1014,11 @@ static void sfs_create(fuse_req_t req, fuse_ino_t parent, const char *name,
 		return;
 	}
 
+#ifdef HAVE_SETXATTR
+	/* Apply SELinux security context if present in FUSE extension */
+	apply_selinux_context(req, inode_p.fd, name);
+#endif
```
bsbernd: Could you also add to sfs_tmpfile()?

example/passthrough_hp.cc
```diff
@@ -476,6 +476,64 @@ static void sfs_lookup(fuse_req_t req, fuse_ino_t parent, const char *name)
...
+	fuse_req_secctx_reset(req);
+	while (fuse_req_secctx_next(req, &ctx_name, &ctx_value, &ctx_len) == 0) {
+		if (strcmp(ctx_name, "security.selinux") == 0) {
+			/*
+			 * Set SELinux label on the newly created file.
+			 * Use fsetxattr with AT_SYMLINK_NOFOLLOW to set xattr relative to parent_fd.
+			 * On Linux, we can use setxattr with the path constructed via /proc/self/fd
+			 */
+			char procpath[64];
+			sprintf(procpath, "/proc/self/fd/%i/%s", parent_fd, name);
```
bsbernd: snprintf

example/passthrough_hp.cc
kchen-ddn: done

example/passthrough_hp.cc
kchen-ddn: done

example/passthrough_hp.cc
kchen-ddn: done

example/passthrough_hp.cc
kchen-ddn: done

example/passthrough_hp.cc
kchen-ddn: The reason I do the reset here, is just to make sure the iteration starts from begining, in case somewhere else called fuse_req_secctx_next() without  fuse_req_secctx_reset() at the end.

lib/fuse_lowlevel.c
kchen-ddn: Good point. Changed it.

kchen-ddn:
> > @kchen-ddn Could you add an example how to use it? Ideally in passthrough_hp.cc, so that it gets tested during xfstests?
> 
> I added some code to iterate the secctxs in passsthrough_hp.cc. And for security label to exists, we need to support SELinux, hence the SYNC init is also needed. So the fuse_session_want_sync_init needs to be called, so I added it to the versionscript file.
> 
> But I'm not sure if the passthrough_hp file system really need the added handling of security label. Seems that when creating the file in the dest file system, the kernel has already set the initial label of the new file.

I have added some comment to passthrough_hp.cc:

* NOTE: This is primarily a DEMONSTRATION of the FUSE security context extension API.
 * For passthrough_hp specifically, this is not strictly necessary because when we create
 * files on the underlying filesystem (via mknodat/openat/etc.), the Linux kernel's VFS
 * layer already ensures files get the correct SELinux labels based on the creating
 * process's context, parent directory context, and SELinux policy transition rules.
 *
 * However, this example is valuable for:
 * - Demonstrating the security context extension API for filesystem developers
 * - Providing a reference implementation for filesystems that DO need explicit context
 *   handling (e.g., network filesystems, FUSE filesystems backed by non-Linux storage)

lib/fuse_lowlevel.c
kchen-ddn: The logic is actually in that 'if' branch, but yes, calling the reset function directly is more clear.

```
-                               /* Initialize iterator to point after the header */
-                               req->secctx_iter_index = 0;
-                               req->secctx_iter_ptr = (const char *)req->secctx +
-                                       sizeof(struct fuse_secctx_header);
+                               /* Initialize iterator */
+                               fuse_req_secctx_reset(req);
```

lib/fuse_lowlevel.c
kchen-ddn: I'll make the max_strlen = buf_end - ctx_name -1; given that the strlen doesn't count the '\0' at the end.

```
@@ -3502,9 +3502,16 @@ int fuse_req_secctx_next(fuse_req_t req, const char **name,

        /* Parse current security context */
        const struct fuse_secctx *fctx = (const struct fuse_secctx *)req->secctx_iter_ptr;
-       const char *ctx_name = (const char *)(fctx + 1);

-       const char *ctx_value = ctx_name + strnlen(ctx_name, buf_end - ctx_name) + 1;
+       /* empty security context is invalid */
+       if (fctx->size == 0) {
+               fuse_log(FUSE_LOG_ERR, "secctx has zero size\n");
+               return -EIO;
+       }
+
+       const char *ctx_name = (const char *)(fctx + 1);
+       size_t max_strlen = buf_end - ctx_name - 1;
+       const char *ctx_value = ctx_name + strnlen(ctx_name, max_strlen) + 1;

        /* Validate that value fits in buffer */
        if (ctx_value + fctx->size > buf_end) {
@@ -3521,7 +3528,7 @@ int fuse_req_secctx_next(fuse_req_t req, const char **name,
                *value_len = fctx->size;

        /* Advance iterator to next context (8-byte aligned) */
-       req->secctx_iter_ptr += FUSE_REC_ALIGN(sizeof(*fctx) + strlen(ctx_name) + 1 + fctx->size);
+       req->secctx_iter_ptr += FUSE_REC_ALIGN(sizeof(*fctx) + strnlen(ctx_name, max_strlen) + 1 + fctx->size);
        req->secctx_iter_index++;
```

PR #1503, opened 2026-05-20
fixes for synchronous FUSE_INIT

lostjeffle:
several fixes for the synchronous FUSE_INIT feature.

### DIFF
--- a/lib/mount_fsmount.c
+++ b/lib/mount_fsmount.c
@@ -407,16 +407,6 @@ int fuse_kern_fsmount(const char *mnt, unsigned long flags, int blkdev,
 		goto out_free;
 	}
 
-	/* Apply mtab options (overlap with kernel_opts is filtered and tolerated) */
-	err = apply_mount_opts(fsfd, mtab_opts);
-	if (err < 0) {
-		log_fsconfig_kmsg(fsfd);
-		fprintf(stderr,
-			"fuse: failed to apply mtab options '%s'\n",
-			mtab_opts);
-		goto out_free;
-	}
-
 	/* Create the filesystem instance */
 	res = fsconfig(fsfd, FSCONFIG_CMD_CREATE, NULL, NULL, 0);
 	if (res == -1) {
--- a/lib/mount_fsmount.c
+++ b/lib/mount_fsmount.c
@@ -421,9 +421,8 @@ int fuse_kern_fsmount(const char *mnt, unsigned long flags, int blkdev,
 	flags = ms_flags_to_mount_attrs(flags, &mount_attrs);
 	if (flags != 0) {
 		/* unsupported flags found, either fsconfig or mount attr flags  */
-		//fuse_log(FUSE_LOG_ERR, "Unsupported mount flags found: %d", flags);
-		errno = -ENOTSUP;
-		return -1;
+		errno = ENOTSUP;
+		goto out_free;
 	}
 
 	/* Create mount object with mount attributes */
--- a/lib/mount_i_linux.h
+++ b/lib/mount_i_linux.h
@@ -11,6 +11,7 @@
 #define FUSE_MOUNT_I_LINUX_H_
 
 #include <sys/mount.h>
+#include <linux/mount.h>
 
 struct fuse_args;
 


lib/mount_util.c
```diff
@@ -31,6 +31,7 @@
 #endif
 #include <sys/stat.h>
 #include <sys/wait.h>
+#include <linux/mount.h>
```
bsbernd: @lostjeffle Could you move this up into the `#ifdef __linux__` condition?

bsbernd: Also, are you sure it is needed, not handled by `#include "fuse_mount_compat.h"`, which has `#include <sys/mount.h>`. Maybe we make an explicit include of `sys/mount.h here`, instead of linux/mount.h?

example/passthrough_hp.cc
```diff
@@ -1649,6 +1649,8 @@ int main(int argc, char *argv[])
 	if (fs.bg_debug_fuse)
 		fuse_session_set_debug(se);
 
+	fuse_session_want_sync_init(se);
```
bsbernd: Uh, the bug is in the condition, not here `if (!fuse_daemonize_is_used() || !se->want_sync_init) {` - this has to be `&&`. And we need to rename to `fuse_daemonize_early_is_used()`

bsbernd: Hmm, well, maybe `fuse_daemonize_early`, should set `se->want_sync_init` on its own, so that the daemon can override.

lib/mount_util.c
lostjeffle: > Also, are you sure it is needed, not handled by `#include "fuse_mount_compat.h"`, which has `#include <sys/mount.h>`. Maybe we make an explicit include of `sys/mount.h here`, instead of linux/mount.h?

In my building environment (centos like distro), "MOUNT_ATTR_*" macros are defined in linux/mount.h rather than sys/mount.h.

Or maybe include <inux/mount.h> inside mount_i_linux.h?

example/passthrough_hp.cc
```diff
@@ -1679,9 +1679,6 @@ int main(int argc, char *argv[])
 
 	fuse_session_unmount(se);
 
-err_out4:
-	if (fuse_daemonize_early_is_active())
-		fuse_daemonize_early_fail(ret);
```
bsbernd, line 1682: I will check later, but that is not supposed to happen, I would prefer if applications wouldn't need these many goto jumps - `fuse_daemonize_early_is_active()` needs to act on `daemonize.active` only. We need to remove that `daemonize.daemonized` condtion in _is_active, but need to check first if something currently relies on it.

lib/mount_util.c
bsbernd: Ah interesting that `sys/mount.h` doesn't include it, I'm tempted to argue that is a bug there :) Yes, please add to `mount_i_linux.h`

example/passthrough_hp.cc
lostjeffle: Yes, it would be better if daemonize.daemonized can be remove from fuse_daemonize_early_is_active().

lib/mount_util.c
bsbernd: I guess the BSD failure is because of this unconditional include.

example/passthrough_hp.cc
lostjeffle: From the semantics of what the function does, the current fuse_daemonize_early_is_active() would be better to be renamed to fuse_daemonize_is_used() (i.e. `daemonize.daemonized || daemonize.active`), while the current fuse_daemonize_is_used() be renamed to fuse_daemonize_early_is_active() (i.e. `daemonize.active`)

bsbernd: Yeah, I will check in about an hour if we need `fuse_daemonize_is_used()` at all

bsbernd:
@lostjeffle Sorry, didn't come to it all day :/ I will try in the morning.

lostjeffle:
changes since last version:
- "lib mount: include linux/mount.h in mount_i_linux.h" move "linux/mount.h" to mount_i_linux.h to avoid building failure  on BSD

example/passthrough_hp.cc
lostjeffle: Hi @bsbernd ,sorry for the gentle ping.  Is there any comment on this?

bsbernd:
@lostjeffle Really sorry, didn't get to it until today. Please see here https://github.com/libfuse/libfuse/pull/1505

There is a critical fix for a buffer overflow in fusermount.c included - I'm going to merge once tests pass. If there are any issues, please let me know and I will fix after. I very much hope that your changes to `passthrough_hp.cc` can be dropped with that PR.

lostjeffle:
> @lostjeffle Really sorry, didn't get to it until today. Please see here #1505
> 
> There is a critical fix for a buffer overflow in fusermount.c included - I'm going to merge once tests pass. If there are any issues, please let me know and I will fix after. I very much hope that your changes to `passthrough_hp.cc` can be dropped with that PR.

No worries.  I would rebase #1505 and test if the issue is gone.  Thanks for fixing it!

lostjeffle:
Hi @bsbernd   I have rebased https://github.com/libfuse/libfuse/pull/1505 and it looks good to me.  Would you please take a look at the remaining patches?

bsbernd (approved):
Thanks @lostjeffle !

PR #1508, opened 2026-05-29
fuse: add submount capability and lowlevel attr_flags

ukernel:
Add submount capability negotiation and lowlevel entry attr_flags so filesystems can mark lookup/create results as submounts.

- add FUSE_CAP_SUBMOUNTS in fuse_common.h and wire it into init capability negotiation
- add attr_flags to struct fuse_entry_param
- add FUSE_LL_ATTR_SUBMOUNT in fuse_lowlevel.h
- map FUSE_LL_ATTR_SUBMOUNT to FUSE_ATTR_SUBMOUNT in entry reply paths

To avoid breaking older callers that were built without attr_flags in struct fuse_entry_param, libfuse only reads e->attr_flags when the session reports a caller libfuse version >= 3.20 and the submount capability is enabled. Older callers therefore continue on the legacy path and never access the new field.

### DIFF
--- a/include/fuse_lowlevel.h
+++ b/include/fuse_lowlevel.h
@@ -102,6 +102,18 @@ struct fuse_entry_param {
 	    that come through the kernel, this should be set to a very
 	    large value. */
 	double entry_timeout;
+
+	/** Entry attribute flags. See enum fuse_attr_flags below. */
+	uint32_t attr_flags;
+};
+
+/**
+ * Inode attribute flags
+ *
+ * FUSE_LL_ATTR_SUBMOUNT: Inode is a submount root
+ */
+enum fuse_attr_flags {
+	FUSE_LL_ATTR_SUBMOUNT = (1 << 0),
 };
 
 /**
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -465,8 +465,19 @@ static unsigned int calc_timeout_nsec(double t)
 		return (unsigned int) (f * 1.0e9);
 }
 
+static inline bool fuse_entry_param_has_attr_flags(struct fuse_session *se)
+{
+	const struct libfuse_version *v = &se->version;
+	if ((v->major > 3) || (v->major == 3 && v->minor >= 20)) {
+		/* if any enabled capability uses attr_flags, return true */
+		if (se->conn.want_ext & FUSE_CAP_SUBMOUNTS)
+			return true;
+	}
+	return false;
+}
+
 static void fill_entry(struct fuse_entry_out *arg,
-		       const struct fuse_entry_param *e)
+		       const struct fuse_entry_param *e, bool has_attr_flags)
 {
 	arg->nodeid = e->ino;
 	arg->generation = e->generation;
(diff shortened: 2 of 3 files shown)


include/fuse_lowlevel.h
```diff
@@ -102,6 +102,18 @@ struct fuse_entry_param {
 	    that come through the kernel, this should be set to a very
 	    large value. */
 	double entry_timeout;
+
+	/** Entry attribute flags. See enum fuse_attr_flags below. */
+	uint32_t attr_flags;
+};
```
bsbernd, line 108: I'm afraid it is not possible in this form. Please take a look at all example/ file systems - they all define fuse_entry_param. I.e. this causes total ABI breakage. It also causes API breakage if fuse_entry_param is not entirely zeroed.

lib/fuse_lowlevel.c
```diff
@@ -465,8 +465,19 @@ static unsigned int calc_timeout_nsec(double t)
 		return (unsigned int) (f * 1.0e9);
 }
 
+static inline bool fuse_entry_param_has_attr_flags(struct fuse_session *se)
+{
+	const struct libfuse_version *v = &se->version;
+	if ((v->major > 3) || (v->major == 3 && v->minor >= 20)) {
```
bsbernd, line 471: Unfortunately not enough, what we here need it FUSE_USE_VERSION in libfuse_version. I had started to do it in this PR https://github.com/libfuse/libfuse/pull/1382/commits (see the tmp commit), but due to ABI compatibility it got really complex. There might be a a more easy approach by using the existing `uint32_t padding`.

ukernel: Thanks for your review.  If fuse_reply_entry_3xx with user_apiabi_version is implemented (compatibility is properly handled according to the user_apiabi_version).  Is 'adding attr_flags to fuse_entry_param'  acceptable?

bsbernd: I'm going to give adding FUSE_USE_VERSION to struct fuse_session another try over the weekend. If an application set that version we can at least hope that the developer has read the corresponding changelog entry. However, this needs to supported by upstream fuse (and not only for virtiofs) and we need a test. And maybe there is also another solution than changing a very common struct.

bsbernd:
@ukernel After looking in detail, I think your current approach is too risky. There is no way that all file systems on this world will ever catch up zero initializing `struct fuse_entry_param`. At least there is no way libfuse can rely on it. I don't think that struct has changed in the last 25 years and there is no enforcement that the struct gets memset to 0 or compiler initialized to 0. Extending the struct would bear the risk that a random file system might just set all fields and the developer might not have have read the changelog for FUSE_USE_VERSION API changes.

Here is my suggestion:

1) Introduce  `struct fuse_entry_param2`
2) Extend `struct fuse_req` with `struct fuse_entry_param2`
3) Add new functions `fuse_req_set_entry_param2(fuse_req_t req, fuse_node_t node_id, uint64_t generation, struct statx attr, struct timespec attr_timeout, struct timespec entry_timeout);` and `fuse_req_set_submount`
4) Add `fuse_reply_entry2`, `fuse_reply_create2`, with only `struct fuse_req` as parameter.
5) readdirplus would call `fuse_req_set_submount`. We can also add `fuse_add_direntry_plus2` for it to allow setting timeouts as `struct timespec`. Debatable if it should take plain `struct stat` or `struct statx`. I'm more for statx, even if there is no way to send full statx to kernel yet.

Also note statx in fuse_req_set_entry_param2 - that is to repare the lookupx call.

If you agree with the approach we need to send out a proposal to `fuse-devel@lists.linux.dev`, because there is a bit design change involved.

ukernel:
thanks for your suggestion. I will try implementing it.

PR #1511, opened 2026-06-03
Update homepage at man page

wilkis3:


bsbernd:
@wilkis3 could you change the subject to "man page"?

PR #1520, opened 2026-06-12
fuse-io-uring: fix req_header_sz size calculation

lostjeffle:
The header buffer is described by "struct fuse_uring_req_header" rather than "struct fuse_ring_ent".


bsbernd:
Thank you @lostjeffle !

PR #1522, opened 2026-06-15
io uring teardown fixes

bsbernd:


bsbernd:
@abhinavagarwal07 Did you have a chance to look at these fixes?

agabhin:
@bsbernd I’ll review today

lib/fuse_uring.c
```diff
@@ -389,24 +395,109 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
...
+	       (now.tv_sec == deadline->tv_sec && now.tv_nsec >= deadline->tv_nsec);
+}
+
+/* Number of entries the application still holds (ref_cnt > 1). */
+static size_t fuse_uring_queue_held(struct fuse_ring_queue *queue, size_t depth)
+{
+	size_t held = 0;
+
+	for (size_t idx = 0; idx < depth; idx++)
+		if (atomic_load(&queue->ent[idx].req.ref_cnt) > 1)
```
abhinavagarwal07, line 454: `ref_cnt == 1` isn't a safe "idle" signal. In `fuse_free_req()`'s io-uring branch, `--req->ref_cnt` (2→1) is published *before* the trailing `fuse_chan_put(req->ch)` / `req->ch = NULL`. So teardown can read 1, free `queues`, and race the tail of `fuse_free_req()` — a UAF on the entry, so likely a blocker. Could teardown gate on a flag set after the reply path's last access, instead of the refcount?

lib/fuse_uring.c
```diff
@@ -420,13 +511,20 @@ static void fuse_session_destruct_uring(struct fuse_ring_pool *fuse_ring)
...
 
 		pthread_mutex_destroy(&queue->ring_lock);
 	}
 
+	/*
+	 * The queues live in one allocation that a detached thread may still
+	 * touch, so leak the backing pool too if anything was abandoned.
+	 */
+	if (leaked)
+		return;
```
abhinavagarwal07: This leaks the ring but `se` can still be freed:
held reply → teardown timeout → `if (leaked) return` → `fuse_uring_stop()` returns 0 → `fuse_session_destroy()` frees `se` → late reply / detached thread derefs `ring_pool->se` / `req->se` → UAF.

I think the leaked state has to propagate to session teardown so `se` is kept alive too — returning an error won't help unless `fuse_session_destroy()` honors it.

lib/fuse_uring.c
```diff
@@ -389,24 +395,109 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
...
+	 */
+	for (size_t qid = 0; qid < fuse_ring->nr_queues; qid++) {
+		struct fuse_ring_queue *queue =
+			fuse_uring_get_queue(fuse_ring, qid);
+		uint64_t value = 1ULL;
+
+		if (queue->tid == 0)
+			continue;
+
+		if (write(queue->eventfd, &value, sizeof(value)) != sizeof(value))
```
abhinavagarwal07: Minor/defensive: retry on `EINTR`? A dropped wake here becomes a 30s timeout + leak, so the retry loop looks like cheap insurance.

abhinavagarwal07: Phase-1's eventfd wake only reaches threads already in `io_uring_submit_and_wait()`. A thread parked in `sem_wait(&init_sem)` (before the main loop) never sees it — and the startup error path tears down before `fuse_uring_wake_ring_threads()` runs, so those threads block until the timeout and get leaked.

A `sem_post` per queue wakes them, but isn't enough alone: `mt_exited` is still false on that path, so a woken thread enters the ring loop instead of exiting. Would setting the exit flag before waking both the sem and io_uring waiters cover it?

test/test_deferred_reply_teardown.c
```diff
@@ -0,0 +1,237 @@
...
+	/* Deferred reply: drops ref_cnt 2 -> 1 so the drain can complete. */
+	fuse_reply_err(dt.deferred_req, ENOENT);
+	return NULL;
+}
+
+static void watchdog_handler(int sig)
+{
+	(void)sig;
+
+	fprintf(stderr, "Test FAILED: teardown did not finish (deadlock?)\n");
```
abhinavagarwal07: The handler prints and returns, so it doesn't actually fail fast. And `fprintf` isn't async-signal-safe in a handler — better to `write()` then `_exit(1)`.

abhinavagarwal07 (commented):
lgtm directionally, please see some concerns in the inline comments.

agabhin:
I forgot to switch account and posted comments via my personal github @abhinavagarwal07

bsbernd:
@abhinavagarwal07 Thanks for review, I will get it addresses ASAP - was busy with other work.

test/test_deferred_reply_teardown.c
bsbernd: This is actually auto-solved now, the new tests I had to add to catch some bugs that were really hard to get by with pytest also has per test timeouts - no need to handle that in tests anymore.

lib/fuse_uring.c
```diff
@@ -897,7 +1065,8 @@ static void *fuse_uring_thread(void *arg)
 		io_uring_submit(&queue->ring);
 
 	/* Not using fuse_session_exited(se), as that cannot be inlined */
-	while (!atomic_load_explicit(&se->mt_exited, memory_order_relaxed)) {
+	while (!atomic_load_explicit(&se->mt_exited, memory_order_relaxed) &&
+	       !atomic_load_explicit(&ring_pool->stopping, memory_order_relaxed)) {
```
abhinavagarwal07, line 1069: Could the ring-thread exit path clear `cqe_processing`? When the teardown eventfd makes
`fuse_uring_queue_handle_cqes()` return `-ENOTCONN`
([:874](https://github.com/libfuse/libfuse/blob/35a1e7d3a29b4a96bbb83db1439f63f10628ec9b/lib/fuse_uring.c#L874)),
the thread takes the `goto err` at :1098 (label at :1116) and skips the clear at :1100.
A deferred reply's `commit_sqe()` then reads the stale `true` at
[:229](https://github.com/libfuse/libfuse/blob/35a1e7d3a29b4a96bbb83db1439f63f10628ec9b/lib/fuse_uring.c#L229)
and skips the submit — and no ring thread is left to flush the batch, so the SQE is
discarded by `io_uring_queue_exit()`. `fuse_free_req()` drops `ref_cnt` either way, so the
drain sees the entry idle and teardown reports success while the client stays blocked.

I believe this is what the [io-uring CI failures](https://github.com/libfuse/libfuse/actions/runs/33690808918/job/100448875371)
are — the captured stacks show the daemon's main thread in `pthread_join(client_thread)`
and the client still in `request_wait_answer` / `fuse_lookup_name`. Closing the ring fd
doesn't release it: an entry already committed to userspace stays pending until the
*connection* is aborted, and unmount sits behind that join.

The `stopping` check on this line only helps a thread that isn't parked in the wait,
which is where teardown finds it.

Would clearing the flag on `err:` / `err_non_fatal:` be enough, or would you rather have
`commit_sqe()` submit unconditionally when `stopping` is set? The latter doesn't depend on
which exit the thread took.

lib/fuse_uring.c
```diff
@@ -419,24 +436,159 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
...
-
-			rc = write(queue->eventfd, &value, sizeof(value));
-			if (rc != sizeof(value))
-				fprintf(stderr,
-					"Wrote to eventfd=%d err=%s: rc=%d\n",
-					queue->eventfd, strerror(errno), rc);
-			pthread_cancel(queue->tid);
-			pthread_join(queue->tid, NULL);
-			queue->tid = 0;
+			pthread_detach(queue->tid);
```
abhinavagarwal07, line 579: What keeps the application's callback state alive after this detach? The session refcount
covers `se`, but `fuse_session_loop_mt()` can return while this thread is still inside a
`fuse_lib_*` callback, and
[`fuse_destroy()`](https://github.com/libfuse/libfuse/blob/35a1e7d3a29b4a96bbb83db1439f63f10628ec9b/lib/fuse.c#L5321)
then destroys `f->lock` and frees `f->fs` and `f`. If the thread resumes afterwards it's
working on freed memory, and no session reference can prevent that.

Should the loop avoid returning while a callback can still resume, or is the intent that
crossing the deadline is already a "process is broken" state?

lib/fuse_lowlevel.c
```diff
@@ -4318,6 +4329,37 @@ void fuse_session_destroy(struct fuse_session *se)
...
+	}
+
+	fuse_session_free(se);
+}
+
+void fuse_session_destroy(struct fuse_session *se)
+{
+	fuse_ll_clear_pipe(se);
+	se->destroy_called = true;
+	fuse_session_put(se);
```
abhinavagarwal07, line 4360: This can now return before `fuse_session_free()` runs, which moves the lifetime contract
a caller sees. With a request still outstanding, `.destroy(se->userdata)` at
[:4299](https://github.com/libfuse/libfuse/blob/35a1e7d3a29b4a96bbb83db1439f63f10628ec9b/lib/fuse_lowlevel.c#L4299)
moves to whichever thread later drops the last ref — potentially after the caller has
already released that userdata. `fuse_destroy()` frees `f` immediately after this returns.

This isn't a corner case for `fuse_main()` users: the loop stops reading `/dev/fuse`
before `fuse_unmount()` ([helper.c:522](https://github.com/libfuse/libfuse/blob/35a1e7d3a29b4a96bbb83db1439f63f10628ec9b/lib/helper.c#L522))
generates FUSE_DESTROY, so nobody consumes it, `got_destroy` stays 0, and this has always
been where `.destroy` actually gets delivered. The PR makes exactly that call deferrable.

The header also documents that a custom session fd "will be closed when
`fuse_session_destroy()` is called", which deferring changes observably.

Separately, should `se->interrupts` be drained here? A FUSE_INTERRUPT whose target isn't
on `se->list` parks there via `_do_interrupt()` (:2362) holding the ref
`fuse_ll_alloc_req()` took, and is only reaped when a later non-interrupt `/dev/fuse`
request pops it in `check_interrupt()` (:2386-2390). Nothing drains the list at destroy.
Pre-PR that was a leaked `malloc`; now it pins the session, so `destroy_called` is set but
the count never reaches 0 — `se` is never freed and `.destroy(userdata)` never runs, with
no log line, since the "released before destroy" warning only fires when the count hits 0.

That looks likeliest on uring sessions: interrupts arrive over `/dev/fuse` while their
targets live on the ring and never on `se->list`, so `find_interrupted()` always misses and
every interrupt parks until some other `/dev/fuse` request happens along.

Adding `fuse_ll_clear_pipe()` here reads like you hit the same shape for the TLS pipe —
could the destroy callback stay synchronous even if freeing `se` is deferred?

lib/fuse_uring.c
```diff
@@ -419,24 +436,159 @@ static int fuse_queue_setup_io_uring(struct io_uring *ring, size_t qid,
...
+			fuse_log(FUSE_LOG_ERR,
+				 "qid=%d ring thread stuck in application; leaking queue\n",
+				 queue->qid);
+			continue;
+		}
+
+		held = fuse_uring_queue_held(queue, fuse_ring->queue_depth);
+		if (held) {
```
abhinavagarwal07, line 587: Is the queue meant to stay leaked after the held request eventually replies?
`fuse_uring_pool_put()` only runs from this function
([:610](https://github.com/libfuse/libfuse/blob/35a1e7d3a29b4a96bbb83db1439f63f10628ec9b/lib/fuse_uring.c#L610),
[:619](https://github.com/libfuse/libfuse/blob/35a1e7d3a29b4a96bbb83db1439f63f10628ec9b/lib/fuse_uring.c#L619)),
which has already passed, so a reply landing just past the deadline drops its request and
session refs but not the queue's pool ref — retaining the ring fd, mappings, pool, and the
session for the life of the process.

Could the last late reply trigger deferred cleanup instead — whichever of "thread exited"
and "held count reached zero" happens last drops the ref?

test/test_deferred_reply_teardown.c
```diff
@@ -0,0 +1,232 @@
...
+/* Generates the request whose reply we defer; blocks until it is answered. */
+static void *client_thread_func(void *arg)
+{
+	char path[PATH_MAX];
+	struct stat stbuf;
+
+	(void)arg;
+
+	snprintf(path, sizeof(path), "%s/deferred", dt.mountpoint);
+	stat(path, &stbuf);  /* ENOENT is expected; we only need the round trip */
```
abhinavagarwal07, line 97: Could the client record and check this as `-1` / `ENOENT`? As written the test passes
whether the deferred `fuse_reply_err()` was delivered or the connection aborted the
request instead, since both unblock the `stat()`.

Worth keeping the join ahead of `fuse_session_unmount()`: that way only a delivered reply
can unblock the `stat()`, so a lost one fails on its own rather than depending on what
errno the abort path happens to produce. An `alarm()` in the child would turn that into a
fast failure — the parent's `WIFSIGNALED` branch already reports it.

lib/fuse_uring.c
```diff
@@ -683,6 +844,9 @@ static void fuse_uring_handle_cqe(struct fuse_ring_queue *queue,
 	req->interrupted = 0;
 	list_init_req(req);
 
+	/* dropped by fuse_free_req() when the application replies */
+	fuse_session_get(fuse_ring->se);
```
abhinavagarwal07, line 848: This puts two seq-cst RMWs on one `se->ref_cnt`, shared by all queues, on every io-uring
request (and on `/dev/fuse` too, via `fuse_ll_alloc_req`/`destroy_req`). There it's not a
new *kind* of cost — that path already does a malloc, `se->lock`, and a channel mutex per
request — but the uring path has `req->ch == NULL` and no `se->list` insertion, so as far
as I can tell this is the first per-request cross-core traffic there.

Since a held request retains its queue, the queue retains the pool, and the pool retains
the session, could a pool-scoped reference give the same guarantee for the uring path
without the per-request atomics? A `/dev/fuse` request held past destroy would still need
its own.

abhinavagarwal07 (commented):
Added a few comments

Issue #1515, opened 2026-06-05
removing libnuma dependency


benjaminp:
Post 44a1f759c2b9175fcc5e98d12162bb22cc40c980, fuse_uring uses only `numa_alloc_local` and `numa_free` from `libnuma`. `numa_alloc_local` is `mmap` plus `mbind` and `numa_free` is simply `munmap`. Would completely removing the `libnuma` dependency be reasonable?

bsbernd:
`mbind` is not libnuma?

```
mbind(2)                      System Calls Manual                      mbind(2)

NAME
       mbind - set memory policy for a memory range

LIBRARY
       NUMA (Non-Uniform Memory Access) policy library (libnuma, -lnuma)

SYNOPSIS
       #include <numaif.h>

```

```
bernd@srv1 ~>apt-file search numaif.h
libnuma-dev: /usr/include/numaif.h        
```

I guess you could do with syscalls, what is the real purpose here?

benjaminp:
Indeed, libnuma has the `mbind` C wrapper, but it is just doing `syscall(NR_mbind, ...)`, which libfuse could do.

I was just running into friction with libnuma as dependency. It has a library "initializer" and a version script, which makes it harder to use statically. I thought the easiest path might be eliminating it.

bsbernd:
I'm not opposed to remove libnuma dependency on libfuse, but for me it has no priority. Also not an issue either, as it all works as designed. Please feel free to submit a patch.

PR #1523, opened 2026-06-15
libfuse: remove libnuma dependency

benjaminp:
As discussed on https://github.com/libfuse/libfuse/issues/1515, libfuse has a very light dependency on libnuma, only using very light wrappers over mmap, mbind, and munmap. Remove the libnuma dependency by calling those syscalls directly.

### DIFF
--- a/lib/fuse_uring.c
+++ b/lib/fuse_uring.c
@@ -22,7 +22,9 @@
 #include <stdbool.h>
 #include <string.h>
 #include <unistd.h>
-#include <numa.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <linux/mempolicy.h>
 #include <pthread.h>
 #include <stdio.h>
 #include <linux/sched.h>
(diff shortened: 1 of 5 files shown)


lib/fuse_uring.c
```diff
@@ -732,6 +734,16 @@ static void fuse_uring_set_thread_core(int qid)
 	}
 }
 
+static void *alloc_local(size_t size)
+{
+	void *p = mmap(NULL, size, PROT_READ | PROT_WRITE,
+		       MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+	if (p == MAP_FAILED)
```
bsbernd, line 741: Could you add in here

madvise(p, siz, MADV_POPULATE_WRITE);

The project I was mainly working on till end of April is using mmap for all allocations and then got SIGBUS on customer systems, because mmap had not failed, even with MAP_POPULATE.

benjaminp: added

bsbernd (approved):

bsbernd:
Thank you @benjaminp !

PR #1524, opened 2026-06-16
fix fsconfig SElinux option parsing

lostjeffle:


lib/mount_fsmount.c
```diff
@@ -262,11 +262,37 @@ static int is_mtab_only_opt(const char *opt)
...
+	if (*p == ',') {
+		*p = '\0';
+		*pp = p + 1;
+	} else {
+		*pp = p;
+	}
+
+	return start;
+}
+
```
bsbernd, line 298: I find this hard to read, could you change to something like this?
```
/*
  * Example: parsing  rw,fd=7,context="u:r:t,s0"                                                                                                                                                            
   *                                                                                                                                                                                                         
   *   call 3 -> "context=\"u:r:t,s0\""   (comma inside quotes kept, not split)                                                                                                                              
   *   call 4 -> NULL                                                              
   * /                                                                                                                          
  static char *next_mount_opt(char **cursor)                                                                                                                                                                 
  {                                                                                                                                                                                                          
        char *opt_start;                                                                                                                                                                                     
        char *scan;                                                                                                                                                                                          
        int in_quote = 0;                                                                                                                                                                                    
                                                                                                                                                                                                             
        if (!*cursor || !**cursor)                                                                                                                                                                           
                return NULL;                                                                                                                                                                                 
                                                                                                                                                                                                             
        opt_start = *cursor;                                                                                                                                                                                 
        for (scan = opt_start; *scan; scan++) {                                                                                                                                                              
                if (*scan == '"')                                                                                                                                                                            
                        in_quote = !in_quote;                                                                                                                                                                
                else if (*scan == ',' && !in_quote)                                                                                                                                                          
                        break;                                                                                                                                                                               
        }                                                                                                                                                                                                    
                                                                                                                                                                                                             
        if (*scan == ',') {                                                                                                                                                                                  
                *scan = '\0';                                                                                                                                                                                
                *cursor = scan + 1;                                                                                                                                                                          
        } else {                                                                                                                                                                                             
                *cursor = scan;                                                                                                                                                                              
        }                                                                                                                                                                                                    

        return opt_start;
  }

```

bsbernd (approved):

PR #1528, opened 2026-06-18
fuse_lowlevel: check fuse_ll_ops2 in uring CQE dispatch

lostjeffle:
The uring CQE path dispatches handlers from fuse_ll_ops2[].


bsbernd:
Thank you @lostjeffle !

PR #1547, opened 2026-07-10
meson.build: Add Rocky Linux 9 compatibility for new mount API

kchen-ddn:
Rocky Linux 9 has kernel 5.14 which includes the new mount API syscalls (fsopen, fsconfig, fsmount, move_mount) introduced in Linux 5.2, but ships with glibc 2.34 which lacks the wrapper functions for these syscalls (wrappers were added in glibc 2.36).

Changes:
- meson.build: Detect Rocky Linux 9 at build time and force-enable HAVE_NEW_MOUNT_API instead of relying on compile check that would fail due to missing glibc wrappers

- mount_fsmount.c: Provide syscall wrappers for new mount API when glibc < 2.36 using direct syscall() invocations via __NR_* syscall numbers

This allows libfuse to use the new mount API on Rocky 9 systems by calling the kernel syscalls directly when glibc wrappers are not available, while automatically using glibc wrappers on systems with glibc >= 2.36.

### DIFF
--- a/lib/mount_fsmount.c
+++ b/lib/mount_fsmount.c
@@ -30,7 +30,93 @@
  * This file is only compiled conditionally when support for the new
  * mount API is detected - only flags that were not in the initial linux
  * commit introducing that API are defined here.
+ *
+ * Syscall number definitions and wrapper functions for new mount API.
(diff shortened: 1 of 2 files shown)


bsbernd:
I'm closing, because I made updates - in PR #1548

PR #1548, opened 2026-07-10
mount: detect new mount API libc-wrapper gap

bsbernd:
The kernel mount API syscalls (fsopen/fsconfig/fsmount/move_mount) landed in Linux 5.2, but libc wrapper functions can lag behind - e.g. glibc added them only in 2.36. RHEL9-family distros with an older glibc but a new-enough kernel (Rocky, RHEL, AlmaLinux, CentOS Stream, Oracle Linux 9 all ship glibc 2.34) fail the existing meson capability check even though the kernel itself supports the syscalls.

Detect this with two independent meson checks: one confirms the kernel headers declare the API, the other confirms libc provides the wrapper functions. When the headers are present but libc isn't, mount_fsmount.c supplies its own syscall()-based wrappers, and it defines any mount API constants a partial uapi header omits so the build never depends on the system header being complete. Neither check names a distro or a libc version, so the fallback engages for any libc/distro with the same gap, and it stays correct under cross-compilation since it is a real compile+link probe rather than a host-OS string match.

Changes by Bernd: reworked the original Rocky-9-specific detection to be portable - replaced the /etc/os-release distro-name match with the two independent header and libc probes above.

### DIFF
--- a/lib/mount_fsmount.c
+++ b/lib/mount_fsmount.c
@@ -26,6 +26,87 @@
 #include <sys/mount.h>
 #include <sys/syscall.h>
 
+#ifdef NEED_NEW_MOUNT_API_SYSCALL_WRAPPERS
+/*
+ * Kernel supports the new mount API (fsopen/fsconfig/fsmount/move_mount)
+ * but libc doesn't provide the wrapper functions yet (e.g. glibc added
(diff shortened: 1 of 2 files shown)


bsbernd:
The checkpatch failure is a false positive - I'm going to ignore it. @kchen-ddn Could you please review?

PR #1552, opened 2026-07-15
fuse_service: bound argc and arg len read from the args memfd

Keerthana-64:
fuse_service_append_args() takes the argument count and each argument length straight from the args memfd, which this file already treats as untrusted (see the SO_PASSRIGHTS guard against a malicious mount helper). Both fields are uint32_t and feed allocation math with no bound: `calloc(memfd_args.argc + existing_args->argc, ...)` wraps in unsigned arithmetic and undersizes the argv array, while `calloc(1, memfd_arg.len + 1)` wraps to a zero-size buffer when len is 0xffffffff, which the following pread() then overflows.

v2 after review. The first version bounded both fields against FUSE_SERVICE_MAX_CMD_SIZE, which was wrong: that constant caps a single socket command in receive_command() and says nothing about the argv memfd, which had no cap at all on either side.

This version adds FUSE_SERVICE_MAX_ARGV_SIZE and checks it on both ends. mount.service refuses to write a string that would push the file past the cap, and the server fstat()s the file and refuses to parse one larger than it. The file size then bounds the rest: argc cannot exceed the number of iovecs that fit between the header and the strings, and no string can be longer than the file holding it. argc 0 is rejected as well, since only the first loop iteration assigns argv[0].

1 MiB is the value I picked for the cap, and it is the one number that is a judgement call. It only needs to sit well clear of a real mount(8) argv while keeping argc and len far from the wrap, so it is easy to change.


Keerthana-64:
any update?

bsbernd:
@Keerthana-64 I don't think this commit is quite right. I have a branch that documents the machinery, not merged yet. Here is the flow graph that also includes your PR. I.e., without the PR one command at a time is tested, with the PR commit the sum all commands get checked.


```
 mount.service (separate proc)          ║  lib/fuse_service.c                      │  fuse server app
 util/mount_service.c                   ║  (linked into the server)                │  example/service_ll.c
 privileged, invoked by mount(8)        ║                                          │
════════════════════════════════════════╬══════════════════════════════════════════╧════════════════════════
                                        ║            one process                   │
 mount_service_connect() done   :354    ║   fuse server has been spawned by        │
   (see "Entry points" above)           ║   systemd with the connected fd          │
                                        ║                                          │
                                        ║                                          │  main()                 :240
                                        ║  fuse_service_accept()       :526 <──────┼──  fuse_service_accept()
                                        ║    count_listen_fds()        :331        │
                                        ║    find_socket_fd()          :392        │
                                        ║      fstat S_ISSOCK                      │
                                        ║      getsockname AF_UNIX                 │
                                        ║      check_sendbuf_size()    :368        │
                                        ║    calloc(sizeof(sf))        :565        │
                                        ║                                          │
 send_hello()                  :1807 ───╫─> negotiate_hello()          :448        │
   struct fuse_service_hello            ║     __recv_packet(&hello, 12) :459       │
   min/max_version, flags               ║     size != sizeof -> -EBADMSG           │
                                        ║     minver > MAX_PROTO                   │
                                        ║            -> -EOPNOTSUPP                │
                                        ║     flags & ~SERVICE_FLAGS               │
                                        ║            -> -EINVAL                    │
                                        ║     ALLOW_OTHER -> sf->allow_other       │
                                        ║     FUSEBLK     -> sf->can_fuseblk       │
        <───────────────────────────────╫── __send_packet(&reply, 8)   :514        │
                                        ║                                          │
 capture_args()                 :461    ║                                          │
   memfd_create("fuse svc argv") :484   ║                                          │  ┌──────────────────────────┐
   per arg: capture_arg()       :419    ║                                          │  │ NO SIZE CAP: argc and    │
     pwrite(string)             :430    ║                                          │  │ total memfd size are     │
     pwrite(memfd_arg)          :442    ║                                          │  │ unbounded on this side   │
   args.argc = htonl()          :516    ║                                          │  └──────────────────────────┘
   pwrite(&args, 8, 0)          :517    ║                                          │
                                        ║                                          │
 send_required_files()         :1819    ║                                          │
   SCM_RIGHTS: ARGV memfd       :625 ───╫─> receive_file("ARGV")       :580        │
   SCM_RIGHTS: /dev/fuse fd ────────────╫─> receive_file("FUSEDEV")    :590        │
                                        ║     __recv_fd()               :59        │
                                        ║     buf sized from OUR strlen(path)      │
                                        ║     size > bufsize  -> -EBADMSG          │
                                        ║     MSG_CTRUNC      -> -EBADMSG          │
                                        ║     bad cmsg len/level/type              │
                                        ║                     -> -EBADMSG          │
── fuse_service_accept() returns ───────╫─ traffic direction reverses ─────────────┼─ app drives below ───
                                        ║                                          │  fuse_service_accepted() :243
                                        ║  fuse_service_append_args()  :625 <──────┼──  append_args(&args)   :246
  ┌─ the memfd is read HERE;            ║    pread(&memfd_args, 8, 0)  :639        │
  │  every number below was             ║    magic != ARGS_MAGIC                   │
  │  chosen by mount.service            ║           -> -EBADMSG        :651        │
  └──────────────────────────           ║    argc = htonl(argc)        :656        │
                                        ║    argc > MAX_CMD_SIZE -> -EBADMSG       │  <== PR 1552, hunk 1
                                        ║    calloc(argc + existing->argc)         │
                                        ║                              :660        │
                                        ║    copy server's own argv    :676        │
                                        ║    for i < argc:             :700        │
                                        ║      pread(&memfd_arg)       :704        │
                                        ║      len = htonl(len)        :722        │
                                        ║      len >= MAX_CMD_SIZE                 │  <== PR 1552, hunk 2
                                        ║             -> -EBADMSG                  │
                                        ║      calloc(1, len + 1)      :726        │
                                        ║      pread(str, len, pos)    :736        │  (pos unvalidated)
                                        ║      i == 0 -> argv[0] = fstype          │
                                        ║    memcpy(existing_args)     :765 ───────┼─>  args now populated
                                        ║                                          │
                                        ║                                          │  fuse_opt_parse()       :249
                                        ║  parse_cmdline_opts()       :1198 <──────┼──  ...cmdline_opts()    :252
                                        ║  finish_file_requests()      :780 <──────┼──  ...file_requests()   :277
                                        ║    setsockopt(SO_PASSRIGHTS, 0)          │  no more fds accepted
                                        ║                                          │  fuse_session_new()     :283
                                        ║  fuse_service_session_mount()            │
                                        ║                             :1012 <──────┼──  session_mount()      :299
                                        ║                                          │
 main loop                     :1825    ║                                          │  one round trip per command
   receive_command()           :641 <───╫──── send_fsopen()            :812        │
     __recv_packet_size()       :162    ║      send_string(SOURCE)      :856       │
       MSG_PEEK|MSG_TRUNC               ║      send_string(MNTOPTS)     :856       │
     alleged_size < 0 / == 0            ║      send_string(MTABOPTS)    :856       │
     alleged_size < sizeof(hdr)         ║      send_mountpoint()        :909       │
     alleged_size >                     ║      send_mount()             :964       │
       FUSE_SERVICE_MAX_CMD_SIZE        ║                                          │  <== the ONLY pre-PR use
           -> reject            :666    ║                                          │
     calloc(1, alleged_size + 1):671    ║                                          │
     __recv_packet()            :679    ║                                          │
   switch (ntohl(p->magic))    :1834    ║                                          │
     OPEN / BDEV / FSOPEN /             ║                                          │
     SOURCE / MNTOPTS / MNTPT /         ║                                          │
     MTABOPTS / MOUNT                   ║                                          │
   send_reply() ────────────────────────╫──> __recv_packet(&reply, 8)              │
     fuse_service_simple_reply          ║      size != sizeof -> -EBADMSG          │
     (8 bytes, fixed)                   ║                                          │
                                        ║                                          │
                                        ║  fuse_service_send_goodbye() :304 <──────┼──  send_goodbye()       :304
 handle_bye_cmd()             :1866 <───╫── BYE_CMD                                │
   free(p); goto out                    ║                                          │  fuse_service_release()  :305
 exits                                  ║                                          │  fuse_session_loop()     :306
                                        ║                                          │  fuse_service_destroy()  :328
                                        ║                                          │  fuse_service_exit()     :331
```

bsbernd:
I think something like this might work
```
diff --git a/lib/fuse_service.c b/lib/fuse_service.c
index b42e6b53f516..ec77549e5522 100644
--- a/lib/fuse_service.c
+++ b/lib/fuse_service.c
@@ -629,8 +629,10 @@ int fuse_service_append_args(struct fuse_service *sf,
        struct fuse_args new_args = {
                .allocated = 1,
        };
+       struct stat statbuf;
        char *str = NULL;
        off_t memfd_pos = 0;
+       off_t max_argc;
        ssize_t received;
        unsigned int i;
        int ret;
@@ -656,6 +658,34 @@ int fuse_service_append_args(struct fuse_service *sf,
        memfd_args.argc = htonl(memfd_args.argc);
        memfd_pos += sizeof(memfd_args);
 
+       ret = fstat(sf->argvfd, &statbuf);
+       if (ret) {
+               int error = errno;
+
+               fuse_log(FUSE_LOG_ERR, "fuse: service args file stat: %s\n",
+                        strerror(error));
+               return -error;
+       }
+       if (statbuf.st_size > FUSE_SERVICE_MAX_ARGV_SIZE) {
+               fuse_log(FUSE_LOG_ERR, "fuse: service args file too large\n");
+               return -EBADMSG;
+       }
+
+       /*
+        * The array of argv iovecs sits between the header and the strings, so
+        * the file size bounds argc.  Reject a count the file cannot hold: the
+        * sum below is computed in unsigned arithmetic and would otherwise wrap
+        * and undersize the array.  argc 0 is rejected as well, because only
+        * the first loop iteration fills argv[0].
+        */
+       max_argc = (statbuf.st_size - (off_t)sizeof(memfd_args)) /
+                  (off_t)sizeof(struct fuse_service_memfd_arg);
+       if (memfd_args.argc == 0 || memfd_args.argc > max_argc) {
+               fuse_log(FUSE_LOG_ERR, "fuse: service args file argc %u invalid\n",
+                        memfd_args.argc);
+               return -EBADMSG;
+       }
+
        /* Allocate a new array of argv string pointers */
        new_args.argv = calloc(memfd_args.argc + existing_args->argc,
                               sizeof(char *));
@@ -722,6 +752,19 @@ int fuse_service_append_args(struct fuse_service *sf,
                memfd_arg.len = htonl(memfd_arg.len);
                memfd_pos += sizeof(memfd_arg);
 
+               /*
+                * A string cannot be longer than the file holding it.  len
+                * UINT32_MAX would make len + 1 wrap to zero below, handing
+                * calloc() a zero-size buffer for the pread() to overflow.
+                */
+               if (memfd_arg.len >= statbuf.st_size) {
+                       fuse_log(FUSE_LOG_ERR,
+                                "fuse: service args file argv[%u] len %u too large\n",
+                                i, memfd_arg.len);
+                       ret = -EBADMSG;
+                       goto out_new_args;
+               }
+
                /* read arg string from file */
                str = calloc(1, memfd_arg.len + 1);
                if (!str) {
diff --git a/util/mount_service.c b/util/mount_service.c
index 0b3266309a8a..5a61cf42ed98 100644
--- a/util/mount_service.c
+++ b/util/mount_service.c
@@ -427,6 +427,16 @@ static int mount_service_capture_arg(const struct mount_service *mo,
        };
        ssize_t written;
 
+       /*
+        * string_pos already covers the header and the whole array, so this
+        * bounds the entire memfd.  The server rejects anything larger.
+        */
+       if (*string_pos + (off_t)string_len > FUSE_SERVICE_MAX_ARGV_SIZE) {
+               fprintf(stderr, "%s: memfd argv[%u] exceeds %d byte limit\n",
+                       mo->msgtag, args->argc, FUSE_SERVICE_MAX_ARGV_SIZE);
+               return -1;
+       }
+
        written = pwrite(mo->argvfd, string, string_len, *string_pos);
        if (written < 0) {
                fprintf(stderr, "%s: memfd argv write: %s\n",
```

However, that needs to go via patch to `fuse-devel@lists.linux.dev first, as the author of the fuse service mount feature does not have a github account.

Keerthana-64:
You're right, and the flow graph makes it obvious why. FUSE_SERVICE_MAX_CMD_SIZE is the per-packet cap in receive_command() and, as your graph notes, its only pre-existing use, so reusing it for argc and len meant asserting a memfd limit that never existed. The helper side writes the memfd with no cap at all.

Pushed your version instead: FUSE_SERVICE_MAX_ARGV_SIZE checked in mount_service_capture_arg() on the write side, fstat() on the read side, argc bounded by the number of iovecs the file can hold, len bounded by the file size, plus the argc == 0 reject for the NULL argv[0] you pointed out. The cap value is the one thing your diff left open and I set it to 1 MiB. Happy with whatever you prefer there.

I left memfd_arg.pos alone even though the graph flags it as unvalidated. A bogus pos just short-reads and trips the existing `received < memfd_arg.len` check, so it ends up -EBADMSG rather than a safety issue. Can make it explicit if you'd rather not rely on that.

Can't build this here (Linux and HAVE_SERVICEMOUNT only), so I checked the new arithmetic on its own: argc and len of 0xffffffff are both rejected, and at the largest values that still pass, neither `argc + existing_args->argc` nor `len + 1` wraps. checkpatch is clean, and `argc > max_argc` doesn't trip -Wsign-compare since off_t is wider than uint32_t.

Understood on fuse-devel. I can post it there as a patch, or you can fold it into your machinery branch if that's easier to review as one series, since that branch is the thing that documents why these bounds exist.

PR #1554, opened 2026-07-18
fuse_lowlevel: fix out-of-bounds read in fuse_req_secctx_next

Keerthana-64:
fuse_req_secctx_next iterates the security-context extension libfuse copies out of a create/mknod/mkdir request into req->secctx. the header-fits check at the top of the loop uses a strict '>' against buf_end, so a fuse_secctx header can sit flush with the end of that copy; ctx_name then equals buf_end and buf_end - ctx_name - 1 underflows to SIZE_MAX, so the following strnlen reads off the end of the allocation. a 16-byte extension, a fuse_secctx_header plus one fuse_secctx with a nonzero size, trips it on the first call. add the missing name-fits check before that length is computed, matching the per-field -EIO checks already in the loop.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -3528,6 +3528,19 @@ int fuse_req_secctx_next(fuse_req_t req, const char **name,
 	}
 
 	const char *ctx_name = (const char *)(fctx + 1);
+
+	/*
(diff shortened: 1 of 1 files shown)


kchen-ddn (approved):
#LGTM

bsbernd:
Thanks, merged.

PR #1561, opened 2026-07-21
fuse_uring: defer REGISTER until the connection is initialized

joannekoong:
The io-uring ring threads send FUSE_IO_URING_CMD_REGISTER once they are woken from init_sem. They were woken in _do_init() right after send_reply_ok().

With async init this is fine since send_reply_ok() writes the FUSE_INIT reply and the kernel marks the connection initialized before write() returns, which means when the ring threads submit REGISTER, the connection is already initialized.

With sync init however, this races. _do_init() runs on the sync-init worker thread while the mount is still in progress, and the connection is not marked initialized until the mount completes (which happens after _do_init() returns). Waking the ring threads in _do_init() therefore lets REGISTER reach the kernel before initialization, and the kernel returns -EAGAIN.

Only wake the ring threads in _do_init() for async init, and for sync init defer the wake to fuse_session_mount_new_api() after the mount has completed and the connection is initialized. Sync init is only supported via the new mount API, so that is the only place the deferred wake is needed. se->is_sync_init is not set yet in _do_init(), so async is detected there via init_wakeup_fd == -1.

Signed-off-by: Joanne Koong <joannelkoong@gmail.com>

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -3065,7 +3065,19 @@ _do_init(fuse_req_t req, const fuse_ino_t nodeid, const void *op_in,
 	se->got_init = 1;
 	fuse_daemonize_set_got_init();
 	send_reply_ok(req, &outarg, outargsize);
-	if (enable_io_uring)
+	/*
+	 * With async init, send_reply_ok() has already marked the connection
+	 * initialized (that happens in the reply write()), so the ring threads
+	 * can send REGISTER now. With sync init, _do_init() runs on the
+	 * sync-init worker thread while the mount is still in progress. The
+	 * connection is not initialized until the mount completes, which is
+	 * after _do_init() returns. Waking the ring threads here would let
+	 * REGISTER race ahead of initialization and get -EAGAIN, so defer the
+	 * wake to fuse_session_mount_new_api() after the mount returns.
+	 * se->is_sync_init is not set yet at this point, so detect sync init
+	 * via init_wakeup_fd
+	 */
+	if (enable_io_uring && se->init_wakeup_fd == -1)
 		fuse_uring_wake_ring_threads(se);
 
 	/*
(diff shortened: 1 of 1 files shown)


bsbernd:
@joannekoong This is actually also racing without sync-init, which is mainly why there is an -EAGAIN handler. In principle we need to put the register requests onto a waitq in dev_uring.c in kernel, as long as the FUSE_INIT and mount are not completed. Problem is that we do not know in libfuse when FUSE_INIT is exactly handled - deferring the the wake reduces the race, but doesn't eliminate it.

bsbernd:
I also just merged another fix PR, I guess this PR here will time-out without rebase.

joannekoong:
>  This is actually also racing without sync-init, which is mainly why there is an -EAGAIN handler. 

Hi @bsbernd. Where do you observe the race without sync-init? AFAICT, the async init reply sets fch->initialized synchronously in the reply write() (the FR_ASYNC end callback runs in fuse_dev_do_write()), which means it's guaranteed the initialization state has been set by the time the write() returns, and libfuse only wakes the ring threads after that write (they block on init_sem before submitting REGISTER, which ensures no REGISTERS are submitted until after the wake).

bsbernd:
@joannekoong hmm yeahm you are right and we probably need to investigate the "why", but I was definitely seeing registrations that failed with EGAIN and got lost before commit 2bb27d8197cb9a3c6f0e7f17aa0cd96f1053ff47. I remember that I had investigated kernel and that FUSE_INIT was not completed yet.

lib/fuse_lowlevel.c
```diff
@@ -3065,7 +3065,19 @@ _do_init(fuse_req_t req, const fuse_ino_t nodeid, const void *op_in,
...
+	/*
+	 * With async init, send_reply_ok() has already marked the connection
+	 * initialized (that happens in the reply write()), so the ring threads
+	 * can send REGISTER now. With sync init, _do_init() runs on the
+	 * sync-init worker thread while the mount is still in progress. The
+	 * connection is not initialized until the mount completes, which is
+	 * after _do_init() returns. Waking the ring threads here would let
+	 * REGISTER race ahead of initialization and get -EAGAIN, so defer the
+	 * wake to fuse_session_mount_new_api() after the mount returns.
+	 * se->is_sync_init is not set yet at this point, so detect sync init
```
bsbernd, line 3077: Hi @joannekoong the fact that `se->is_sync_init` was not set was a critical bug and the reason why this PR here initially failed (all of libfuse was out of the sudden failing, just when we went on vacation *sigh* and why I got rather busy during my vacation in the nights (and in the end then didn't manage to review your buffer pool PR :/ ). The bad part about it was that pytest didn't want to hand out logs and I didn't even know the failing command. Main reason for a new rather large series to move away  from pytest to a self build runner: https://github.com/libfuse/libfuse/pull/1569

Anyway, could you please change from `se->init_wakeup_fd` to `se->is_sync_init`? Please see commit 507ff773c90c

bsbernd (changes requested):

bsbernd:
Closing in favor of new #1604

PR #1562, opened 2026-07-22
fuse_lowlevel: fix extension parse over-read on splice path

Keerthana-64:
Repro: on a splice_read mount with a security context, an unprivileged user makes a symlink with a ~4KB target, so the request arrives as FUSE_BUF_IS_FD.
Cause: fuse_session_process_buf_internal parses the extensions before the payload is copied in, so on that path it reads past the header-sized mbuf using in->len.
Fix: move the parse below the payload copy where 'in' spans the whole request, as the io_uring path already does.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -3984,16 +3984,7 @@ void fuse_session_process_buf_internal(struct fuse_session *se,
 	}
 
 	fuse_session_in2req(req, in);
-	if (in->total_extlen)
-		fuse_req_parse_extensions(req, in->total_extlen, in, in->len);
 	req->ch = ch ? fuse_chan_get(ch) : NULL;
-
-	if (se->debug && req->secctx_len > 0) {
-		fuse_log(FUSE_LOG_DEBUG,
-			"  secctx: %zu bytes (%u contexts)\n",
-			req->secctx_len,
-			req->secctx_count);
-	}
 
 	err = fuse_req_opcode_sanity_ok(se, in->opcode);
 	if (err)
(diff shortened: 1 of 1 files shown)


Keerthana-64:
any thoughts on this?

bsbernd:
Hi @Keerthana-64, thanks for the fix! Would you mind to slightly update it?
```
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -3931,8 +3931,10 @@
 		sizeof(struct fuse_write_in);
 	struct fuse_bufvec bufv = { .buf[0] = *buf, .count = 1 };
 	struct fuse_bufvec tmpbuf = FUSE_BUFVEC_INIT(write_header_size);
 	struct fuse_in_header *in;
+	/* 'in' spans the whole request, not just the fixed header */
+	bool in_splice_is_complete;
 	const void *inarg;
 	struct fuse_req *req;
 	void *mbuf = NULL;
 	int err;
@@ -3953,10 +3955,12 @@
 		if (res < 0)
 			goto clear_pipe;
 
 		in = mbuf;
+		in_splice_is_complete = (tmpbuf.buf[0].size == buf->size);
 	} else {
 		in = buf->mem;
+		in_splice_is_complete = true;
 	}
 
 	trace_request_process(in->opcode, in->unique);
 
@@ -4032,17 +4036,27 @@
 		if (res < 0)
 			goto reply_err;
 
 		in = mbuf;
+		in_splice_is_complete = true;
 	}
 
 	/*
-	 * Extensions are appended at the end of the request. On the splice path
-	 * 'in' only holds the fixed header until the payload is copied in above,
-	 * so parse them here, where 'in' spans the whole request for every path.
+	 * Extensions sit at the end of the request and are only reachable once
+	 * 'in' spans it. The splice path leaves FUSE_WRITE (with write_buf) and
+	 * FUSE_NOTIFY_REPLY payloads in the pipe; as of Linux 7.1 the kernel
+	 * attaches no extensions to those.
 	 */
-	if (in->total_extlen)
-		fuse_req_parse_extensions(req, in->total_extlen, in, in->len);
+	if (in->total_extlen) {
+		if (in_splice_is_complete)
+			fuse_req_parse_extensions(req, in->total_extlen, in,
+						  in->len);
+		else
+			fuse_log(FUSE_LOG_WARNING,
+				"fuse: %s: %zu extension bytes outside the header buffer, ignored\n",
+				opname((enum fuse_opcode) in->opcode),
+				FUSE_EXT_SIZE(in->total_extlen));
+	}
 
 	if (se->debug && req->secctx_len > 0) {
 		fuse_log(FUSE_LOG_DEBUG,
 			"  secctx: %zu bytes (%u contexts)\n",
```

bsbernd:
Well, maybe better `buf_is_complete`

PR #1566, opened 2026-07-28
mount_util: mark -o dev unsafe for unprivileged mounts

Keerthana-64:
Repro: as a normal user, `fusermount3 -o dev <mnt>` mounts with `rw,nosuid,relatime` and prints nothing, while `-o suid` prints "unsafe option suid ignored" and keeps `nodev`.
Cause: the `dev` row in the shared mount_flags table carries safe=1, and `safe` is what both `find_mount_flag()` in fusermount and `adjust_nonroot_mount_flags()` in mount.service use to decide what a non-root caller may clear. fusermount does the mount(2) as root, so nothing puts MS_NODEV back and the filesystem can serve device nodes the kernel honors. The row was safe=0 in the fusermount-local table until eec851b ("Move 'struct mount_flags' to util.h") merged it here.
Fix: restore safe=0 on that row, which covers both callers, and pin the safe column for the two flags in the unprivileged baseline with a unit test.


bsbernd:
Thank you very much @Keerthana-64 ! I got two security reports about this while being on vacation and wanted to fix that today.

PR #1571, opened 2026-07-30
mount_util: terminate the mount and umount argv with --

Keerthana-64:
Repro: as a normal user, `fusermount3 -o fsname=-oro,--target=/etc <mnt>` hands `-oro,--target=/etc` to `/bin/mount` as a positional operand. The same value arrives from an unprivileged fuse server as the `SOURCE` command payload on the mount.service socket.
Cause: `add_mount()` places `fsname` and `mnt` after `-o opts` with nothing terminating the option list, and `fuse_mount_spawn()` raises the real uid to 0 right before `posix_spawn()`, so mount(8) is not in restricted mode. getopt_long reads the operand as a second `-o` and the source operand is lost, which puts the caller in control of arguments to a root mount(8) and of the mtab/utab record this call exists to write.
Fix: add `--` after the options. `exec_umount()` and `remove_mount()` build their `/bin/umount` argv through the same spawn helper, so they get it too.

### DIFF
--- a/lib/mount_util.c
+++ b/lib/mount_util.c
@@ -279,9 +279,14 @@ static int add_mount(const char *progname, const char *fsname,
 		return -1;
 	}
 
+	/*
+	 * fsname comes from -ofsname= (or the service SOURCE command), so it
+	 * can start with '-'. Terminate the options with "--" to keep mount(8)
+	 * from parsing the operands as further options.
+	 */
 	char const * const argv[] = {
 		cmd, "--no-canonicalize", "-i", "-f", "-t", type, "-o", opts,
-		fsname, mnt, NULL
+		"--", fsname, mnt, NULL
 	};
 
 	res = fuse_mount_spawn(&pid, cmd, (char * const *) argv, &oldmask);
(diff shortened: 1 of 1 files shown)


bsbernd:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

One point surfaced during review but did not clear the confidence bar for a formal finding, worth a human look: `add_mount()`/`exec_umount()`/`remove_mount()` all exec `/bin/mount` or `/bin/umount`, which on some embedded/OpenWRT setups is a symlink to BusyBox (noted in an earlier fix to this file, https://github.com/libfuse/libfuse/pull/1247). BusyBox's mount/umount applets use their own option parser rather than glibc's `getopt_long`, so it's worth double-checking that `--` is honored as an option terminator there the same way it is by util-linux's `mount(8)` — otherwise the fix's protection may not be universal across all `mount(8)` implementations libfuse runs against.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

bsbernd:
Hi @Keerthana-64 Thanks for the fix. I had investigated busybox code for this - issue not fixed. I'm going to work on a follow up commit.

PR #1573, opened 2026-08-01
Fuse session exit doc and update fuse-loop logic to use 

### LINKED ISSUE #1572: fuse_session_exit() wakeup: #1201 missing from all 3.17.x releases; single-threaded loop still never woken; docs don't state the new guarantee
While fixing a shutdown-hang regression in xdg-desktop-portal's document portal (flatpak/xdg-desktop-portal#2083, flatpak/xdg-desktop-portal#999), I ended up characterizing the `fuse_session_exit()` wake behavior across releases and found three related gaps. Filing them together since they share one root: `fuse_session_exit()` historically only sets `se->exited`, and #997/#1201 fixed that for one of the two loops, on one branch.

**1. The mt-loop fix (#1201, `c5dbcdce`) is absent from every 3.17.x release.** It merged 2025-04-24 and first shipped in 3.18.0; 3.17.3 (2025-07-16) and 3.17.4 (2025-08-19) were both cut months after the merge without taking it. Measured with a teardown harness (session in a thread, external `fuse_session_exit()` + lazy unmount + join, a client holding an open fd so the unmount can't abort the connection): deterministic hang on 3.10.5, 3.13.1, 3.16.2 and 3.17.4; clean immediate exit on 3.18.2 and master. This matters downstream because distros freeze on 3.16/3.17 for years (Debian 13: 3.17.2; Fedora 43: 3.16.2; Ubuntu 24.04: 3.14.0) while consumers of the documented API keep hitting the pre-#1201 behavior. **Ask: would you take a cherry-pick of `c5dbcdce` onto `fuse-3.17.x`?** I realize the branch has been quiet since 3.17.4; if the answer is "no more 3.17.x releases", that's a useful signal for downstreams too.

**2. The single-threaded `fuse_session_loop()` still has no wake on any release, master included.** #1201 only covered `fuse_session_loop_mt()`'s control thread. A single-threaded loop blocked in `read(/dev/fuse)` never notices an external `fuse_session_exit()` until the next request. This pattern is live in shipping code: gtk-frdp (GNOME Connections) joins an st loop thread with no wake at all, and gnome-remote-desktop and FreeRDP wake it with a `stat()` of the mount root, which the kernel answers from the attribute cache whenever the entry is fresher than `attr_timeout` (both use 1.0 s), generating no FUSE request. Measured: harness with g-r-d's exact teardown order and attr timeout hangs 3/3 when the root was stat'ed ≤1 s before teardown; gtk-frdp's shape hangs 8/8 on 3.17.4 with a holder fd. The documented signal nudge, applied to the st loop (no-op handler + `pthread_kill()` of the loop thread after `fuse_session_exit()`), fixed it in 16/16 trials across 3.16.2 and 3.18.2; the loop's EINTR path rechecks the exited flag at both versions, so the mechanism is sound, it's just not something libfuse does for the caller. Worth fixing in-library (eventfd/pipe in the read path, or the #997 discussion's shapes), or at minimum worth an explicit statement that st-loop consumers must provide their own wake.

**3. The docs still describe only the old contract.** The `fuse_session_exit()` header comment (added 2023, `75672680`) tells non-worker-thread callers to send a signal afterwards. This is unchanged in 3.18.2 and master, with no mention that ≥ 3.18.0's mt loop no longer needs it. A sentence in the doc (and/or the 3.18 ChangeLog) stating what is now guaranteed (mt loop woken, st loop not) would let consumers gate their nudges on `FUSE_VERSION` deliberately (the portal PR compiles its nudge out at `>= FUSE_MAKE_VERSION(3, 18)`) and eventually drop them.

For what it's worth, this bug class is not unique to this codebase: macFUSE 5.3.3 (2026-07) independently fixed the same "exit flag set, blocked loop thread never woken" teardown hang, and the signal-wake pattern goes back to Miklos Szeredi's 2012 `fuse_terminate()` work (`pthread_kill(SIGTERM)` specifically because a bare exit flag doesn't wake a blocked `read()`). Happy to share the harnesses and full measurement logs for any of the above.

bsbernd:
fuse_session_exit() cannot wake the traditional single-threaded loop while it is blocked reading /dev/fuse. This change makes single-threaded loops built with FUSE_USE_VERSION 319 use one
fuse_session_loop_mt() worker, preserving serialized request handling while allowing fuse_session_exit() to wake the control thread.

fuse_session_loop()
      API before 3.19 --> caller-thread loop
                         fuse_session_exit() does not wake read(/dev/fuse)

      API 3.19+    --> one fuse_session_loop_mt() worker
                         fuse_session_exit() wakes the control thread
                         requests remain serialized

The branch also:

- Passes FUSE_USE_VERSION through versioned header entry points and validates it against the caller’s header version.
- Preserves source and ABI compatibility, including the pre-3.19 entry points and cuse_lowlevel_main().
- Keeps API 3.19 high-level loops wakeable when remember is enabled by running cache cleanup in the existing cleanup thread.
  - Fixes and enables test_setattr, using ftruncate() so the kernel supplies the expected file handle.
  - Documents which loops fuse_session_exit() can wake and lists API changes newest first.

  Fixes: https://github.com/libfuse/libfuse/issues/1572


pnowack:
So if I understand correctly, then `fuse_session_loop()` creates now its own thread here, where all fuse callbacks are handled?
If so, is then `fuse_session_exit()` now a blocking call (waiting for the session loop to be stopped), so that `fuse_session_unmount()` and `fuse_session_destroy()` can directly be called afterwards without having to do some kind of wait logic until the session loop is stopped?

bsbernd:
> So if I understand correctly, then `fuse_session_loop()` creates now its own thread here, where all fuse callbacks are handled?

@pnowack Changes only with FUSE_USE_VERSION 319 in order to avoid change of behavior. However, fuse_session_loop with FUSE_USE_VERSION < 319 is marked as deprecated. 

And yes, callback and signal handling is the reason why this is not the automatic default.

> If so, is then `fuse_session_exit()` now a blocking call (waiting for the session loop to be stopped), so that 

No, `fuse_session_exit()` just sets the session exit flag and updates the semaphore.

>`fuse_session_unmount()` and `fuse_session_destroy()` can directly be called afterwards without having to do some kind of wait logic until the session loop is stopped?

Yes, these should be always called after return of `fuse_session_loop`

pnowack:
@bsbernd The reason why I'm asking all this is, because in g-r-d, we currently use `FUSE_USE_VERSION 35` and we currently run `fuse_session_loop` in a dedicated thread there: https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/blob/038caa603604af85dd89bd92692c59bb57cfd10a/src/grd-rdp-fuse-clipboard.c#L1353
It looks like `FUSE_USE_VERSION 319` spawns now its own dedicated thread, so I won't need to create my own anymore.

So, if with `FUSE_USE_VERSION 319` `fuse_session_loop()` runs a dedicated thread on its own, but `fuse_session_exit()` would still just set a flag, wouldn't this mean that if I call `fuse_session_exit()`, then `fuse_session_unmount()` + `fuse_session_destroy()`, that the application can just crash, if the dedicated `fuse_session_loop()` didn't exit yet?
That is the question here. So basically I want with `FUSE_USE_VERSION 319` to ensure that the session loop is gone before destroying the session here.

My interpretation is that if `fuse_session_loop()` now with `FUSE_USE_VERSION 319` spawns its own dedicated thread, then it now longer blocks until `fuse_session_exit()`, but returns directly once that new dedicated thread is spawned. Or do I miss something here?

bsbernd:
@pnowack I'm sorry, I'm confused by your reply. Do you need something like `fuse_start_nonblocking_loop`? The PR and setting FUSE_USE_VERSION 319 only change that a fuse_session_exit() can wake threads that were blocked in kernel request read, but not that `fuse_session_loop` is blocking.

pnowack:
> Do you need something like `fuse_start_nonblocking_loop`?

No, I was concerned about the cleanup routine with `FUSE_USE_VERSION 319` now (compared to `35`) + whether the behaviour of `fuse_session_loop` is now changed. The confusion came from the comment in https://github.com/libfuse/libfuse/issues/1572#issuecomment-5151668227 and the three lines here: https://github.com/libfuse/libfuse/pull/1573/changes#diff-8f70e2866671fc5be68875d6621d18b0268099c86b398c87ff4277b5af9ede25R62, making me interpret that the behaviour of `fuse_session_loop` changes in the way, that it would create its own thread.

But your following response answers my question now:

> The PR and setting FUSE_USE_VERSION 319 only change that a fuse_session_exit() can wake threads that were blocked in kernel request read, but not that fuse_session_loop is blocking.

So it sounds like I can just change `FUSE_USE_VERSION 35` to `FUSE_USE_VERSION 319`, once FUSE 3.19 is released, and that's it. Sorry for the confusion.

PR #1579, opened 2026-08-07
lib: share the new mount API syscall fallbacks with util/

quobert:
mount_service_fsopen_mount() called fsconfig(), fsmount() and move_mount() directly, while every other new mount API call in that function goes through the helpers exported from mount_fsmount.c. On a libc without wrappers for those syscalls meson sets NEED_NEW_MOUNT_API_SYSCALL_WRAPPERS, but the fuse_-prefixed syscall() fallbacks were static inline inside mount_fsmount.c and thus invisible to util/.

Move the fallback block into mount_i_linux.h, which meson.build already documents as its home and which mount_service.c already includes, and use the fuse_-prefixed names at those three call sites. The header now pulls in fuse_config.h itself so it no longer depends on every includer having done so first.

### DIFF
--- a/lib/mount_fsmount.c
+++ b/lib/mount_fsmount.c
@@ -24,59 +24,6 @@
 #include <fcntl.h>
 #include <errno.h>
 #include <sys/mount.h>
-#include <sys/syscall.h>
-
-#ifdef NEED_NEW_MOUNT_API_SYSCALL_WRAPPERS
-/*
- * Kernel supports the new mount API (fsopen/fsconfig/fsmount/move_mount)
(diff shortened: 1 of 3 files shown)


bsbernd:
@quobert I'm going to hold back on merging this immediately, I would actually prefer if mount_service.c would call into mount_fsmount.c to acoid code duplication. Code dup came up because mount_service.c was developed rather exactly at the same time as mount_fsmount.c

quobert:
Yes, makes sense. Closing this PR as you address this in #1584 already.

PR #1580, opened 2026-08-07
lib: relay the KILLPRIV_V2 kill-suidgid flags to the filesystem

quobert:
With FUSE_CAP_HANDLE_KILLPRIV_V2 the kernel stops clearing suid/sgid and security.capability itself and tells the server per request when to do it. Of the three flags carrying that request only FATTR_KILL_SUIDGID made it through, because FUSE_SET_ATTR_KILL_SUID happens to be the same bit. FUSE_OPEN_KILL_SUIDGID and FUSE_WRITE_KILL_SUIDGID were dropped, so a server that negotiated KILLPRIV_V2 never saw the request on O_TRUNC open and on write, and the bits survived.

Decode both into a new fuse_file_info::kill_suidgid, taken from the padding bits, in open, create, write and write_buf. In create the flag sits in fuse_create_in::open_flags, which only exists since 7.12, hence the existing proto_minor guard.

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -96,8 +96,13 @@ struct fuse_file_info {
 	    file */
 	uint32_t parallel_direct_writes : 1;
 
+	/** Set in open, create and write if FUSE_CAP_HANDLE_KILLPRIV_V2 was
+	    enabled and the filesystem has to unset the setuid and setgid bits
+	    and drop the capability xattr. */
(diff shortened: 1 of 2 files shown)


bsbernd:
@quobert Thank you and sorry for delay, in order to proceed I opened an updated PR: https://github.com/libfuse/libfuse/pull/1599

PR #1582, opened 2026-08-10
lib: commit the ring entry in fuse_reply_none()

quobert:
Over io_uring a ring entry only goes back to the kernel via COMMIT_AND_FETCH. fuse_reply_none() merely freed the request, so a FORGET answered this way leaked the entry and left the kernel-side request outstanding.


bsbernd (approved):
Uh oh, interesting that I never noticed. Good catch!

PR #1588, opened 2026-08-17
fuse_lowlevel: add buffer size sanity check for init requests

### LINKED ISSUE #1587: oob read in `_do_init` (`fuse_session_process_buf`) when a `FUSE_INIT` request payload is smaller than `struct fuse_init_in`
## Summary

When `fuse_session_process_buf()` processes a memory-backed `fuse_buf` whose contents form a `FUSE_INIT` request with a payload smaller than `struct fuse_init_in` (64 bytes), the FUSE_INIT handler `_do_init()` (`lib/fuse_lowlevel.c:2744`) reads `arg->minor` past the end of the buffer.

`fuse_session_process_buf_internal()` (`fuse_lowlevel.c:4034`) computes `inarg = (void *) &in[1]` (a bare pointer to the payload, with no length) and dispatches via `fuse_ll_ops[in->opcode].func(req, in->nodeid, inarg)` (line 4040) directly into `do_init()` → `_do_init()`. The handler signature carries no payload length, and the dispatcher never validates that `buf->size - sizeof(struct fuse_in_header)` is large enough for the opcode's expected input struct. `_do_init()` therefore casts the undersized payload to `const struct fuse_init_in *arg` (64 bytes) and dereferences fields beyond the allocation.

## Version

```bash
$ git describe --tags
fuse-3.16.2-666-g267436b8
```

Pinned commit: `267436b811cff72aa5a1c939597a0583735d359d`

## Description

`fuse_session_process_buf_internal()` reads the `fuse_in_header` from the buffer, runs opcode-type sanity checks, then dispatches to the per-opcode handler. On the memory-backed path (no `FUSE_BUF_IS_FD` flag) it sets `in = buf->mem` directly:

```c
// fuse_lowlevel.c:3928-3946 (abridged)
if (buf->flags & FUSE_BUF_IS_FD) {
    ...                          // FD/pipe path: not taken here
} else {
    in = buf->mem;              // <-- in points at the caller's buffer
}
```

After the header is parsed and opcode sanity passes, it dispatches:

```c
// fuse_lowlevel.c:4034-4040 (abridged)
inarg = (void *) &in[1];                                   // bare payload pointer, no length
if (in->opcode == FUSE_WRITE && se->op.write_buf)
    do_write_buf(req, in->nodeid, inarg, buf);
else if (in->opcode == FUSE_NOTIFY_REPLY)
    do_notify_reply(req, in->nodeid, inarg, buf);
else
    fuse_ll_ops[in->opcode].func(req, in->nodeid, inarg);  // <-- do_init(req, nodeid, inarg)
```

The two checks that gate dispatch are type-only and never look at size:

```c
// fuse_lowlevel.c:3722-3737 — opcode TYPE only
static inline int fuse_req_opcode_sanity_ok(struct fuse_session *se, enum fuse_opcode in_op)
{
    if (!se->got_init) {
        enum fuse_opcode expected = se->cuse_data ? CUSE_INIT : FUSE_INIT;
        if (in_op != expected) return err;   // our opcode IS FUSE_INIT -> passes
    } ...
    return 0;
}
// fuse_lowlevel.c:3994 — bounds the opcode index, not the payload
if (in->opcode >= FUSE_MAXOP || !fuse_ll_ops[in->opcode].func) goto reply_err;
```

`do_init()` is a thin wrapper that forwards the bare pointer:

```c
// fuse_lowlevel.c:3080-3084
static void do_init(fuse_req_t req, fuse_ino_t nodeid, const void *inarg)
{
    _do_init(req, nodeid, inarg, NULL);   // no size carried
}
```

`_do_init()` then casts the payload to the 64-byte struct and dereferences it immediately:

```c
// fuse_lowlevel.c:2720-2744 (abridged)
_do_init(fuse_req_t req, const fuse_ino_t nodeid, const void *op_in, const void *in_payload)
{
    const struct fuse_init_in *arg = op_in;   // 5-byte payload reinterpreted as 64-byte struct
    ...
    se->conn.proto_major = arg->major;        // offset 0, 4 bytes -> payload bytes 0..3  (in bounds)
    se->conn.proto_minor = arg->minor;        // offset 4, 4 bytes -> payload bytes 4..7  (OOB: only byte 4 exists)
    ...
}
```

`struct fuse_init_in` is 64 bytes (`include/fuse_kernel.h:902-909`):

```c
struct fuse_init_in {
    uint32_t major; uint32_t minor; uint32_t max_readahead;
    uint32_t flags; uint32_t flags2; uint32_t unused[11];
};
```

### How the trigger parameters are reached

The trigger uses only the public API declared in `include/fuse3/fuse_lowlevel.h`:

- `fuse_session_new(&args, &ops, sizeof(ops), NULL)` — constructs a valid low-level session with stub `init`/`destroy` callbacks. `se->got_init == 0`, so the sanity check expects the first request to be `FUSE_INIT`.
- `fuse_session_process_buf(se, &buf)` — documented (`fuse_lowlevel.h:2459-2468`) as *"Process a raw request supplied in a generic buffer. The fuse_buf may contain a memory buffer or a pipe file descriptor."* No precondition on buffer or payload size.
- `fuse_buf` is memory-backed (`flags == 0`, so `FUSE_BUF_IS_FD` is unset; the library uses `buf->mem` per `fuse_common.h:927`). Only `size` and `mem` are set, per the documented contract.

No memory ownership or call-order rule is violated: the buffer is heap-allocated by the caller and remains valid across the call; `fb.size` matches the actual allocation; `in->len` matches `fb.size`. The overflow is manufactured entirely inside the library's dispatch + handler, by the missing payload-size check.

## PoC Code

```cpp
#define FUSE_USE_VERSION 312
#include <cstdint>
#include <cstdlib>
#include <cstring>
#include <fuse3/fuse_lowlevel.h>
#include <fuse3/fuse_common.h>
#include <fuse3/fuse_opt.h>

// fuse_in_header lives in the internal fuse_kernel.h; redefine locally (sizeof==40).
struct fuse_in_header {
    uint32_t len, opcode; uint64_t unique, nodeid;
    uint32_t uid, gid, pid; uint16_t total_extlen, padding;
};
static_assert(sizeof(fuse_in_header) == 40, "layout");
static void stub_init(void *u, struct fuse_conn_info *c) { (void)u; (void)c; }
static void stub_destroy(void *u) { (void)u; }

int main(void) {
    char *argv[] = {(char *)"poc", nullptr};
    struct fuse_args args = FUSE_ARGS_INIT(1, argv);
    struct fuse_lowlevel_ops ops;
    memset(&ops, 0, sizeof(ops));
    ops.init = stub_init;
    ops.destroy = stub_destroy;

    struct fuse_session *se = fuse_session_new(&args, &ops, sizeof(ops), nullptr);
    if (!se) { fuse_opt_free_args(&args); return 1; }

    // FUSE_INIT (26) + 5-byte payload: arg->major (off 0, 4 B) is in bounds,
    // arg->minor (off 4, 4 B) straddles the 5-byte payload end -> OOB read.
    // Heap-allocated so ASan reports heap-buffer-overflow, matching crash_report.md.
    constexpr size_t kTotal = sizeof(fuse_in_header) + 5;
    uint8_t *buf = (uint8_t *)calloc(1, kTotal);
    auto *h = reinterpret_cast<struct fuse_in_header *>(buf);
    h->len = kTotal;
    h->opcode = 26; // FUSE_INIT
    h->uid = 0;     // root: avoids allow_root rejection

    struct fuse_buf fb;
    memset(&fb, 0, sizeof(fb));
    fb.size = kTotal;
    fb.mem = buf;

    fuse_session_process_buf(se, &fb);
    fuse_session_destroy(se);
    fuse_opt_free_args(&args);
    free(buf);
    return 0;
}
```

## Stack Trace

```bash
ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6d4b157e007c at pc 0x5ef7121d07f9 bp 0x7ffc826d84f0 sp 0x7ffc826d84e8
READ of size 4 at 0x6d4b157e007c thread T0
    #0 0x5ef7121d07f8 in _do_init lib/fuse_lowlevel.c:2744:30
    #1 0x5ef7121c61e8 in do_init lib/fuse_lowlevel.c:3083:2
    #2 0x5ef7121b23b1 in fuse_session_process_buf_internal lib/fuse_lowlevel.c:4040:3
    #3 0x5ef7121af500 in fuse_session_process_buf lib/fuse_lowlevel.c:3910:2
    #4 0x5ef712197a59 in main poc.cpp:44:5
    #5 0x710b1640c1c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #6 0x710b1640c28a in __libc_start_main csu/../csu/libc-start.c:360:3
    #7 0x5ef7120ac764 in _start (poc+0xa3764) (BuildId: 4f8fb576ed7734768c9a954b2a4a721e239c4cde)

0x6d4b157e007d is located 0 bytes after 45-byte region [0x6d4b157e0050,0x6d4b157e007d)
allocated by thread T0 here:
    #0 0x5ef7121520ad in calloc (poc+0x1490ad) (BuildId: 4f8fb576ed7734768c9a954b2a4a721e239c4cde)
    #1 0x5ef7121977c9 in main poc.cpp:33:31
    #2 0x710b1640c1c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x710b1640c28a in __libc_start_main csu/../csu/libc-start.c:360:3
    #4 0x5ef7120ac764 in _start (poc+0xa3764) (BuildId: 4f8fb576ed7734768c9a954b2a4a721e239c4cde)

SUMMARY: AddressSanitizer: heap-buffer-overflow lib/fuse_lowlevel.c:2744:30 in _do_init
```

## Reproduction Step

```bash
# 1. Checkout libfuse at the version above
cd libfuse
git checkout 267436b811cff72aa5a1c939597a0583735d359d

# 2. Build the library with ASan+UBSan (static libs + headers)
export CC=clang CXX=clang++ LD=clang++
export CFLAGS="-fsanitize=address,undefined -fno-sanitize=function \
  -fsanitize-address-use-after-scope -g -O0 -fstandalone-debug -w -fPIC -fno-inline \
  -fno-omit-frame-pointer -fno-optimize-sibling-calls \
  -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types \
  -Wno-error=int-conversion -Wno-error=deprecated-declarations \
  -Wno-error=implicit-function-declaration -Wno-error=implicit-int \
  -Wno-error=vla-cxx-extension -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
export CXXFLAGS="$CFLAGS" LDFLAGS="$CXXFLAGS"
meson setup . build-san --prefix="$PWD/build-san/install" \
  --default-library=static --buildtype=plain \
  -Dutils=false -Dexamples=false -Dtests=false \
  -Ddisable-libc-symbol-version=true -Denable-io-uring=false \
  -Denable-usdt=false -Duseroot=false
ninja -C build-san -j && ninja -C build-san install

# 3. Build the standalone PoC against the sanitizer build
clang++ -g -O0 -fno-omit-frame-pointer -ftrivial-auto-var-init=zero \
    -fsanitize=address,undefined \
    -Ibuild-san/install/include \
    poc.cpp -o poc \
    -Wl,--start-group build-san/install/lib/libfuse3.a -Wl,--end-group \
    -lpthread -ldl

# 4. Run
ASAN_OPTIONS=detect_leaks=0 ./poc
```

The crash is deterministic and reproduces on every run.

## Submission Statement

This report was produced by FuzzAnything's AI-assisted library fuzzer and manually verified by a team member. We reviewed the PoC against the upstream API documentation — call order, parameters, and memory ownership — and found no API misuse.

> Signed-off-by: FuzzAnything <fuzzanything@gmail.com>

yvonnelxxxx:
This PR attempts to fix [#1587](https://github.com/libfuse/libfuse/issues/1587#issue-5169821882)

## Problem

`fuse_session_process_buf_internal()` (`lib/fuse_lowlevel.c`) dispatches into per-opcode handlers with a bare payload pointer and no payload-size check:

```c
inarg = (void *) &in[1];                                   // fuse_lowlevel.c:4034
...
fuse_ll_ops[in->opcode].func(req, in->nodeid, inarg);      // fuse_lowlevel.c:4040
```

The two guards that gate dispatch — `fuse_req_opcode_sanity_ok()` (opcode *type* only, lines 3722–3737) and the `FUSE_MAXOP` / non-NULL-func check (line 3994) — never look at size. The handler signature `void (*func)(req, node, arg)` carries no length. So when a memory-backed `fuse_buf` contains a `FUSE_INIT` request whose payload is shorter than `struct fuse_init_in` (64 bytes), `_do_init()` casts the short payload to the 64-byte struct and dereferences past the end:

```c
const struct fuse_init_in *arg = op_in;          // fuse_lowlevel.c:2724
...
se->conn.proto_major = arg->major;               // :2743  off 0, 4 B — in bounds
se->conn.proto_minor = arg->minor;               // :2744  off 4, 4 B — OOB read
```

## Fix

A single guarded block in `fuse_session_process_buf_internal()`, inserted right after the existing `FUSE_MAXOP` / non-NULL-func guard (line 3995) and before any handler is entered. A `FUSE_INIT` / `CUSE_INIT` request whose payload is shorter than the fields `_do_init` reads unconditionally is replied to with `EPROTO` via the existing `reply_err` path — the same errno `_do_init` already uses for an unsupported protocol version (`fuse_lowlevel.c:2755`) — and the handler is never called:

```c
/* fuse_lowlevel.c — in fuse_session_process_buf_internal(),
   right after the FUSE_MAXOP / func-null guard (line 3995) */
if (in->opcode == FUSE_INIT || in->opcode == CUSE_INIT) {
	size_t hdr  = sizeof(struct fuse_in_header);
	size_t have = buf->size < in->len ? buf->size : in->len;
	size_t need = offsetof(struct fuse_init_in, max_readahead); /* 8: major+minor */
	if (have < hdr || have - hdr < need) {
		err = EPROTO;
		goto reply_err;
	}
}
```

This closes the reported `FUSE_INIT` path and the symmetric `CUSE_INIT` path (same `fuse_init_in` cast). Other opcodes have the same latent shape but are not part of this report; a per-opcode minimum-size table covering them is noted as a separate follow-up.

### Sanitizer rebuild

```bash
export CC=clang CXX=clang++ LD=clang++
export CFLAGS="-fsanitize=address,undefined -fno-sanitize=function \
  -fsanitize-address-use-after-scope -g -O0 -fPIC -fno-inline \
  -fno-omit-frame-pointer -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
export CXXFLAGS="$CFLAGS" LDFLAGS="$CXXFLAGS"
meson setup src/libfuse build-san --prefix="$PWD/build-san/install" \
    --default-library=static --buildtype=plain \
    -Dutils=false -Dexamples=false -Dtests=false \
    -Ddisable-libc-symbol-version=true -Denable-io-uring=false \
    -Denable-usdt=false -Duseroot=false
ninja -C build-san -j && ninja -C build-san install
```

### Before fix

```bash
$ clang++ -g -O0 -fsanitize=address,undefined \
    -Ibuild-san/install/include poc.cpp -o poc \
    build-san/install/lib/libfuse3.a -lpthread -ldl
$ ASAN_OPTIONS=detect_leaks=0 ./poc
SUMMARY: AddressSanitizer: heap-buffer-overflow fuse_lowlevel.c:2744:30 in _do_init
```
Exit status: non-zero (ASan abort). Crash is deterministic, reproduces on every run.

### After fix

```bash
$ ASAN_OPTIONS=detect_leaks=0 ./poc
fuse: writing device: Bad file descriptor
$ echo $?
0
```

No ASan report; the process exits 0. The `Bad file descriptor` line is the *reply* write failing on a missing `/dev/fuse` fd:
- The check fires → `goto reply_err` → `fuse_reply_err(req, EPROTO)` → `fuse_send_msg` → `writev(se->fd, …)` at `fuse_lowlevel.c:289`.
- The standalone PoC never opens `/dev/fuse`, so `se->fd == -1` and the `writev` fails with `EBADF`; `perror("fuse: writing device")` at `fuse_lowlevel.c:295` prints the line.

The proof the fix works is **what is absent**: before the fix `_do_init` runs and ASan aborts at line 2744 *before* any reply is sent (so this line never appears); after the fix `_do_init` is skipped, the EPROTO reply path finally runs, and the only thing that fails is the harmless reply write to a non-existent device. `_do_init` is never entered; the OOB read cannot occur. (The EBADF would appear for *any* reply this standalone harness triggers — it is a property of having no mounted FUSE channel, not of the fix.)

### Test-suite results

The existing built-in test suite was rebuilt and run under the fix to confirm no regression. The check only rejects messages that previously caused undefined behaviour, so no well-formed request changes behaviour:

```bash
$ meson setup --reconfigure build -Dtests=true
$ ninja -C build
$ meson test -C build -v
```
The suite passes.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -4079,6 +4079,15 @@ void fuse_session_process_buf_internal(struct fuse_session *se,
 	err = ENOSYS;
 	if (in->opcode >= FUSE_MAXOP || !fuse_ll_ops[in->opcode].func)
 		goto reply_err;
+	if (in->opcode == FUSE_INIT || in->opcode == CUSE_INIT) {
+		size_t hdr  = sizeof(struct fuse_in_header);
+		size_t have = buf->size < in->len ? buf->size : in->len;
+		size_t need = offsetof(struct fuse_init_in, max_readahead); /* 8: major+minor */
+		if (have < hdr || have - hdr < need) {
+			err = EPROTO;
+			goto reply_err;
+		}
+	}
 	/* Do not process interrupt request */
 	if (se->conn.no_interrupt && in->opcode == FUSE_INTERRUPT) {
 		if (se->debug)


bsbernd:
@yvonnelxxxx I'm not against an additional sanity check, as this might help to discover generic libfuse buffer handling issues, there is just no bug/issue here. Could please force push and add a commit message stating that this is an additonal sanity check?

yvonnelxxxx:
Thanks for the review @bsbernd. I've updated the commit message to state that this is an additional sanity check and brought the branch up to date with master.

bsbernd:
@yvonnelxxxx The commit message is still empty and your Signed-off-by: line is missing as well. Updating the PR message is not going to help - it has to be the commit message, i.e. you need to use "git push -f"

PR #1589, opened 2026-08-19
lib: read the retrieve reply args from the io_uring payload buffer

lostjeffle:
FUSE_NOTIFY_RETRIEVE is the only notification the kernel answers, and FUSE_NOTIFY_REPLY is that answer: it arrives as a request carrying a fuse_notify_retrieve_in followed by the retrieved data.  It is therefore the only notification whose reply the daemon has to decode at all.

That request has no op header of its own, the kernel leaving in_args[0] zero sized, so fuse_uring_args_to_ring() skips the copy into the op_in area altogether and the fuse_notify_retrieve_in is placed in the payload buffer, with the data right behind it.

Fixes: d4abf661f994 ("Add fuse_session_process_uring_cqe")

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -3208,7 +3208,7 @@ static void list_init_nreq(struct fuse_notify_req *nreq)
 	nreq->prev = nreq;
 }
 
-static void do_notify_reply(fuse_req_t req, fuse_ino_t nodeid,
+static void _do_notify_reply(fuse_req_t req, fuse_ino_t nodeid,
 			    const void *inarg, const struct fuse_buf *buf)
 {
 	struct fuse_session *se = req->se;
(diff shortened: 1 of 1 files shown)


bsbernd:
Hi @lostjeffle I think this is another argument for https://github.com/libfuse/libfuse/pull/1212. I'm going to rebase/update it and add your commit with slight changes on top.
I was bit hesitating for #1212 because it needs a function argument and register just for one single user, but I don't like all these if condition exceptions and with your change here it probably makes even more sense.

bsbernd:
Well, I can also merge your change first and then rebase #1212 on top of it

lostjeffle:
> Hi @lostjeffle I think this is another argument for #1212. I'm going to rebase/update it and add your commit with slight changes on top. I was bit hesitating for #1212 because it needs a function argument and register just for one single user, but I don't like all these if condition exceptions and with your change here it probably makes even more sense.

Yes only the io_uring path can be refactored like what https://github.com/libfuse/libfuse/pull/1212 does, while these if conditions keep there for the classical /dev/fuse path.

bsbernd:
Thank you @lostjeffle ! And sorry for late merge.

PR #1592, opened 2026-08-25
fuse_service: free command buffers on send failure

junkilee80:
## Problem

`send_string()` and `send_mountpoint()` allocate their command buffers before
calling `__send_packet()`. If `sendmsg()` fails, both functions return before
reaching the existing success-path `free(cmd)`, leaking one command buffer per
failed send.

Closing the peer of an `AF_UNIX` `SOCK_SEQPACKET` socketpair before calling the
two helpers forces `EPIPE`. LeakSanitizer then reports two direct leaks from the
command allocations (21 and 29 bytes with the reproduction strings used).

## Fix

Route every post-allocation return in both functions through a shared cleanup
path. Each error path stores the return value in `err`, while a valid reply sets
it to zero before the common `free(cmd)`. This fixes the send-error leak, keeps
the returned errors and successful flow unchanged, and makes future error paths
less likely to miss the caller-owned buffer.

## Verification

- Before the change, the forced `EPIPE` reproduction reports two direct leaks.
- After the change, both send-error paths are exercised without a
  LeakSanitizer report.
- The patched tree builds with Clang 18 and ASan/UBSan on Ubuntu 24.04 ARM64.
- `meson test` passes `fuse-tests`; the io-uring variant is skipped because the
  build disables io-uring.
- The patch passes the repository's `checkpatch.pl` with 0 errors and 0
  warnings.

### DIFF
--- a/lib/fuse_service.c
+++ b/lib/fuse_service.c
@@ -861,49 +861,51 @@ static int send_string(const struct fuse_service *sf, uint32_t command,
 	const size_t valuelen = strlen(value);
 	const size_t cmdsz = sizeof_fuse_service_string_command(valuelen);
 	ssize_t size;
+	int err;
 
 	cmd = calloc(1, cmdsz);
 	if (!cmd) {
-		int error = errno;
-
+		err = -errno;
 		fuse_log(FUSE_LOG_ERR, "fuse: alloc service string send: %s\n",
-			 strerror(error));
-		return -error;
+			 strerror(-err));
+		goto out;
 	}
 	cmd->p.magic = htonl(command);
 	memcpy(cmd->value, value, valuelen + 1);
 
 	size = __send_packet(sf, cmd, cmdsz);
 	if (size < 0) {
-		int error = errno;
-
+		err = -errno;
 		fuse_log(FUSE_LOG_ERR, "fuse: send service string: %s\n",
-			 strerror(error));
-		return -error;
+			 strerror(-err));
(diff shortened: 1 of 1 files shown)


bsbernd:
Hi @junkilee80 , I don't manage to review today anymore and will be traveling the next two days - going to review on Friday.

lib/fuse_service.c
bsbernd: How about refactoring the whole function a bit to make it less prone to those leaks?

```
    int err;

...

     *errorp = ntohl(reply.error);
     err = 0;
out:
    free(cmd);
    return err;
```

lib/fuse_service.c
```diff
@@ -931,6 +932,7 @@ static int send_mountpoint(const struct fuse_service *sf, mode_t expected_fmt,
 	if (size < 0) {
 		int error = errno;
 
+		free(cmd);
```
bsbernd: Same for this function?

lib/fuse_service.c
junkilee80: Thanks, agreed. 
I refactored send_string() to route every post-allocation exit through a single out: cleanup path. 
err preserves the error result and is set to zero only after a valid reply.

lib/fuse_service.c
junkilee80: Done. Thank you.

bsbernd (approved):
Thank you @junkilee80 !

PR #1593, opened 2026-08-26
fuse io-uring: add support for bufpools

joannekoong:
Bufpools were added to the kernel in 7.3. This PR adds the userspace-side code to use bufpools in libfuse. Changes for zero copy will be in a separate PR.


bsbernd:
@joannekoong pipeline is now also running with io-uring, could you please add a run with buf pools?

lib/fuse_uring.c
```diff
@@ -817,6 +1027,14 @@ static int fuse_uring_init_queue(struct fuse_ring_queue *queue)
 	queue->req_header_sz = ROUND_UP(sizeof(struct fuse_uring_req_header),
 				       page_sz);
 
+	queue->bufpool_sz = ring->queue_depth * ring->max_req_payload_sz;
```
abhinavagarwal07, line 1030: Should this use the same chunk size the kernel will?

libfuse allocates `queue_depth * max_write`. The kernel splits that by [`max(FUSE_MIN_READ_BUFFER, max_write, max_pages * PAGE_SIZE)`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fuse/dev_uring.c?h=v7.3-rc1#n317) and [`-EINVAL`s if no whole chunk fits](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fuse/dev_uring.c?h=v7.3-rc1#n1600).

They agree at the default, but not for a filesystem that lowers `conn->max_write` in `init()` — that runs before the clamp at `fuse_lowlevel.c:2996-2997`. With `max_write=4096` at depth 8 the kernel rounds to 8192 and the queue gets 4 chunks for 8 entries; a non-page-multiple value loses one the same way. The ring silently runs at a fraction of its depth.

Should the stride here, and the range check at :877, be the kernel's rounded value rather than `max_req_payload_sz`?

lib/fuse_uring.c
```diff
@@ -888,6 +1132,18 @@ static void *fuse_uring_thread(void *arg)
 
 	sem_wait(&ring_pool->init_sem);
 
+	/*
+	 * The connection is initialized now, so the kernel accepts the queue
+	 * setup commands. This is also what prepares the registration SQEs.
+	 */
+	err = fuse_uring_start_queue(queue);
```
abhinavagarwal07, line 1139: What happens to the mount if the kernel rejects one of these?

Before this patch queue setup was preparation only — `fuse_uring_register_queue()` ended with `/* Only preparation until here, no submission yet */`, so nothing reached the kernel until after the INIT reply, and a failure in `fuse_uring_start()` just cleared `FUSE_OVER_IO_URING` and served over `/dev/fuse` (`fuse_lowlevel.c:3113-3118`).

ADD_QUEUE and ADD_BUFPOOL have to run after `send_reply_ok()` (`fuse_lowlevel.c:3145`), and they submit and wait for a verdict. So any rejection now sets `se->error` here and exits the session at :1203, on a mount the kernel has already been told is io-uring.

Could a queue that fails ADD_BUFPOOL fall back to per-entry registration instead of taking the session down — the mapping at :1030 is allocated either way, and `io_uring_register_buffers()` at :626 is already best-effort?

lib/fuse_uring.c
```diff
@@ -675,6 +869,22 @@ static void fuse_uring_handle_cqe(struct fuse_ring_queue *queue,
...
+		 */
+		if (unlikely(ent_in_out->offset >
+			     queue->bufpool_sz - ent->req_payload_sz)) {
+			fuse_log(FUSE_LOG_ERR,
+				 "Received out of range bufpool offset %u\n",
+				 ent_in_out->offset);
+			abort();
+		}
+
+		ent->op_payload = (char *)queue->bufpool + ent_in_out->offset;
```
abhinavagarwal07, line 885: How can userspace tell chunk 0 from no chunk here?

The kernel assigns one only when [`in_numargs > 1 || out_numargs`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fuse/dev_uring.c?h=v7.3-rc1#n958), and writes `offset` [only when it did](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fuse/dev_uring.c?h=v7.3-rc1#n912); otherwise it stays 0 from the [initializer](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fuse/dev_uring.c?h=v7.3-rc1#n857). `payload_sz` doesn't separate them either — a READ holding chunk 0 also reports `offset=0, payload_sz=0`.

Nothing in-tree trips over it: libfuse's own replies skip the copy when `argsize == 0` (:319), and the only `fuse_req_get_payload()` caller is `passthrough_hp.cc:1333` in `do_read()`, which always owns a chunk. But on FLUSH/RELEASE/FSYNC that call returns the pool base with `req_payload_sz` bytes, which may be another in-flight request's chunk — so the "the req is the buffer owner" contract in `fuse_lowlevel.h:2524` no longer holds for chunk-less requests.

Is this better fixed in the uapi while 7.3 is in rc — a flag bit in `ent_in_out.flags`, or an invalid-offset sentinel?

lib/fuse_uring.c
```diff
@@ -541,11 +722,21 @@ static struct fuse_ring_pool *fuse_create_ring(struct fuse_session *se)
...
 	size_t queue_sz;
 
 	if (se->debug)
-		fuse_log(FUSE_LOG_DEBUG, "starting io-uring q-depth=%d\n",
-			 se->uring.q_depth);
+		fuse_log(FUSE_LOG_DEBUG,
+			 "starting io-uring q-depth=%d bufpool=%d\n",
+			 se->uring.q_depth, use_bufpool);
+
+	if (se->uring.q_depth > UINT32_MAX / payload_sz) {
```
abhinavagarwal07, line 734: Can `payload_sz` be zero here? `se->op.init()` runs at `fuse_lowlevel.c:2959`, before the clamp, so a filesystem setting `conn->max_write = 0` reaches this divide — undefined behaviour, SIGFPE on x86-64. Today that config fails cleanly and falls back to `/dev/fuse`.

Also, this is an ADD_BUFPOOL limit (`bufpool.len` is `uint32_t`, `fuse_kernel.h:1354`) but it runs unconditionally, so it also refuses plain `-o io_uring` sessions that never issue ADD_BUFPOOL. Only above 4 GiB per queue, so nothing usable — but `use_bufpool` is right there at :725, and gating on it fixes the zero case too:

```c
if (use_bufpool && (uint64_t)se->uring.q_depth * payload_sz > UINT32_MAX)
```

test/run-tests.py
```diff
@@ -757,6 +759,8 @@ def prepare(self, spec: TestSpec):
         })
         if self.io_uring_depth is not None:
             env['FUSE_URING_QUEUE_DEPTH'] = str(self.io_uring_depth)
+        if self.io_uring_bufpool:
+            env['FUSE_URING_BUFPOOL'] = '1'
```
abhinavagarwal07, line 763: Should `--io-uring-bufpool` require `FUSE_CAP_IO_URING_BUFPOOL`?

`preflight_io_uring()` checks `FUSE_CAP_OVER_IO_URING` only (:695), and the per-session check just wants `io_uring=on` (:108). Without the cap the want bit is never set (`fuse_lowlevel.c:2946` → :2683), so the run quietly exercises the per-entry path and passes — the flag is currently the only way to reach the new code, so it's the one that shouldn't be able to lie.

`printcap` reads `capable_ext`, so the cap already lands in `self.fuse_caps` on a capable kernel, and `report_init_test_status()` emits `FUSE_INIT: FUSE_CAP_IO_URING_BUFPOOL` only when it is genuinely negotiated (:3124 clears it otherwise). Could the runner gate on the first and assert the second?

lib/fuse_uring.c
joannekoong: imo, I don't think this needs or should have a kernel-side fix. Userspace handling a buffer-less request shouldn't be touching the payload pointer in the first place. Servers calling fuse_req_get_payload() should only be invoking it on operations that actually have payloads.

lib/fuse_uring.c
joannekoong: re: payload_sz being 0 - good catch, I'll add a guard for this scenario

re: ADD_BUFPOOL limit - I'll add in a "use_bufpool" gating check, as you suggested.

lib/fuse_uring.c
joannekoong: I'll update this to round up by the page_size with it floored to FUSE_MIN_READBUFFER.

For the range check at :877, I think it'd be better to not tie this to the kernel's chunk-size logic. If the sizing picks a number that's off, the queue just runs shallower but the range check in :877 ends in abort(), so I think it'd be better to not couple :877 to internal kernel logic that could change.

lib/fuse_uring.c
joannekoong: Nice suggestion, I'll make this change to have the error path fail back to per-entry registration.

joannekoong (commented):
Thanks for taking a look at this PR @abhinavagarwal07!

test/run-tests.py
joannekoong: I'll add this in.

joannekoong:
@bsbernd I will add a run with bufpools to the pipeline.

bsbernd:
@joannekoong Thank you. I'm working on scripts to set up VMs, that can be used locally and in github in order to get runs with development kernels.

bsbernd:
@joannekoong The feature can be tested now, there are new `kernel VM` tests and bufpools need to get added there, see `.github/workflows/kernel-vm.yml`

https://github.com/libfuse/libfuse/actions/runs/34537408556/job/103072121147?pr=1593

```
=== System ===
Linux virtme-ng 7.3.0-070300rc2-generic #202609070545 SMP PREEMPT_DYNAMIC Thu Sep 10 12:34:01 UTC 2026 x86_64 GNU/Linux
Distributor ID:	Ubuntu
Description:	Ubuntu 26.04.1 LTS
...
```

bsbernd:
@joannekoong I get it reviewed till Sunday. Would you mind to put the strings into one line? Makes it easier to grep. If checkpatch complains about those long lines I will it ignore it.  I could edit this myself, though.

joannekoong:
@bsbernd will do! Thanks for taking a look

meson.build
```diff
@@ -230,24 +230,32 @@ private_cfg.set('FUSE_SYNC_INIT_DEFAULT',
      'always': 'FUSE_SYNC_INIT_ENABLED',
      'never':  'FUSE_SYNC_INIT_DISABLED'}[get_option('sync-init')])
 
-# Check for liburing with SQE128 support
+# Check for liburing with SQE128 and fixed-buffer uring_cmd support.
+# IORING_URING_CMD_FIXED and sqe->uring_cmd_flags landed a release after
+# IORING_SETUP_SQE128, so testing only the latter leaves headers that pass the
+# probe and then fail to build lib/fuse_uring.c.
```
bsbernd, line 236: I don't think we can increase the requirement for liburing from 2.2 to 2.3 for a feature that is optional - that would be libfuse regression without a really good reason, IMHO.

lib/fuse_uring.c
```diff
@@ -161,6 +207,22 @@ fuse_uring_sqe_prepare(struct io_uring_sqe *sqe, struct fuse_ring_ent *req,
...
+ */
+static void fuse_uring_sqe_set_bufpool(struct io_uring_sqe *sqe,
+				       const struct fuse_ring_queue *queue)
+{
+	if (queue->bufpool_buf_index < 0)
+		return;
+
+	sqe->uring_cmd_flags |= IORING_URING_CMD_FIXED;
+	sqe->buf_index = queue->bufpool_buf_index;
+}
```
bsbernd, line 224:
```
#ifdef IORING_URING_CMD_FIXED
...
#else
    (void)sqe;
```

The check for `queue->bufpool_buf_index < 0` can stay

lib/fuse_uring.c
```diff
@@ -463,6 +529,149 @@ static void fuse_session_destruct_uring(struct fuse_ring_pool *fuse_ring)
...
+
+/*
+ * Register the bufpool as an io_uring fixed buffer (if possible), so that the
+ * kernel can skip pinning overhead per i/o. This is best effort. This may
+ * fail because of RLIMIT_MEMLOCK. If it fails, the pool still works without
+ * it, but i/o requests incur pinning overhead.
+ */
+static void fuse_uring_register_bufpool(struct fuse_ring_queue *queue)
+{
+	struct iovec iov = {
```
bsbernd, line 629:
```
#ifdef IORING_URING_CMD_FIXED
#else
   (void)queue;
#

Or maybe ifdef of the entire function and then a no-op function under #else

File: lib/fuse_uring.c [hunk 9]
```diff
@@ -151,7 +196,8 @@ fuse_uring_sqe_prepare(struct io_uring_sqe *sqe, struct fuse_ring_ent *req,
 	sqe->flags = IOSQE_FIXED_FILE;
 	sqe->fd = 0;
 
-	sqe->rw_flags = 0;
+	sqe->uring_cmd_flags = 0;
```
bsbernd, line 199: Keep `sqe->rw_flags = 0;` and `sqe->uring_cmd_flags = 0;` under the same ifdef? Although both zero the same union.

bsbernd (approved):
@joannekoong I'm going to do the edits myself. I'm today working on https://github.com/libfuse/libfuse/pull/1522 and will submit that first - it will cause a couple of merge conflicts and change tear down a bit - better if I handle the merge myself.

lib/fuse_lowlevel.c
bsbernd, line 3061: @joannekoong we had several issues before already when replied flags didn't match requested flag - better add in `FUSE_CAP_IO_URING_BUFPOOL` to `outargflags`

joannekoong:
@bsbernd sounds good, please feel free to edit this however you'd like. Thanks!

PR #1595, opened 2026-08-26
fuse_common: Improve documentation of fuse_conn_info max_write/max_read

z-s-e:
Add some more detail on how the max_write and max_read members of the fuse_conn_info struct are handled.

### DIFF
--- a/include/fuse_common.h
+++ b/include/fuse_common.h
@@ -547,6 +547,12 @@ struct fuse_loop_config_v1 {
 #define FUSE_IOCTL_MAX_IOV	256
 
 /**
+ * This value mirrors the hard coded minimum size for read/write chunking
+ * in the kernel.
+ */
+#define FUSE_MIN_CHUNK_SIZE	4096
+
+/**
  * Connection information, passed to the ->init() method
  *
  * Some of the elements are read-write, these can be changed to
@@ -569,7 +575,20 @@ struct fuse_conn_info {
 	uint32_t proto_minor;
 
 	/**
-	 * Maximum size of the write buffer
+	 * Maximum size of write requests. Initialized with the
+	 * maximum size libfuse supports.
+	 *
+	 * The kernel may always chunk writes to at least
+	 * FUSE_MIN_CHUNK_SIZE, so setting this value to something
+	 * smaller than that will effectively be the same as setting
+	 * it to FUSE_MIN_CHUNK_SIZE.
+	 *
+	 * The kernel builds cached write requests out of whole pages.
+	 * If this value is not a multiple of the system page size
+	 * (getpagesize()), the leftover bytes either go unused or make
+	 * the page at each request boundary be sent in two requests
+	 * instead of one, thus using a multiple of the page size for
+	 * this value is recommended.
 	 */
 	uint32_t max_write;
 
@@ -578,6 +597,11 @@ struct fuse_conn_info {
 	 * limit. However, even if the filesystem does not specify a
 	 * limit, the maximum size of read requests will still be
 	 * limited by the kernel.
+	 *
+	 * The kernel may always chunk reads to at least
+	 * FUSE_MIN_CHUNK_SIZE, so setting this value to a non-zero
+	 * value smaller than that will effectively be the same as
+	 * setting it to FUSE_MIN_CHUNK_SIZE.
 	 *
 	 * NOTE: For the time being, the maximum size of read requests
 	 * must be set both here *and* passed to fuse_session_new()
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2993,6 +2993,11 @@ _do_init(fuse_req_t req, const fuse_ino_t nodeid, const void *op_in,
 	se->conn.max_write = MIN(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE);
 	se->bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
 
+	if (se->debug && se->conn.max_write < FUSE_MIN_CHUNK_SIZE) {
+		fuse_log(FUSE_LOG_DEBUG,
+			 "INIT: max_write value below FUSE_MIN_CHUNK_SIZE, may be a bug\n");
+	}
+
 	if (arg->flags & FUSE_MAX_PAGES) {
 		outarg.flags |= FUSE_MAX_PAGES;
 		outarg.max_pages = (se->conn.max_write - 1) / getpagesize() + 1;


include/fuse_common.h
```diff
@@ -569,7 +569,16 @@ struct fuse_conn_info {
 	uint32_t proto_minor;
 
 	/**
-	 * Maximum size of the write buffer
+	 * Maximum size of write requests. Initialized with the
+	 * maximum size libfuse supports.
+	 *
+	 * The kernel will always use at least page sized writes, so
+	 * setting this value to something smaller than the page size
+	 * will effectively be the same as setting it to the page size.
```
bsbernd: This is only correct for 4K pages, in fuse kernel `process_init_reply`

```
                fc->max_write = arg->minor < 5 ? 4096 : arg->max_write;                                                                                                                                      
                fc->max_write = max_t(unsigned, 4096, fc->max_write);                                                                                                                                        
``

File: include/fuse_common.h [hunk 2]
```diff
@@ -569,7 +569,16 @@ struct fuse_conn_info {
 	uint32_t proto_minor;
 
 	/**
-	 * Maximum size of the write buffer
+	 * Maximum size of write requests. Initialized with the
+	 * maximum size libfuse supports.
+	 *
+	 * The kernel will always use at least page sized writes, so
```
bsbernd: I find this rather confusing, with direct-io, FOPEN_DIRECT_IO, passthrough writes there is nothing like page size writes, it is only true for write-back

include/fuse_common.h
```diff
@@ -569,7 +569,16 @@ struct fuse_conn_info {
...
+	 * setting this value to something smaller than the page size
+	 * will effectively be the same as setting it to the page size.
+	 *
+	 * Since the kernel also uses smaller writes to align cached
+	 * writes to the page size, it is recommended to have this value
+	 * be a multiple of the page size for best performance.
```
bsbernd: Not true either for all IO cases.

include/fuse_common.h
```diff
@@ -579,6 +588,10 @@ struct fuse_conn_info {
 	 * limit, the maximum size of read requests will still be
 	 * limited by the kernel.
 	 *
+	 * The kernel will always use at least page sized reads, so
+	 * setting this value to something smaller than the page size
+	 * will effectively be the same as setting it to the page size.
```
bsbernd: Only for pages reads, but not for O_DIRECT (DIO) and FOPEN_DIRECT_IO.

include/fuse_common.h
z-s-e: Thanks for having a look at this, first off.

Is there a good reason the kernel hard-codes 4K for this? It would make sense to me that it would rather use the actual page size of the system...

Anyway, so should I rather document that hard-coded 4K value then? That only applies to the write-side, right?

include/fuse_common.h
z-s-e: Well, I know it does not apply to all IO cases, I specified it for **cached** writes, as I assume it is true for all writes that go through the page cache. Is that not right?

include/fuse_common.h
z-s-e: What I meant here is that a fuse server will always have to be prepared/be able to read a full page, as it will not chunk up read requests that are larger than a page read to reads of size smaller than a page size. From my experiment that is true even for O_DIRECT access.

Edit for clarification: So in other words, you are right that with a max_read = 1000, an O_DIRECT read of 500 bytes will generate a 500 byte read in the fuse server (and not a full 4K one), but an O_DIRECT read of 1500 bytes will not generate two reads of 1000 + 500, but will pass on the 1500 byte read un-chunked, and will only start chunking at 4K sizes, which is what I referring to. So my formulation "will always use" meant that you cannot prevent the kernel from issuing reads below a page size with this, not that it will apply to all read operations.

include/fuse_common.h
z-s-e: I have tested it for O_DIRECT file access, write requests do not get chunked below 4K here.

But it is true that this documentation probably should be expanded to include the other "corner cases", I was mainly concerned with the "standard" operations.

include/fuse_common.h
bsbernd: Well, let's day page size is 64K, why should it use page size then? The only thing that kernel could improve is to define a value like `FUESE_MIN_PAGE_SZ 4096` to avoid magic values in the middle.

bsbernd: `fc->max_read = max_t(unsigned int, 4096, ctx->max_read);`

bsbernd: So no, not limited to writes and if it would have been, that would have been a bug to correct...

z-s-e: Hm ok, so that 4K is just a fairly arbitrary value, not related to any page size? I'll replace all mentions of "page size" in this patch with 4K then.

include/fuse_common.h
bsbernd: But there is nothing like "The kernel will always use at least page sized writes" - it is only true write-back. If you you a 1B DIO write, it will stay a 1B DIO write. And DIO and FOPEN_DIRECT_IO is a "standard" operation.

include/fuse_common.h
bsbernd: Yes, in that sense 4K is an arbitrary value that coincidences with page size on x86

include/fuse_common.h
bsbernd: Ok, I had missed `cached`

include/fuse_common.h
bsbernd: Could you improve the wording please, if I get confused others will be probably even more.

include/fuse_common.h
```diff
@@ -546,6 +546,12 @@ struct fuse_loop_config_v1 {
 
 #define FUSE_IOCTL_MAX_IOV	256
 
+/**
+ * This value mirrors the hard coded minimum size for read/write chunking
+ * in the kernel.
+ */
+#define FUSE_MIN_CHUNK_SIZE	4096
```
bsbernd, line 553: It is a define only used in comments and not in the code - I think the comments should explicitly say "4096". Kernel should better  add that define, instead of its hard coding.

include/fuse_common.h
```diff
@@ -569,7 +575,18 @@ struct fuse_conn_info {
...
+	 * maximum size libfuse supports.
+	 *
+	 * The kernel may always chunk writes to at least
+	 * FUSE_MIN_CHUNK_SIZE, so setting this value to something
+	 * smaller than that will effectively be the same as setting
+	 * it to FUSE_MIN_CHUNK_SIZE.
+	 *
+	 * Since the kernel also uses smaller writes to align cached
+	 * writes to the FUSE_MIN_CHUNK_SIZE, it is recommended to
+	 * have this value be a multiple of FUSE_MIN_CHUNK_SIZE for
```
bsbernd: This part is not right, cached writes are really related to page cache and not fuse internal limits.
How about
```
        /**
         * ...
         * The kernel builds cached write requests out of whole pages.
         * If this value is not a multiple of the system page size
         * (getpagesize()), the leftover bytes either go unused or make
         * the page at each request boundary be sent in two requests
         * instead of one, so it should be a multiple of the page size.
         */
``

File: include/fuse_common.h [hunk 7]
```diff
@@ -579,6 +596,11 @@ struct fuse_conn_info {
 	 * limit, the maximum size of read requests will still be
 	 * limited by the kernel.
 	 *
+	 * The kernel may always chunk reads to at least
+	 * FUSE_MIN_CHUNK_SIZE, so setting this value to a non-zero
+	 * value smaller than that will effectively be the same as
+	 * setting it to FUSE_MIN_CHUNK_SIZE.
+	 *
```
bsbernd, line 605: Only true for cached reads and for those the it is page size - cached reads use pages/folios.

include/fuse_common.h
z-s-e: As a user of the library I personally would prefer that such constants be defined in code, rather than in comments (even if the library itself does not use it). Otherwise each application wanting to respect this value must extract it out of the comment and define it themself.

Do you insist?

include/fuse_common.h
z-s-e: Ok, I'll change

include/fuse_common.h
z-s-e: Uh wait, I don't understand. As I mentioned I have tried this with O_DIRECT (which should be uncached, right?) and there it also did not chunk below 4K. And why is the page size suddently relevant again, I thought you said it was not?

My understanding was that when the page cache is involved the kernel does full page reads, but the fuse module then still chunks those page reads further into FUSE_MIN_CHUNK_SIZE sized reads (assuming max_read is set to that or below, and the page size is actually larger) before it gets actually passed to the fuse server, no?

include/fuse_common.h
bsbernd: It will make any static analyzer to fail that checks for unused defines. If it currently succeeds that is a testing bug. So if you want to have the define, please use it somewhere in the code.

z-s-e: That is an odd requirement... then how about something like this:

```
index d310001d..55b032ec 100644
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -2993,6 +2993,10 @@ _do_init(fuse_req_t req, const fuse_ino_t nodeid, const void *op_in,
        se->conn.max_write = MIN(se->conn.max_write, bufsize - FUSE_BUFFER_HEADER_SIZE);
        se->bufsize = se->conn.max_write + FUSE_BUFFER_HEADER_SIZE;
 
+       if (se->debug && se->conn.max_write < FUSE_MIN_CHUNK_SIZE) {
+               fuse_log(FUSE_LOG_DEBUG, "INIT: max_write value below FUSE_MIN_CHUNK_SIZE, may be a bug\n");
+       }
+
        if (arg->flags & FUSE_MAX_PAGES) {
                outarg.flags |= FUSE_MAX_PAGES;
                outarg.max_pages = (se->conn.max_write - 1) / getpagesize() + 1;
```

bsbernd: I don't think a code checker that checks for dead code is an odd requirement. And depending on the checker it probably would ignore comments. Personally I think it would be easier to have 4096 hard coded into the comment - easier to read than to look up a define that is not used anywhere in the code. 

I think the check for se->conn.max_write above is a good idea!

z-s-e: A dead code checker is not the odd part, treating defines differently than functions is. You don't require libfuse to internally call all its public API functions either. Imho all public API definitions should be automatically considered alive. Anyway, no matter, I'll add this to the patch then.

bsbernd: To be honest, everything that doesn't have a caller is an issue - everything added should have a test, at least.

include/fuse_common.h
```diff
@@ -577,7 +575,20 @@ struct fuse_conn_info {
...
+	 * FUSE_MIN_CHUNK_SIZE, so setting this value to something
+	 * smaller than that will effectively be the same as setting
+	 * it to FUSE_MIN_CHUNK_SIZE.
+	 *
+	 * The kernel builds cached write requests out of whole pages.
+	 * If this value is not a multiple of the system page size
+	 * (getpagesize()), the leftover bytes either go unused or make
+	 * the page at each request boundary be sent in two requests
+	 * instead of one, thus using a multiple of the page size for
+	 * this value is recommended.
```
bsbernd, line 591: How about
```
       /**
         * Maximum size of write requests, in bytes. Defaults to 
         * 256 * getpagesize(), matching the Linux kernel's default
         * page limit below.
         *
         * Must be at least FUSE_MIN_MAX_WRITE. libfuse sizes its read
         * buffer as max_write plus 4096 bytes of header room, and the
         * Linux kernel refuses a read buffer under 8192 bytes, so a
         * smaller value makes every attempt to receive a request fail
         * with EINVAL.
         *
         * The Linux kernel also limits a request to a number of whole
         * pages, read from /proc/sys/fs/fuse/max_pages_limit (256 by
         * default), so a value above that many pages has no effect.
         *
         * Use a multiple of getpagesize(). On Linux, writeback fills a
         * request with whole pages and stops before it would pass
         * max_write, so a remainder shorter than a page stays unused. A
         * write through the page cache instead splits exactly at
         * max_write, and the page straddling that boundary is then sent
         * in two requests.
         */
```

bsbernd: Or 
```
       /**
         * Maximum size of write requests, in bytes. Defaults to
         * 256 * getpagesize(), matching the Linux kernel's default
         * page limit below.
         *
         * Must be at least FUSE_MIN_MAX_WRITE. libfuse sizes its read
         * buffer as max_write plus 4096 bytes of header room, and the
         * Linux kernel refuses a read buffer under 8192 bytes, so a
         * smaller value makes every attempt to receive a request fail
         * with EINVAL.
         *
         * The Linux kernel also limits a request to a number of whole
         * pages, read from /proc/sys/fs/fuse/max_pages_limit (256 by
         * default), so a value above that many pages has no effect.
         *
         * Writeback is page bound: the kernel fills a request with
         * whole pages and stops before it would pass max_write, so a
         * remainder shorter than a page goes unused. Use a multiple of
         * getpagesize() to avoid that.
         *
         * A write through the page cache splits at max_write exactly,
         * so with a value that is not a page multiple the page at each
         * request boundary is sent in two requests. Direct I/O writes
         * split the same way, and a user buffer that is not page
         * aligned runs out of the page budget early, making the request
         * shorter than max_write.
         */
```

include/fuse_common.h
bsbernd: How about
```
●       /**
         * Maximum size of read requests. A value of zero indicates no
         * limit. However, even if the filesystem does not specify a
         * limit, the maximum size of read requests will still be
         * limited by the kernel.
         *
         * This member is never sent to the kernel. The limit reaches it
         * only through the ``-o max_read=<n>`` mount option, which is
         * what the clamping below applies to; see the NOTE.
         *
         * The Linux kernel raises a limit below 4096 bytes to 4096, and
         * stops a read request at /proc/sys/fs/fuse/max_pages_limit
         * whole pages (256 by default) whatever this value is.
         *
         * Zero here means no mount option was given. Passing
         * ``-o max_read=0`` is not the same thing: the kernel turns
         * that into 4096, not into "no limit".
         *
         * Reads through the page cache, readahead included, are page
         * bound: a request carries whole pages only, and the kernel
         * stops adding pages before it would pass max_read. The limit
         * therefore acts as the largest multiple of getpagesize() that
         * is not greater than it - with 4 KiB pages, max_read = 5000
         * yields requests of 4096 bytes.
         *
         * Direct I/O reads are not page bound. The kernel splits at
         * max_read exactly, from the caller's offset, so a request can
         * start and end mid page. A user buffer that is not page
         * aligned runs out of the page budget early and makes the
         * request shorter than max_read.
         *
         * NOTE: For the time being, the maximum size of read requests
         * must be set both here *and* passed to fuse_session_new()
         * using the ``-o max_read=<n>`` mount option. At some point
         * in the future, specifying the mount option will no longer
         * be necessary.
         */
```

bsbernd:
@z-s-e Sorry for late reply, I'm trying to prepare new releases and finding the right words and explaining the right limits takes time. I made suggestions, through rather long paragraphs, I wonder if it should go into `doc/`. Anyway for these limits it is really important to differentiate between cached/non-cached - if I would have read your suggested changes not known the real limits, I would have been rather confused. And you got obviously confused by my explanations about `max_read`.

include/fuse_common.h
z-s-e: I see. But then what happens on a system with e.g. 64K page size when max_read is set to 4K? The kernel can't read a single page then for the page cache, how is that handled? Does that disable all caching?

I also did not understand your last sentence "A user buffer that is not page aligned runs out of the page budget early and makes the request shorter than max_read." This seems to contradict what was said directly before that, in that for direct IO page size / boundaries / alignments do not matter. Or is that not talking about direct IO anymore?

include/fuse_common.h
z-s-e: Hm, so I just tested setting max_write to the nonsensical value 0, and see how it behaves when doing various write requests. I haven't seen it fail with EINVAL, just it silently using up to 4K requests anyway. Not sure whether this is working "accidentally", but maybe it would be better to outright fail the _do_init() with a value below 4K (instead of what I do now, just doing a debug output and carrying on)?

z-s-e:
Well, personally I enjoy understanding what the kernel does exactly for cached and non-cached operations (so thanks for your explanations), but I'm not sure it is really necessary for a developer using this library to know all this in detail usually.

My reason for looking into these parameter was simply that I want to preallocate an internal buffer at startup large enough to handle all requests, so I don't need to check (and possibly resize) on every request. And for that it is sufficient to know that no matter what type of request (cached/uncached/writeback/write through or whatever), the requests reaching the fuse server in the end are guaranteed to never exceed the value set for max_read resp. max_write.

So I'd say for the primary documentation at least it is enough to say what the minimum/maximum/default/recommended values are for these parameter, without going into a deep explanation of why they are what they are, and how the kernel uses them exactly to split up the various types of requests. Maybe this is even more of an implementation detail that should not be depended upon too much anyway...

include/fuse_common.h
z-s-e: Just FYI, I tested a little bit more; I'm only testing the single threaded `fuse_session_loop()` which internally sets up a 1028K `fuse_buf` receive buffer even before init, so that buffer is not influenced by the `max_write` value and always large enough to not trigger EINVAL.

With `max_write` set to zero, the `fuse_session::bufsize` is set to 4K though, so below the 8K you mentioned, and if in multithreaded operation `_fuse_session_receive_buf()` needs to dynamically allocate the buffer, it probably will make them too small then...

bsbernd: Even single threaded would work, either by `#define FUSE_USE_VERSION FUSE_MAKE_VERSION(3, 19)`  - will not use fuse_session_loop() anymore. Or needs a recent kernel (6.18+) and usage of fuse_daemonize_early_*() - that will switch to synchronous FUSE_INIT - FUSE_INIT comes during the mount and not after mount in `fuse_session_loop`

bsbernd: Oh, actually, with single threaded mode still only one buffer that is allocated before the request, so will not trigger an issue. The correct test need sync-FUSE_INIT.

PR #1606, opened 2026-09-01
fusermount: don't try opening mnt with privs during auto_umount

szmi:
Always switch to original uid/gid before testing if opening mountpoint
would return ENOTCONN.

Otherwise it is possible for a malicious actor to replace the mountpint
with a symlink and allow opening/closing an attacker controlled path with
elevated privileges, possibly causing side effects.

### DIFF
--- a/.github/workflows/abicheck_prev_release.yml
+++ b/.github/workflows/abicheck_prev_release.yml
@@ -1,5 +1,5 @@
 ---
-name: 'libfuse ABI check'
+name: 'libfuse ABI check against previous major release'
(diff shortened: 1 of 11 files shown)


bsbernd:
@szmi could you rebase or change the destination branch to 3.18.x?

bsbernd:
@szmi If you should fight with the github interface, I got used to it, I can also do it myself

szmi:
thanks, I would have given up after 10 minutes, but fixed now

PR #1612, opened 2026-09-03
lib: skip the fusermount3 --sync-init fallback without sync init

quobert:
On EPERM the new mount API always falls back to fusermount3 --sync-init, without checking whether sync init is in use at all. fusermount3 grew that option only in 3.19, so an unprivileged mount against an older system binary logs two errors before fuse_session_mount() quietly retries via fuse_kern_mount(). Skip the fallback when sync init is off and go straight to the legacy path.

### DIFF
--- a/lib/fuse_lowlevel.c
+++ b/lib/fuse_lowlevel.c
@@ -5112,6 +5112,14 @@ static int fuse_session_mount_new_api(struct fuse_session *se,
 	if (err < 0 && errno == EPERM) {
 		char *fusermount_opts = NULL;
 
+		/*
+		 * Without sync init the handshake is pointless, and older
+		 * fusermount3 rejects --sync-init. Leave the new API;
+		 * fuse_session_mount() falls back to fuse_kern_mount().
+		 */
+		if (!se->is_sync_init)
+			goto err;
+
 		close(fd);
 		fd = -1;
 		se->fd = -1;


lib/fuse_lowlevel.c
```diff
@@ -5112,6 +5112,14 @@ static int fuse_session_mount_new_api(struct fuse_session *se,
 	if (err < 0 && errno == EPERM) {
 		char *fusermount_opts = NULL;
 
+		/*
+		 * Without sync init the handshake is pointless, and older
+		 * fusermount3 rejects --sync-init. Leave the new API;
+		 * fuse_session_mount() falls back to fuse_kern_mount().
+		 */
+		if (!se->is_sync_init)
+			goto err;
```
bsbernd, line 5121: I get what you are up, but I do not thing this is right. The new mount API is a requirement for sync-init, but it does not hold true the other way around. New mount API is supported for a long time.

If you are running with an older version of fusermount3 neither of it will work, but that is different question than sync init. I think we need to add support for a feature query of what fusermount3 supports.

bsbernd:
@quobert I think this is what should do the job https://github.com/libfuse/libfuse/pull/1613

bsbernd:
@quobert I'm closing this, I think #1613 is the better solution. Please let me know if you have objections.

quobert:
@bsbernd sorry for the late reply, I was busy with other projects lately. Thank you for quickly picking this up!

PR #1622, opened 2026-09-14
example/passthrough: fix buffered write on FreeBSD

### LINKED ISSUE #1354: passthrough/tst_append() fails on FreeBSD
**Describe the bug**
I updated the [FreeBSD port filesystems/fusefs-libs3](https://www.freshports.org/filesystems/fusefs-libs3) to version 3.17.4 locally and added the test-target.
```make
do-test:
	${SU_CMD} "${PYTHON_CMD} -m pytest ${WRKSRC}/_build/test/"
```

I then added the following patch for the `tests/` to skip the Linux-only exercises:
```diff
--- test/test_ctests.py 2025-08-19 14:07:15.000000000 -0400
+++ test/test_ctests.py 2025-10-24 11:15:19.343865000 -0400
@@ -52,5 +52,5 @@
 
 names = [ 'notify_inval_inode', 'invalidate_path' ]
-if fuse_proto >= (7,15):
+if fuse_proto >= (7,15) and os.uname().sysname == 'Linux':
     names.append('notify_store_retrieve')
 @pytest.mark.skipif(fuse_proto < (7,12),
--- test/test_custom_io.py      2025-08-19 14:07:15.000000000 -0400
+++ test/test_custom_io.py      2025-10-24 11:24:08.029581000 -0400
@@ -59,5 +59,6 @@
     assert response_payload[0] == FUSE_MAJOR_VERSION
 
-
+@pytest.mark.skipif(os.uname().sysname != 'Linux',
+    reason = 'hello_ll_uds is a Linux-only feature')
 def test_hello_uds(output_checker):
     cmdline = base_cmdline + [pjoin(basename, 'example', 'hello_ll_uds')]
--- test/test_examples.py       2025-08-19 14:07:15.000000000 -0400
+++ test/test_examples.py       2025-10-24 11:36:26.807342000 -0400
@@ -96,4 +96,9 @@
 @pytest.mark.parametrize("name", ('hello', 'hello_ll', 'test/hello'))
 def test_hello(tmpdir, name, options, cmdline_builder, output_checker):
+    if cmdline_builder.__name__.startswith('invoke_mount_fuse') and\
+        sys.platform != 'linux':
+        pytest.skip('mount.fuse3 is only available on Linux')
+    if name == 'hello_ll' and not hasattr(os, 'setxattr'):
+        pytest.skip('setxattr is not present on %s' % os.uname().sysname)
     logger = logging.getLogger(__name__)
     mnt_dir = str(tmpdir)
```

And now I'm looking at what seems like a _real_ test-failure: the `tst_append()` function found in `test/test_examples.py`:
```python
def tst_append(src_dir, mnt_dir):
    name = name_generator()
    os_create(pjoin(src_dir, name))
    fullname = pjoin(mnt_dir, name)
    with os_open(fullname, os.O_WRONLY) as fd:
        os.write(fd, b'foo\n')
    with os_open(fullname, os.O_WRONLY|os.O_APPEND) as fd:
        os.write(fd, b'bar\n')

    with open(fullname, 'rb') as fh:
        assert fh.read() == b'foo\nbar\n'
```
fails at its second `os.write()` call:
```
...
  File "/envy/ports/filesystems/fusefs-libs3/work/fuse-3.17.4/_build/test/test_examples.py", line 191, in test_passthrough
    tst_append(src_dir, work_dir)
  File "/envy/ports/filesystems/fusefs-libs3/work/fuse-3.17.4/_build/test/test_examples.py", line 670, in tst_append
    os.write(fd, b'bar\n')
OSError: [Errno 9] Bad file descriptor
========================================================================== short test summary info ===========================================================================
SKIPPED [1] work/fuse-3.17.4/_build/test/test_ctests.py:32: Requires kernel 3.14 or newer
SKIPPED [1] work/fuse-3.17.4/_build/test/test_custom_io.py:61: hello_ll_uds is a Linux-only feature
SKIPPED [6] work/fuse-3.17.4/_build/test/test_examples.py:100: mount.fuse3 is only available on Linux
SKIPPED [1] work/fuse-3.17.4/_build/test/test_examples.py:102: setxattr is not present on FreeBSD
FAILED work/fuse-3.17.4/_build/test/test_examples.py::test_passthrough[False-passthrough-False] - OSError: [Errno 9] Bad file descriptor
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================================== 1 failed, 11 passed, 9 skipped in 27.25s ==================================================================
```
**To Reproduce**
Attempt testing on FreeBSD.

**Expected behavior**
Successful run of the test-suit -- with any tests specific to other operating systems skipped.

**Desktop:**
 - FreeBSD
 - Version 14.3/stable (i386)

CismonX:
On FreeBSD, the kernel may issue a `READ` request during buffered write [^1].
This breaks `fuse_test_append` and `fuse_test_seek` in our test suite.

Closes: https://github.com/libfuse/libfuse/issues/1354

[^1]: https://github.com/freebsd/freebsd-src/blob/release/15.1.0/sys/fs/fuse/fuse_io.c#L576


example/passthrough_hp.cc
```diff
@@ -1314,11 +1314,19 @@ static void sfs_open(fuse_req_t req, fuse_ino_t ino, fuse_file_info *fi)
 {
 	Inode &inode = get_inode(ino);
 
+#ifdef __FreeBSD__
+	/* During buffered write, the kernel may issue a READ request,
+	   irrespective of whether writeback cache is enabled. */
+	if (!(fi->flags & O_DIRECT)) {
```
bsbernd, line 1320: Out of interest, BSD does support FOPEN_DIRECT_IO? 

I think the right condition is `if (!(fi->flags & O_DIRECT) && !fs.direct_io) {` and independent on `__FreeBSD__`. Linux kernel will also do reads on write for buffered write-through if write is less than a page.

CismonX: > BSD does support FOPEN_DIRECT_IO?

Yes, FreeBSD supports `FOPEN_DIRECT_IO`.

> I think the right condition is `if (!(fi->flags & O_DIRECT) && !fs.direct_io) {`

You're right. I overlooked that launch option...

> ...and independent on `__FreeBSD__`. Linux kernel will also do reads on write for buffered write-through if write is less than a page.

Can you elaborate on this? IIRC Linux kernel only do that in writeback-cache mode.

I skimmed through the [kernel source for write-through](https://github.com/torvalds/linux/blob/v7.2/fs/fuse/file.c#L1356) and it doesn't seem to do reads. And the [kernel docs](https://www.kernel.org/doc/html/next/filesystems/fuse/fuse-io.html) doesn't say anything new about this.

bsbernd: Ah sorry, my fault. I mixed it up with write-back.

PR #1624, opened 2026-09-15
example/passthrough: fix readdir on FreeBSD

CismonX:
**Patch 1**: example/passthrough: use `telldir()` instead of `d_off`

Using the `d_off` field of `struct dirent` returned by `readdir()` for `seekdir()` is non-standard, and may not work under FreeBSD.

This breaks the `fuse_test_readdir_big` test for `passthrough_hp`.

---

**Patch 2**: example/passthrough: use `rewinddir()` for zero offset

A zero offset in a `FUSE_READDIR(PLUS)` request implies `rewinddir()`. Calling `seekdir(dirp, 0)` instead is non-standard and may not work under FreeBSD.

Reproducible by running Python `os.listdir()` multiple times on a same non-empty directory.


bsbernd:
@CismonX could you fix checkpatch?

```
#36: FILE: example/passthrough_fh.c:168:
+			/* Subtract the one that we add when calling
+			   telldir() below */

WARNING:BLOCK_COMMENT_STYLE: Block comments use a trailing */ on a separate line
#36: FILE: example/passthrough_fh.c:168:
+			   telldir() below */
```

You can run via `.github/workflows/run-checkpatch.sh`, I'm going to move it into .scripts in the future

CismonX:
@bsbernd Done.

PR #1631, opened 2026-09-18
tests: Enable BSD testing

bsbernd:
This was prepared on switching to the new way to run tests and when the text matrix was introduced, but was not enabled as it likely would have found issues and goal for that PR was to simply running tests. Time to handle BSD now.


CismonX:
There are still issues preventing some tests from passing under FreeBSD. Some of them are tricky to fix so they have to be skipped (at least for now).

I'll spend some time this weekend to work on it. Hopefully we can get this PR merged before long :)

bsbernd:
Thanks @CismonX I'm just adding a script that allows to run VMs in the same way as `test/ci/vm-run.sh` so that it can easily run locally. I will then let an AI agent try to fix all the issues.

bsbernd:
@CismonX @vassilit @asomers If you should have time, please review.

vassilit:
I think we may benefit keeping `vmactions/freebsd-vm`, because the workflow can only be run on Github anyway, even without it (well, there is [act](https://github.com/nektos/act)). Taking out the shell scripts from the YAML file is absolutly the way to go to enable local runs.
I do the same for rmlint ([freebsd.yaml](https://github.com/sahib/rmlint/blob/master/.github/workflows/freebsd.yaml)).

I have a `local-vm.sh` that provision a FreeBSD image using the CLOUDINIT (nuageinit) images and emulate vmactions/freebsd-vm locally: sync the repository with the VM via rsync and execute prepare and test. I have to clean it (because it's a bit dirty now because I wanted to test the UEFI/BIOS routes) and I will open a PR. But without it, you can download from VM-IMAGES, and run it with qemu very simply. For the CI, `vmactions/freebsd-vm` is convenient and we don't have to reinvent the wheel.

bsbernd:
@vassilit Thanks for review! We have the shell script anyway now, so it is not really re-invent the wheel. I personally do not like github actions, because I cannot easily run it on my local system.  So I want to have the script one way or the other. And now pointing the worflow to that script makes it easy for anyone who wants to reproduce issues. Let's say the next linux commit breaks BSD and someone wants to run BSD - my personal flow here is to go to the action, lookup what is does and run it locally. We could (and maybe also should) document how to use the script for BSD, but then one first has to find the documentation - the action file is self explaining.

vassilit:
Also, about 666dbf2, I have a fix I think. Strange timing, as I've stumbled upon this in the morning, because [fuse-archive ](https://github.com/google/fuse-archive) enable _use_ino_ (`-o use_ino` on fuse2 and `cfg->use_ino = true` in the init handler on fuse3) to simulate hard-links from tar archives and I got the FUSE protocol violation (to accomodate NFS) warning.

I will post a PR. But the test can be disabled in the meantime.

a simple one like this should do the work, but it's a draft at this point:
```c
	if (f->conf.use_ino)
		fuse_unset_feature_flag(conn, FUSE_CAP_EXPORT_SUPPORT);
```

For future reference, [here](https://github.com/freebsd/freebsd-src/blob/b774fbdb0077571dd34ca55ecb87199895ff47d3/sys/fs/fuse/fuse_node.c#L317) is where the warning is from.

bsbernd:
@vassilit I dropped the patches for the moment, they are not required by this PR

vassilit (approved):

doc/README.BSD
```diff
@@ -0,0 +1,66 @@
...
+- Avoid it by mounting outside the mirrored tree:
+  -o modules=subdir,subdir=DIR restricts the passthrough to DIR.
+- -o intr lets a signal break the wait instead of the timeout.
+- A fusefs fix (MNTK_LOOKUP_SHARED or an own vop_vptocnp) is the real
+  solution; not reported upstream yet.
+
+Console warnings
+----------------
+
+- "FUSE protocol violation ... file has different inode number and nodeid"
```
asomers, line 64: Yes, FreeBSD does warn about that.  @bsbernd do you know of any legitimate reason why a FUSE daemon might report different inode numbers and nodeids for the same file?

test/cases/examples/ioctl-ll.sh
```diff
@@ -3,6 +3,7 @@
 
 . "$TEST_LIB/common.sh"
 
+_require_linux FUSE_IOCTL
```
asomers, line 6: FYI FreeBSD 16.0 will include FUSE_IOCTL support.

test/cases/lib/passthrough.sh
```diff
@@ -8,6 +8,10 @@
 #                   directory is $TEST_MNT$TEST_SRC rather than $TEST_MNT
 #   PT_SRC_VISIBLE  1 when a change made directly in $TEST_SRC is expected to
 #                   show up under the mount straight away
+#   PT_HIGHLEVEL    1 when the example is on the high-level API, which keeps
+#                   one node per path and so answers LINK with a nodeid that
+#                   is not the source's. The FreeBSD kernel refuses such a
+#                   reply with EIO, so the hard-link checks are left out there.
```
asomers, line 14: This sounds to me like a bug in the example.  Hard links must always share the same inode number.   That's what it means to be a hard link.

doc/README.BSD
CismonX: The libfuse high-level API uses an incremental counter for both nodeid and `st_ino`.

If option `use_ino` is enabled, `st_ino` can be overridden, but nodeid cannot, which leads to different nodeid and inode number for the same file.

test/cases/lib/passthrough.sh
CismonX: Or rather it's a _limitation_ in the example.

Since the high-level API user cannot override nodeid, there's nothing they can do about it.

doc/README.BSD
bsbernd: @asomers Yes absolutely, a node ID is basically a unique identifier for the file, the inode number is what appears in stat(). Example, basically every distributed file system I have worked with needs >64 bit inode numbers. 
What the file system then do is to give uintptr_t of the local in-memory object as node-id and an inode number that is not perfectly (but as much as possible) collision free.

I plan to create distributed example file system, in order to demonstrate what needs to be done for DLM support in Linux-FUSE (we have the patches (linux-kernel/libfuse, but file system is closed source - hard to show the need for these complex changes). That demo file system will likely also need node-ID != stat->st_ino.

test/cases/examples/ioctl-ll.sh
bsbernd: Ok, let's handle it when it is ready :) Please let me then.

test/cases/lib/passthrough.sh
bsbernd: Well, please feel free to submit patches - I'm not using BSD for anything personally, so I can only help with basic support.

PR #1634, opened 2026-09-19
test: skip some tests for open() on FreeBSD

CismonX:
Since commit c4f3115696, `passthrough*` servers under `example/` may fail with `EACCES` when opening a file `O_WRONLY` on FreeBSD, because it attempts to open the underlying file `O_RDWR` instead.

May also happen under Linux when writeback cache is enabled, but since `passthrough-hp.sh` has `PT_SRC_VISIBLE=0` (not running `test_syscalls`), this doesn't break the tests.


bsbernd:
@CismonX I guess it only fails when run as user, but CI runs as root? Let's see if this fails: https://github.com/libfuse/libfuse/pull/1635 without this PR here.

